PR #12, opened 2015-01-14
added functionality to save address and private key pair to disk

dydx:
I know I'd forget that information under my own volition, so why not save it to disk? I opted for 'data/mysites.txt' as an easy place to remember, and alerted the user that such an action was taking place.


HelloZeroNet:
Thx, but I rather not store save the private keys to a simple txt file. From the version 0.1.4 its harder not to save the private key at siteCreate :)

shakna-israel:
I'd totally agree with not saving into the data folder, that's where the serving up happens, but I think saving to disk would make some sense. 

The issue is the key being a long hard to remember sequence, and we deal with this all the time: SSH Keys are saved to file as well.

PR #25, opened 2015-01-18
fixed subtl.py from trying to parse a port string as integer

lacksfish:
When running zeronet.py with the --fileserver_port argument set to an open port, the following error occurs

[22:42:28] Site:1EU1tb..E4Vr Tracker error: error: cannot convert argument to integer in Site.py line 233 > subtl.py line 79

As the fileserver_port is of the type string, parsing it to integer fixes this issue.
I don't know if the subtl library is the correct place to implement this fix, but for now it helped me to get it to run.

### DIFF
--- a/src/lib/subtl/subtl.py
+++ b/src/lib/subtl/subtl.py
@@ -70,6 +70,9 @@ class UdpTrackerClient:
         # Check and raise if missing fields
         self._check_fields(args, fields)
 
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
Thanks for reporting. I rather not touch the external subtl library,but its fixed at config parse: 2000f5b38eb5d7e6f29bc29232b8493d31b6817a

Issue #17, opened 2015-01-15
Open browser after start zeronet.py


HelloZeroNet:
Cons:
 - It could be irratating if you just restart it or you want to put in autorun

Possible solutions:
 - Open by default but can be disabled via `--dont-open-browser` command line parameter
 - Create an another file that starts zeronet.py and opens the browser too: start.cmd
 - After starting zeronet.py display a message: `-- Press enter to open http://localhost:43110/ in your browser --`

shakna-israel:
I like the start.cmd idea.

As to opening in the browser:

```
import webbrowser # From the standard library
tab = 2 # Use a new tab if the default browser is open
url = "http://127.0.0.1:43110" # Set a variable to open
webbrowser.open(url,tab=tab) # Open the web browser!
```

ymski:
I could help out with this. I just did something similar for another project with OSX. This would literally be the command for it.
```
osascript -e "open location \"http://127.0.0.1:43110\""
```

volker48:
Just put in a PR that should take care of this.

ymski:
@volker48 

Just a suggestion and it can always be added later, but some people might want it opened in a specific browser, as well as some people saying this opens a browser other than the default. Maybe we can have this as an option or take in an arg (browser name) that would open it.

Here's how a specified browser can be opened:
https://docs.python.org/2/library/webbrowser.html#webbrowser.get

volker48:
@ymski I updated the pull request. You can now specify the browser when running start.py. Example, `python start.py firefox` will launch firefox instead of the default browser. If you don't specify it still uses the default browser.

HelloZeroNet:
Thanks, merged!

chenshaoju:
--dont-open-browser

This parameter has no long working in current ZeroNet Bundle (Windows)

PR #36, opened 2015-01-26
start.py script to start the server and launch a new browser tab

volker48:
This should resolve issue #17. By default running `python start.py` will launch the zeronet server in a subprocess then open the default browser. The user can also specify which browser to open. For example, `python start.py firefox`.


HelloZeroNet:
Thanks!

PR #46, opened 2015-02-08
Enabled meta viewport, to support Responsive Design.

shakna-israel:
Meta tags inside the iframe are largely ignored, to enable developers to build responsive applications, they need a viewport. Though many developers tweak the viewport setting, the one presented here is by-and-large the most commonly used, as it's simple and effective.

### DIFF
--- a/src/Ui/template/wrapper.html
+++ b/src/Ui/template/wrapper.html
@@ -6,6 +6,7 @@
  <title>{title} - ZeroNet</title>
  <meta charset="utf-8">
  <meta http-equiv="content-type" content="text/html; charset=utf-8" />
+ <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <link rel="stylesheet" href="/uimedia/all.css" />
 </head>
 <body style="{body_style}">


HelloZeroNet:
Not all site optimalized to mobile devices, so maybe its not a good idea to make it default, but added (bccd246f716ce191d872fdae2a84b27b66300641) a viewport option to content.json, you can use it like this:
```json
 ...
  "signs_required": 1, 
  "title": "ZeroTalk", 
  "viewport": "width=device-width, initial-scale=1.0", 
  "zeronet_version": "0.2.0"
}
```
and added WrapperAPI command too: `@cmd "wrapperSetViewport", "width=device-width, initial-scale=1.0"`

I don't have any smartphones to test it, but works in chrome's device emulator. 

Set the viewport to ZeroTalk site, it would be nice if anyone can check if its displays nicely on any small screen device.

shakna-israel:
I'm guessing that "viewport" accepts any input, and plays out into the meta tag on the page.

e.g.
```
"viewport":"content="initial-scale=1, maximum-scale=1"
```
should work equally as well as:
```
"viewport":"content="width=500, initial-scale=1"
```

or the original example:
```
 "viewport": "width=device-width, initial-scale=1.0"
```

If not, it might be an idea due to the size and usefulness of the viewport meta tag: [W3 Draft Specification](http://dev.w3.org/csswg/css-device-adapt/#viewport-meta).

If it does, this is a nice fix, but will need to be clearly spelled out in docs somewhere.

Nice job.

I'll check ZeroTalk soon as I update the Proxy.

shakna-israel:
ZeroTalk seems to work well on mobile - except for one or two things.

New topic button

The Request Access button also throws the user way off screen, unfortunately, I didn't screenshot that, as I realised there was no sign out, and my whole uni has access to the server I was using. So I ran off panicking.

HelloZeroNet:
only ` "viewport": "width=device-width, initial-scale=1.0"` works, `"` quotes will be escaped, so you cant break out from the `content` attribute of the meta tag

PR #50, opened 2015-02-18
Cleaned up some of the language

danielquinn:
Some of the English wasn't quite right, so I thought you'd appreciate some fixes.  I've also added syntax highlighting for the bash code blocks, and switched out the use of `-` to use the more-common `*` for bulletted lists.  Lastly, I fixed some spelling.


HelloZeroNet:
Thanks!

Issue #20, opened 2015-01-16
Detect installed upnpc


arthurlutz:
I think the README should recommend installing upnpc via packages (for example apt-get install miniupnc) and then ZeroNet could check if upnpc is in the path instead of having to explicitly use --upnpc option.

HelloZeroNet:
We have pure python upnp, so fixed in 34f6d1ee7c8b5a8958f675c5e17a95c34c6a4967

PR #53, opened 2015-02-25
Python-only Upnp solution

sirMackk:
To summarize everything here: this solves #5 and #20. 

It's a pure python stdlib + gevent way to open up a port on router to the current host using UPnP. I've tested it quite a bit on my linux box and a bit on a windows box I got access to. I've also had some feedback (and a working solution!) from @hobbldygoop in https://github.com/sirMackk/ZeroNet/commit/fdcd15cf8df0008a2070647d4d28ffedb503fba2#commitcomment-9863928 about making it work on OSX (THANKS!). :+1:


HelloZeroNet:
Thank you! :)

Issue #101, opened 2015-05-13
Printing prints only the link to home on all sites


thesoftwarejedi:
When printing any zeronet site, the only printable content is the notification area and such.  Print styles are needed to do the reverse (hide the parent page overlays, set position of the iframe to fixed).

HelloZeroNet:
merged

PR #102, opened 2015-05-13
#101 add print styles to allow iframe sites to print

thesoftwarejedi:
add print styles to allow pages to print


HelloZeroNet:
thanks

PR #108, opened 2015-05-27
Added Dockerfile

n3r0-ch:
I added a Dockerfile. Could you please create a automated build on Dockerhub?


HelloZeroNet:
Thanks, never used docker before, I try to add it to dockerhub.

PR #118, opened 2015-06-17
Fixing zeroname_updater.py support under windows

Idealcoder:
Several fixes to allow the script to run under windows without errors, fixing:

* Windows use different file locations
* Windows having troubles with UTF8
* Change the config defaults so that it starts scanning at first block with a namecoin domain.
* Added a note about data/names.json file being needed. 

Thanks for @nofish who helped over IRC.


HelloZeroNet:
Thanks!

PR #119, opened 2015-06-17
partial cleanup of main.py

mattseh:


mattseh:
20 minutes with PyCharm :)

HelloZeroNet:
I think we should not move the not-on-top module imports, to allow only load the modules if they are really needed (eg gc module) and it could matter if the module loaded before or after monkey patching (for example in case of proxy).

mattseh:
I think ideally everything should be imported at the start of a file, as it makes the code a lot less brittle. In my experience, there is no real world performance penalty for  loading everything that's needed at the start of the file.

HelloZeroNet:
Unfortunately importing some libs before socket monkey patching broken proxy (TOR) support (caused hang-up on startup), so i reverted the necessary imports to previous state: f58aa5f78e6f223d55d27560a4c034c247ca3ae4

PR #120, opened 2015-06-18
Created Zeroname-local plugin that use local namecoind for lookup

Idealcoder:
Reused Zeroname code, and swapped out the reolveDomain function. It now connects to namecoind directly using RPC and then run rpc.name_show() to find domain.

It might be a bit slow as there is no caching - it looks up the domain each time on the blockchain. 
I'm not sure how to handle the bitcoinrpc dependancy, so I have left that out.

Also, it is disabled by default. If it works well we might want to consider a "automatic" detection system - check if namecoin is installed, and if not fallback to hosted Zeroname.

### DIFF
--- a/plugins/disabled-Zeroname-local/SiteManagerPlugin.py
+++ b/plugins/disabled-Zeroname-local/SiteManagerPlugin.py
@@ -1,0 +1,68 @@
+import logging, json, os, re, sys, time
+import gevent
+from Plugin import PluginManager
+from Config import config
+from Debug import Debug
+from domainLookup import lookupDomain
+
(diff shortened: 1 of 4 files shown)


Idealcoder:
For some reason, when using this plugin ZeroHello uses site addresses rather than namecoin .bit domains, but manually typing the domain works fine?

HelloZeroNet:
Its because ZeroHello currently only checks if Zeroname plugin is enabled:
https://github.com/HelloZeroNet/ZeroHello/blob/master/js/ZeroHello.coffee#L99
I'm going to update it soon

PR #143, opened 2015-08-31
Read bootstrap nodes from file

bashrc:
If a file exists in the current directory called "bootstrap" then its contents will be used as a replacement for the hardcoded bootstrap nodes

### DIFF
--- a/src/Site/Site.py
+++ b/src/Site/Site.py
@@ -591,9 +591,11 @@ class Site:
             for node in bootstrap_nodes:
                 node_fields = node.split(' ')
                 if len(node_fields) == 3:
-                    protocol = node_fields[0].strip()
+                    protocol = node_fields[0].strip().lower()
                     ip = node_fields[1].strip()
-                    port = int(node_fields[2].strip())
+                    port = None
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
I think its better to keep this setting in the config file.

Since the [latest update](https://github.com/HelloZeroNet/ZeroNet/commit/0de6496f96316514a07c4b1cbaf4a9f29a2411cb) you can specify the trackers by creating a `zeronet.conf` file eg.:
```
[global]
trackers =
	udp://sugoi.pomf.se:2710
	http://torrent.gresille.org/announce
```

bashrc:
A problem with this is that the list of trackers is only read once at startup. As peers enter or leave the network the list of available trackers will change, but the system won't be updated.

HelloZeroNet:
[Added an another parameter](fa37f58982d3ea430605929aad961243dfceb600) using `zeronet.py --trackers_file bootstrap.txt` the trackers reloaded before the announces. Example for bootstrap.txt:
```
udp://sugoi.pomf.se:2710
http://torrent.gresille.org/announce
```

bashrc:
Excellent

PR #147, opened 2015-09-06
Update README.md

noflag:
fixes bad msgpack version on mint


HelloZeroNet:
The problem is Debian does not comes with pip installed and if you want to install it using pip you could need other dependencies installed (c++ comiler, python source code and other heading files). Debian 7+ has the correct msgpack version using apt-get so I would leave it as default install method.

For other distributions the pip method is recommended: https://github.com/HelloZeroNet/ZeroNet#other-linux-or-without-root-access

PR #152, opened 2015-09-06
Reap child processes

### LINKED ISSUE #151: Please reap child processes
When calling the external `openssl` command using `subprocess.Popen()` the childprocesses are never reaped, thus leaving zombies behind.  After having read the output, a call to the childprocess's `.wait()` method should be called.

kseistrup:
Previously child processes were not reaped, leaving zombies behind.
Now each child process's `.wait()` method is called before proceeding.

Closes #151.


HelloZeroNet:
Thanks!

PR #174, opened 2015-09-24
Memory imporvements

deavmi:
Import only what you need, not all the other data.


HelloZeroNet:
Thanks, but i think its reduces readability.

deavmi:
There is that point too, I was thinking about that.

deavmi:
Thanks tho. :smile_cat:

PR #175, opened 2015-09-24
Memory imporvements

deavmi:


HelloZeroNet:
Thanks, but i think its reduces readability.

PR #194, opened 2015-10-21
Fix bug when startup with windows is checked it doesn't remember it because of extra spaces in zeronet.cmd

TheNain38:


plugins/Trayicon/TrayiconPlugin.py
```diff
@@ -115,9 +115,9 @@ def formatAutorun(self):
         cmd = " ".join(args)
 
         # Dont open browser on autorun
-        cmd = cmd.replace("start.py", "zeronet.py").replace('"--open_browser"', "").replace('"default_browser"', "")
+        cmd = cmd.replace("start.py", "zeronet.py").replace(' "--open_browser"', "").replace(' "default_browser"', "")
```
TheNain38: It could be replaced with `cmd = cmd.replace('start.py --open_browser "default_browser"', "zeronet.py")` ?

TheNain38:
@HelloZeroNet Is there a reason you don't merge this, or you didn't seen it?

HelloZeroNet:
I'm unable to reproduce the problem

TheNain38:
If you start zeronet.py with `..\Python\python.exe zeronet.py` and you check the option to start with Windows, then you exit and you start ZeroNet by double clicking on `zeronet.cmd` and the option isn't checked

TheNain38:
It's because the current code only replaces `start.py` `"--open_browser"` `"default_browser"` and the two spaces, one before `"--open_browser"` and one before `"default_browser"`. In the file it left two spaces at the end of the last line

HelloZeroNet:
Thanks, I was able to reproduce, added a [different fix](https://github.com/HelloZeroNet/ZeroNet/commit/713baeab63b546dca6a69608a2e3613dd6e87118#diff-96bfdc2cfcb5739e88aad648cba5e983R120), as it conflicts with other pr

TheNain38:
Thanks

PR #197, opened 2015-10-23
Close WebSocket connection gracefully

TheNain38:


HelloZeroNet:
Does it helps your problem?

TheNain38:
No, but it's always better to gracefully close a connection

HelloZeroNet:
I think its better let the browser handle the connection. (close might be not compatible with every browser: http://stackoverflow.com/questions/5545161/unable-to-close-websocket-with-javascript)

TheNain38:
Ok, so you close it?

HelloZeroNet:
Yeah, i think we can close it

PR #235, opened 2016-01-15
Fix error when opening menu if file size is null

TheNain38:
@HelloZeroNet It fixes a division by 0
Or, does it makes more sense to not add the "li" at all if size is zero?


HelloZeroNet:
Thanks

PR #284, opened 2016-03-02
README.md modifications

ashleypt:
Virtualenv is a great way for people cloning from the git repo to setup zeronet's environment with dependencies in a consistent way without requiring a vm like vagrant/docker


ashleypt:
Okay. Switched to source

HelloZeroNet:
something is bad with formatting: https://github.com/ashleypt/ZeroNet#docker

PR #339, opened 2016-03-07
plugins/Trayicon for osx

barrabinfc:
This is the `plugin-trayicon` for osx. It is pretty dumb and it only serves as an
visual indication that ZeroNet is running. You can also quit zeronet daemon from the menu.

screen shot 2016-03-07 at 2 22 03 pm

I didn't added the statistics like the win32 version because the tray app should be a different process (in OSX), and i would have to implement some IPC. So i've stopped before doing this overly complex.


DaniellMesquito:
I suggest show all menus that the Windows version shows.

ysc3839:
Can't you create the icon in same process?

PR #362, opened 2016-03-13
Disabled benchmarking in MultiUser

AceLewis:
The Benchmarking function is CPU intensive, so someone could easily DOS
a proxy.

Pointed out in:
http://127.0.0.1:43110/Talk.ZeroNetwork.bit/?Topic:1_1LJP7tDoGnWNppUGJoNS8cJbmYTS1TecCC/Zero+proxies+vulnerability


HelloZeroNet:
Benchmark is not a websocket function + you redirect it to siteDelete

PR #371, opened 2016-03-17
apt-get commands should be in one line

gyulaweber:
according to https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/ (#apt-get) it's better to do the install and the cleanup with one command


HelloZeroNet:
thanks

PR #379, opened 2016-03-19
Add tools dir to .gitignore

OliverCole:


HelloZeroNet:
Tor directory is necessary to run zeronet on windows

OliverCole:
Agreed, but we don't want to commit the binaries, or have them show up in local diffs just because we ran the code or unit tests, right?

HelloZeroNet:
Adding pattern for non-default files could work.

PR #380, opened 2016-03-19
Add tor and phantomjs dir to .gitignore

OliverCole:


HelloZeroNet:
The problem is it will delete the tor settings and readme

OliverCole:
Do you mean torrc? This won't affect that - just ensure it doesn't appear in diffs or get accidentally committed.

PR #385, opened 2016-03-20
Support all custom namecoin.conf parameters

TheNain38:


kseistrup:
Actually, you can play with more RPC parameters in `~/.namecoin/namecoin.conf`:
```
  -rpcconnect=<ip>
       Send commands to node running on <ip> (default: 127.0.0.1)

  -rpcport=<port>
       Connect to JSON-RPC on <port> (default: 8336 or testnet: 18336)

  -rpcwait
       Wait for RPC server to start

  -rpcuser=<user>
       Username for JSON-RPC connections

  -rpcpassword=<pw>
       Password for JSON-RPC connections

  -rpcclienttimeout=<n>
       Timeout during HTTP requests (default: 900)
```
so perhaps it would be prudent to take all of them into consideration, e.g., something like this:
```python
def init_rpc(config):
    """Initialize Namecoin RPC"""
    rpcdata = {
        'config': config,
        'connect': 'localhost',
        'port': '8336',
        'user': 'PLACEHOLDER',
        'password': 'PLACEHOLDER'
        'clienttimeout': '900',
    }
    try:
        from bitcoinrpc.authproxy import AuthServiceProxy
        fptr = open(rpcdata['config'], 'r')
        lines = fptr.readlines()
        fptr.close()
    except:
        return None  # Or take some other appropriate action

    for line in lines:
        if not line.startswith('rpc'):
            continue
        keyval = line.split(None, 1)[0]
        (key, val) = keyval.split('=', 1)
        if not key or not val:
            continue
        rpcdata[key[3:]] = val

    url = 'http://%(user)s:%(password)s@%(connect)s:%(port)s' % rpcdata

    return AuthServiceProxy(url, timeout=int(rpcdata['clienttimeout']))
```
Cheers!

TheNain38:
@kseistrup Thanks!

kseistrup:
Please not that I haven't tested the actual code, it was just taken from off top of my head. Specifically, I'm used to coding in Python 3, and I'm unsure if the example above will run without errors. I did try to avoid `with open(config, 'r') as fptr:` because I can't remember when the with-statement was introduced. Also, I'm unsure if there was a `.startswith()` method on string objects in Python 2. But such things should be easy to spot for people who are used to Python 2.

TheNain38:
I tested the code that is in domainLookup.py, and it works, also I did not take your code directly, I modified it a little bit

kseistrup:
Cool! :+1:

HelloZeroNet:
Please use http://zeronet.readthedocs.org/en/latest/help_zeronet/contributing/

TheNain38:
@HelloZeroNet Done

PR #386, opened 2016-03-20
Automatically delete empty folders in which, files are deleted

TheNain38:


HelloZeroNet:
I don't think its a good idea listing directory on every update

TheNain38:
It only looks at directories, in which files are deleted

TheNain38:
@HelloZeroNet Now, it only tries once to list and delete the folder

HelloZeroNet:
Yeah it's better this way, but I think importing/using Counter is unnecessary, a standard dict will do, and please put a try/exception block like the files has.

TheNain38:
@HelloZeroNet Done!

TheNain38:
@HelloZeroNet I though of one thing that could be made, that is to only list and delete the leaf directories, that is, for example: if it is deleting files in "foo", "foo/bar and " foo/bar2, then it would only try to delete the deepest ones, that is, "foo/bar" and "foo/bar2" as I'm using `os.removedirs`, it tries to remove the directories recursively.

TheNain38:
@HelloZeroNet Done!

TheNain38:
@HelloZeroNet Wait

HelloZeroNet:
Thanks, i have used a simplier solution for this: https://github.com/HelloZeroNet/ZeroNet/commit/6d222b6ed768700e188bf862ac36ef075e5e989d

PR #389, opened 2016-03-22
Allow to use custom ZeroNet site to resolve .bit domains

TheNain38:


HelloZeroNet:
Thanks (the plugin config bug was fixed in the latest version)

PR #393, opened 2016-03-26
Fix shebang

Nephos:
python2 instead of python

### DIFF
--- a/start.py
+++ b/start.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 
 # Included modules
--- a/start.py
+++ b/start.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python2.7
 
 
 # Included modules
--- a/zeronet.py
+++ b/zeronet.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python2.7
 
 # Included modules
 import os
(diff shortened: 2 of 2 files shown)


HelloZeroNet:
It does not work on osx. according to http://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take `#!/usr/bin/env python` always should point to python2

Nephos:
> #!/usr/bin/env python always should point to python2

No it doesn't. This is why it doesn't work on my linux.

>  $>  /usr/bin/env python --version
> 
> Python 3.5.1


Furthermore, I will quote your source :+1: 

> ... python should be used in the shebang line only for scripts that are source compatible with both Python 2 and 3.

So please **fix** the shebang because it is not.

Nephos:
see: http://127.0.0.1:43110/1MyPktAWvYQQiLBNuApBZKUL2xNqQmAE6D/?Post:6

HelloZeroNet:
```
$ /usr/bin/env python2
/usr/bin/env: python2: No such file or directory
```
for me

Nephos:
I moved my old master branch to patch-shebang, i will reopen a new merge request later.

There is an issue on OSX (no python2 executable).

As OSX has not the standard python2 executable, OSX users should install a virtual env.

see: http://stackoverflow.com/questions/3713097/upgrading-python-in-virtual-env

HelloZeroNet:
please correct your PR then i will accept the modification

Nephos:
Tested on OSX. It seems that OSX has

- python
- python2.7
- python3.3
- python3

but not python2.

Waiting for you reply.

Nephos:
rebased

PR #413, opened 2016-04-08
Update to latest bitcoinrpc

TheNain38:


HelloZeroNet:
why? its the same

TheNain38:
@HelloZeroNet No, it handles the "null" case

HelloZeroNet:
i still dont see the difference, but ok

PR #424, opened 2016-04-18
Added "allow-same-origin" flag to iframe sandbox to allow for DASH videos

frerepoulet:
As per yesterday's IRC conversation. Since larger than 1MB files are not supported yet on ZeroNet, made a change for dash.js to work with ZeroNet's IFRAME. Otherwise the browser does not allow reading the video segment files.

### DIFF
--- a/src/Ui/template/wrapper.html
+++ b/src/Ui/template/wrapper.html
@@ -55,7 +55,7 @@ if (window.opener && window.stop) window.stop()
 
 
 <!-- Site Iframe -->
-<iframe src='about:blank' id='inner-iframe' sandbox="allow-forms allow-scripts allow-top-navigation allow-popups {sandbox_permissions}"></iframe>
+<iframe src='about:blank' id='inner-iframe' sandbox="allow-forms allow-scripts allow-top-navigation allow-popups allow-same-origin {sandbox_permissions}"></iframe>
 
 <!-- Site info -->
 <script>


HelloZeroNet:
This makes the ineffective sandboxing ineffective (required, because every site is served from same host: 127.0.0.1)

up4:
So, to be clear here, @HelloZeroNet (BTW, I work with @frerepoulet ). You do understand that the iframe blocking ZeroNet site access to Window.localStorage (required by both HLS and DASH), combined with the ZeroNet file size restriction effectively ban video content of any reasonable quality on your network ? Don't you feel compelled to provide alternative ideas at all ?

HelloZeroNet:
Disabling the same origin protection would allow any site to access or delete other site's data.

I think webrtc based p2p video is never going to be a reliable solution, because the peers only seeds it until they are on the site, but it could be possible to create a videoplayer plugin that allows you to display videos using DASH/WebTorrent outside of the iframe. (like the sidebar plugin)

up4:
If you would allow us to extend the ZeroFrame API with some kind of Window.localStorage emulation, then I'd be more than happy to fork dash.js to allow it to work with zeronet. I think we would also need to extend UiServer.fileGet for proper progress tracking and be able to display minimal loading status UI. But I would need some kind of reassurance that provided with good enough code you will make those changes to be default and not disabled optional plugins, if the changes are not made available to all users, they are pointless.

HelloZeroNet:
There is an api command to read/write storage: https://zeronet.readthedocs.org/en/latest/site_development/zeroframe_api_reference/#wrappergetlocalstorage
but i think you will run into other sandboxing limitations that makes it impossible

up4:
OK, great I didn't see that. Will try it and see. Do you already have a precise idea of the other sandboxing issues we should expect (given that with "allow-same-origin", it works) ?

HelloZeroNet:
I think this is required by webtorrent, but also disabled by the sandboxing: https://developer.mozilla.org/en-US/docs/Web/HTTP/data_URIs (not sure why is it disabled)

up4:
it might be required (btw, we are looking at Media Source Extensions for both DASH and HLS right now and not WebRTC) but it doesn't seem to be enabled by allow-same-origin which is the only flag we added

up4:
also, what do you feel about extending fileGet (or such) for download progress tracking ? or is there already something to that effect in place ?

HelloZeroNet:
files larger than 1MB is not supported yet, so dl tracking is not any useful

up4:
yes ok but: 1) they should and 2) we are having all sorts of issues with files >100KB  just as well so, if this can't be fixed at a lower level soon enough, something needs to be done for better I/O error reporting and progress UI

anyways, we'll see if we can have DASH work without allow-same-origin and be back on I/O UI later

HelloZeroNet:
Most of the users are using ZeroNet from Tor, which is not suitable to distributing big files, so using other network for this purpose is a better idea (eg. torrent). There is no streaming file support in zeronet which makes the progress reporting not possible yet.

frerepoulet:
just to confirm: data uris are indeed disabled by the iframe sandboxing and indeed break dash and hls... looks like zeronet will not work for our purpose but thank you for your cooperation in the matter

PR #470, opened 2016-05-26
Fix file permissions

Emeraude:
I noticed that the files `data/users.json` and `data/sites.json` have the permissions `644` which means everybody can read them, even if they contain private keys. So I changed these files' permissions to `600` to ensure that only their owner can read them (as ssh does for its private keys).
I also modified the function `atomicWrite` so that the file permissions are preserved (they were reset to `644` each time).

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -1,6 +1,7 @@
 # Included modules
 import os
 import sys
+import stat
 import time
 import logging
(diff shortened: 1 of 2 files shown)


HelloZeroNet:
Thanks!

PR #517, opened 2016-07-22
ZeroNet Sharing Protocol Transport Abstraction

up4:
Self-explanatory.

libzeronet will not support ipv4 (even through Tor exit nodes) nor BitTorrent trackers for hardened security and improved decentralization.

### DIFF
--- a/plugins/AnnounceZero/AnnounceZeroPlugin.py
+++ b/plugins/AnnounceZero/AnnounceZeroPlugin.py
@@ -5,6 +5,7 @@ from Plugin import PluginManager
 from Peer import Peer
 from util import helper
 from Crypt import CryptRsa
+from Config import config
 
 allow_reload = False  # No source reload supported in this plugin
 time_full_announced = {}  # Tracker address: Last announced all site to tracker
@@ -83,7 +84,10 @@ class SitePlugin(object):
             tracker_ip, tracker_port = tracker_address.split(":")
             tracker = Peer(tracker_ip, tracker_port, connection_server=self.connection_server)
             connection_pool[tracker_address] = tracker
+
+        print(request)
         res = tracker.request("announce", request)
+        print(res)
 
         if not res or "peers" not in res:
             self.log.debug("Announce to %s failed: %s" % (tracker_address, res))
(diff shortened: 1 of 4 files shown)


radfish:
I think you's also need to change the ZeroAnnounce to request only onion peers from the Boostrapper.

As is, the CLI argument naming and semantics are a bit hard to understand for somebody who did not read the code. I would do something like this, instead:
```
--transports {tor,ipv4}
Select the set of transports over which the node will communicate with trackers and peers.
Default {tor,ipv4}.
````
Then it is all simple to understand and use, and straightforward to extend to other transports, like I2P. The only subtlety is that with --tor=always *and* ipv4 in --transports, clearnet (ipv4) trackers will be accessed via Tor exit nodes. Intuitively, to prevent that, simply exclude ipv4 from --transports list, in which case only trackers with Tor addresses would be accessed (as of now, this only includes the zero://*.onion super-nodes that run the Bootstrapper plugin). As far as peers go, the behaviour is obvious.

The tracker *protocol* (zero:// aka ZeroAnnounce) should not be mixed into this. Also, the peer *protocols* (PEX) should not be mixed into this. Both, because the same protocol can return peer addresses in multiple transports. These are orthogonal concerns.

up4:
@radfish _Also change ZeroAnnounce_: I agree. I also agree with your `--transports {tor,ipv4}` proposal. But I would change "tor" for "onion" for "onion routing" so it excludes ipv4 via tor exit nodes. I would also change --pex-only to --zero-annouce only, however I am not 100% confident about this and I will need to read more on the whole ZeroAnnounce/Boostraper plugin bundle. But I think it is important to keep this pull request open to signify that a refactoring is in progress regarding that. Will be back with more in the next couple of days. Thanks.

up4:
So, I'm just fixing the current implementation towards benchmarking the siteDownload command the same way it is implemented in libzeronet (pex_only, onion_only). I talked to @shortcutme about merging the "pex" and "announce" commands but I will need more time to think about it and propose something that is backward compatible and will also simplify the code (both commands should be handled by the same code).

up4:
I am closing and re-opening this in its own branch.

PR #521, opened 2016-07-30
Virtual Files with Ui-Level Chunking & External Streams

up4:
Here is a solution for bigger files support and more: virtual files with chunking.

So in your content.json, you would add a "vfiles" section such as this one

"vfiles" : {
    "path/to/vfile1" : [
        "path/to/vfile1/chunk1",
        "path/to/vfile1/chunk2",
        "path/to/vfile1/chunk3",
        "path/to/vfile1/chunk4",
        "path/to/vfile1/chunk5"
    ],
    "path/to/vfileN" : [
        "path/to/vfileN/chunkA",
        "path/to/vfileN/chunkB",
        "path/to/vfileN/chunkC",
        "path/to/vfileN/chunkD",
        "path/to/vfileN/chunkE",
        "path/to/vfileN/chunkF"
    ]
}

This would allow for smaller files ("chunks") to be concatenated from the filesystem to an HTTP response stream by the Ui server without affecting the sharing protocol at all.

Eventually, this could be extended with external vfiles to allow the Ui file server to hide the complexity of installing proxies and additional software by supporting them internally. When an HTTP by the client is made to the Ui server for such a file, the Ui server must handle the request to the external system and stream back the file to the client via http.

"vfiles" : {
    "path/to/vfile2" : {
        uri : "http://sjdf893hiwjkehrod.onion/bigMovie.mp4"
        size : 12567773,
        sha256 : "jk98ey934hoi099u23u423"
    },
    "path/to/vfileZ" : {
        uri : "ipfs://QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG/bigMovie.mp4"
        size : 638236,
        sha256 : "ASDH=ASDB234234"
    }
}

Of course the ZeroFrame Javascript API should also be extended to allow for better progress reporting.


up4:
Hi all. I made a test site for that: 1ASN2GD7AMLcBwo6AViiEAf4KNX55Af6Ki

The HTML contains a HTML5 video element that requests an mp4 file for which there is a "vfiles" entry in the content.json. The array contains an ordered list of normal files that have been obtained with MP4Box with DASH-complient segments of the original MP4 file.

The video is in French and I provided French closed captions as well as English subtitles. Unfortunately, because of the ZeroNet Ui frame sandboxing, WebVTT tracks will not work. 

Now I will update my branch with the code to parse and concatenate all of this on the server-side. In the meantime, within the site, I provided a "concat.sh" script that will concatenate all the files to a single properly named file that will simulate what the server-side code will do for you.

up4:
I will test in-stream VTT files to see if they are loaded.

up4:
So I just committed the first functional "big file" support for ZeroNet. It works with `curl` but not from the browser. It's an HTTP thing. Would appreciate some help in debugging that if possible.

up4:
Got rid of the broken pipe error. But the content still doesn't load in the browser…

grez911:
I still don't understand why are you dividing one "big" file into chunks if their summary size can't be more than 10 MB. In what cases will it be usefull?

up4:
> I still don't understand why are you dividing one "big" file into chunks if their summary size can't be more than 10 MB. In what cases will it be useful? P.S. Are you trying to overcome 1 MB limit per file? Is this limit still in effect?

@grez911 "Still" meaning that we've talked about it before? If so, I don't recall. I've seen ZeroNet choking on files as low as 256K with current code base so obviously, supporting both bigger files _and_ chunking is an immediate, direct improvement. It is also backward compatible with the sharing protocol making it an elegant change too. Also, you can include optional file chunks in the list and they will all get transparently requested to peers with better load balancing between them when http-requesting the concatenated file. And my plan in the future for virtual files is to also support proxy files that can be streamed by the ZeroNet UI server from IPFS, WebTorrent and such without the need for the user to support external deamons.

up4:
I just merged the merger site feature into this branch so it's in sync with the master branch.

up4:
I will pause development of this feature until I hear from Tamas on what his intentions are with eventually pulling it.

HelloZeroNet:
I think the best would be a separate plugin, so the development could be done outside of the ZeroNet core. (I also creating new features as a separate plugin to keep the core simple)

up4:
I don't mind refactoring this into a plugin if you tell me it is possible. But before I do that, I would like to know if you consider this feature worthy of pulling the code in, albeit as a plugin. Also can it be enabled by default ? Thanks.

HelloZeroNet:
If the plugin becomes mature enough, then it's possible to enable by default, but an appstore-like plugin interface is also planned where anyone can enable/disable/install plugins easily.

up4:
If «maturity» is what you are looking for, then I'm going to concentrate on libzeronet and the first application I'm going to build on top of it. When this app is available on the Apple macOS App Store, user will be able to create content where ZeroNet and BitTorrent will be used as storages back-ends. ZeroNet for the Core Data-compliant SQLite3 metadata referring to BitTorrent large file blobs. Until virtual files are pulled in, Python Ui users will only be able to see static JPEG thumbnails of the torrent blobs in the HTML Ui generated by the Cocoa browser. Then, I guess, the feature will be sufficiently mature to implement in Python. But it will take a couple of months at least.

imachug:
@HelloZeroNet Does it still make sense now when we have inbuilt big file support?

HelloZeroNet:
Probably it would, but I think it should be done on application level.

PR #585, opened 2016-09-18
Update Dockerfile base image to ubuntu 16.04 LTS and add Tor support to it.

aitorpazos:
In order to keep the Dockerfile simple, needed opotions have been appended to package provided torrc file and the /etc/init.d/tor script is used to start tor. If further customisation is desired, a wrapper script should be created to have further control on the startup. This would allow things like specifying external (tor) proxies and disable tor startup in those situations.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
 
 MAINTAINER Felix Imobersteg <felix@whatwedo.ch>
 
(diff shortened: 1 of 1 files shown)


casept:
tor probably shouldn't be run in the same container as zeronet.

aitorpazos:
I understand your points. However, this change would improve current situation from not having Tor protection at all to at least have some of it.

HelloZeroNet:
Lots of VPS providers makes your account suspended for running Tor, so I think we should not start it by default.

aitorpazos:
Ok, that's a more compelling argument right now. I'll close this pull request and I'll make tor optional (disabled by default)

PR #587, opened 2016-09-19
Update Dockerfile base image to ubuntu 16.04 LTS and add Tor support to it (disabled by default).

aitorpazos:
The base Ubuntu image has been upgraded to the current LTS release (16.04) and tor support has been added to the Docker image. Tor is disabled by default as it may not be allowed by different hosting providers. In order to enable it, just set `ENABLE_TOR` environment variable to `true` when executing `docker run`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,8 +21,11 @@ RUN \
 ADD . /root
 VOLUME /root/data
 
+#Control if Tor proxy is started
+ENV ENABLE_TOR false
+
 #Set upstart command
-CMD cd /root && /etc/init.d/tor start && python zeronet.py --ui_ip 0.0.0.0
+CMD cd /root && (! ${ENABLE_TOR} || /etc/init.d/tor start) && python zeronet.py --ui_ip 0.0.0.0
 
 #Expose ports
 EXPOSE 43110
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
 
 MAINTAINER Felix Imobersteg <felix@whatwedo.ch>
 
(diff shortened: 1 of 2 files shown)


HelloZeroNet:
Thanks, but probably it would be better if the tor installation would be also based on `ENABLE_TOR` (tor binary starts automatically after the setup and vps providers looks for that)

aitorpazos:
That would mean that we need two separate images, one with tor installed and another one without tor.

However, with this change the tor binary would only run if the user explicitly sets the environment variable ENABLE_TOR. The only way for a VPS provider to know that tor exists in one of these containers is to inspect the image looking for the tor binary, which I don't think they would do.

*Using default config:*
```
$ docker run -d -p 15443:15441 -p 43112:43110 aitorpazos:myZeroNet
b1146f0409083a03d122cbe947cf9011a63cf62b9975870e61654778e8c6e307
$ docker exec -ti trusting_kirch ps -ef
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 21:58 ?        00:00:00 /bin/sh -c cd /root && (! ${ENABLE_TOR} || /etc/init.d/tor start) && python zeronet.py --ui_ip 0.0.0.0
root         6     1  5 21:58 ?        00:00:00 python zeronet.py --ui_ip 0.0.0.0
root        20     0  0 21:58 ?        00:00:00 ps -ef
```
*Using ENABLE_TOR=false:*
```
$docker run -e "ENABLE_TOR=false" -d -p 15442:15441 -p 43111:43110 aitorpazos:myZeroNet
aac074ccefe84379e19f54fa703c605ffbe667002e771c293916d9e8c7431e7c
$ docker exec -ti small_colden ps -ef
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 21:53 ?        00:00:00 /bin/sh -c cd /root && (! ${ENABLE_TOR} || /etc/init.d/tor start) && python zeronet.py --ui_ip 0.0.0.0
root         6     1  2 21:53 ?        00:00:00 python zeronet.py --ui_ip 0.0.0.0
root        20     0  0 21:53 ?        00:00:00 ps -ef
```
*Using ENABLE_TOR=true:*
```
$ docker run -e "ENABLE_TOR=true" -d -p 15444:15441 -p 43113:43110 aitorpazos:myZeroNet
19d06f2e9f65ae778b6666109c09dd5022ef83ed7bedbe4f2eed23f51e51891b
$ docker exec -ti jolly_stonebraker ps -ef
UID        PID  PPID  C STIME TTY          TIME CMD
root         1     0  0 21:59 ?        00:00:00 /bin/sh -c cd /root && (! ${ENAB
debian-+    17     1  8 21:59 ?        00:00:00 /usr/bin/tor --defaults-torrc /u
root        18     1  6 21:59 ?        00:00:00 python zeronet.py --ui_ip 0.0.0.
root        32     0  0 21:59 ?        00:00:00 ps -ef
```

HelloZeroNet:
Thanks, looking nice!

PR #588, opened 2016-09-20
Update UiRequest.py

dldx:
Fixed a bug where a lack of a trailing slash in every url caused zeronet to fail to load the page, even if it existed.

For example, if you go to http://127.0.0.1:43110/1CiDoBP8RiWziqiBGEd8tQMy66A6fmnw2V/big/docs, the page fails to load. However, if you go to http://127.0.0.1:43110/1CiDoBP8RiWziqiBGEd8tQMy66A6fmnw2V/big/docs/, it loads fine.

As far as I can tell, everything now works with this patch but it seems a little inelegant. When I got to such a url, the screen flashes for a second before redirecting. It would be better if it were instantaneous.


HelloZeroNet:
Thanks!

PR #602, opened 2016-10-04
[WIP] Add I2P support

### LINKED ISSUE #45: Use torrents over I2P for anonymity
I2P has bittorrent support, and it works very well (but slow). So suggestion is to use it. There should be an option to use only I2P, only clearnet, or both.

More info:
https://geti2p.net/en/
https://geti2p.net/en/docs/applications/bittorrent

str4d:
This implements support for both connecting to I2P torrent trackers, and communicating with peers over I2P.

Closes #45

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ ENV HOME /root
 RUN \
     apt-get update -y; \
     apt-get -y install msgpack-python python-gevent python-pip python-dev; \
-    pip install msgpack-python --upgrade; \
+    pip install -r requirements.txt --upgrade; \
     apt-get clean -y; \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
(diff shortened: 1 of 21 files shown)


str4d:
This currently does *not* work fully, because [i2p.socket](https://github.com/majestrate/i2p.socket) (which this uses) doesn't listen properly under gevent. This means that ZeroNet can connect to I2P torrent trackers, and it can connect to peers, but it can't successfully listen for peer connections.

Also, the above was true at the end of July when this was based on 523a7d4c16e8c2f590e759e23f0d2d6ae36cff37. I have just rebased this onto master, and have not tested it again, so if there were Tor-related changes in the intervening time, this PR does not currently reflect them (aside from the PEX limit change I spotted).

Plasmmer:
@str4d, do you have news?

asdedhj:
Aye, any updates?

PowerPress:
Is I2P working now?

rex4539:
@str4d Could you get back on this as part of your Friday fun projects?

ghost (approved):

ghost (approved):

PR #607, opened 2016-10-11
service should listen only at localhost by default (docker)

gyulaweber:
If we expose the service to all interfaces (with 43110:43110 port mapping), it'll be reachable from anywhere (from where the box can be reached), and may impersonate the owner of the service

Moreover, the firewall setup is not straightforward for beginners, as docker sets up NAT rules to expose the services, so it cannot be just dropped in the filter/INPUT chain.


HelloZeroNet:
Thanks!

PR #609, opened 2016-10-12
allow private ip ranges in socks proxy

majestrate:
this adds checks for 127.0.0.0/8 and 10.0.0.0/8 and treats them as directly connectable when using socks proxy.

### DIFF
--- a/src/Test/TestIP.py
+++ b/src/Test/TestIP.py
@@ -15,3 +15,5 @@ class TestIP:
         assert not is_private_address('127.a.15.1')
         assert not is_private_address('10.10.15.asd')
         assert not is_private_address("google.com")
+        assert not is_private_address("10.0x10.0.1")
+        assert not is_private_address("127.\x00.0.1")
--- a/src/Test/TestIP.py
+++ b/src/Test/TestIP.py
@@ -1,0 +1,17 @@
+import pytest
+
+from util.IP import is_private_address
+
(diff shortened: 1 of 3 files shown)


AceLewis:
```
test_addresses = ["127.0.0.1", "128.0.0.1", "10.0.0.1", "127.34354.0.1"]
for address in test_addresses:
    print("IP address {:<13} returns :{}".format(address, is_private_address(address)))
```
Returns
```
IP address 127.0.0.1     returns :True
IP address 128.0.0.1     returns :False
IP address 10.0.0.1      returns :True
IP address 127.34354.0.1 returns :True
```
The last one should not return True, see http://stackoverflow.com/a/11264379 for more info.

HelloZeroNet:
Can you please specify the usecase? I think if someone specify a proxy (eg. tor) it means he/she wants every connection to use that. Even a local network connection could lead loss of privacy.

majestrate:
it's related to i2p support, i run my i2p router's sam interface on
10.0.3.1 so if zeronet wants to use tor and i2p then you'd need a way to
bypass the module level hook somehow for i2p connections.
cc @str4d

HelloZeroNet:
Then probably it would be better to create direct connection only to 127.0.0.1 and specified proxy (tor or sam) ips

majestrate:
that is a good point, that would require an entirely different PR.

PR #619, opened 2016-10-19
Add allow-same-origin to sandbox iframe

ceritium:
Allow access to methods of `window.parent.location` like `hash` or `query`.

### DIFF
--- a/src/Ui/template/wrapper.html
+++ b/src/Ui/template/wrapper.html
@@ -55,7 +55,7 @@ if (window.opener && window.stop) window.stop()
 
 
 <!-- Site Iframe -->
-<iframe src='about:blank' id='inner-iframe' sandbox="allow-forms allow-scripts allow-top-navigation allow-popups {sandbox_permissions}"></iframe>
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
every zeronet site is running from the same domain (127.0.0.1), so it would be a security issue to allow same origin access, but you can modify the hash/query using `window.top.location = "?something#hello2"`
or wrapperPushState/wrapperReplaceState zeroframe api commands

ceritium:
You are right, it is a security issue.

As you said I can modify the hash/query with `window.top.location = "?something#hello2"` but I am not able to read the value of hash/query.

screen shot 2016-10-19 at 23 36 32


What do you suggest?

HelloZeroNet:
you able to reach the get parameters directly, using window.location.search (you need to remove `wrapper_none=.*`)
if you call `Page.cmd("innerLoaded")` command the hash will be applied to the inner frame, so you can reach it using window.location.hash
```
> window.location.search
< "something&wrapper_nonce=22f998ea92f1e2574c72cdc96ba7d479d9c1aec8fa31e3873047cbb61987f784"
> window.location.hash
< ""
> Page.cmd("innerLoaded")
> window.location.hash
< "#hello"
```

ceritium:
Thank you very much for the answers: D

PR #628, opened 2016-11-04
Prevent portchecker on tor == always

mishfit:
Perhaps it's just me, but is there a point to testing for the port before checking for tor settings?


HelloZeroNet:
You have to know your exit node ip to avoid re-connect to yourself and it's also allow to put yourself on the map (allows you to verify Tor is working)

PR #646, opened 2016-11-19
Add Chinese (Simplified) Translate

l5h5t7:


HelloZeroNet:
Thanks!

HelloZeroNet:
Please also translate these files, so I can also add it to ZeroHello:

https://github.com/HelloZeroNet/ZeroNet/blob/master/src/Translate/languages/hu.json
https://github.com/HelloZeroNet/ZeroNet/blob/master/plugins/Sidebar/languages/hu.json
https://github.com/HelloZeroNet/ZeroHello/blob/master/languages/hu.json
https://github.com/HelloZeroNet/ZeroMe/blob/master/languages/hu.json
https://github.com/HelloZeroNet/ZeroTalk/blob/master/languages/hu.json

Simply make a copy of them to the same dir where they are with the language short code you are going to translate to and replace the Hungarian translation.

To test them before submitting to github:
 - By default ZeroNet uses the operating system language, but you can also specify it in zeronet.conf (where zeronet.py is)
 - Open sidebar, check "This is my site"
 - Open javascript console (F12) check "Disable cache" on Network tab to avoid caching, keep window open
 - To reload Core language files (src/Translate/languages/hu.json and plugins/Sidebar/languages/hu.json) without restart you have to click on you language again on ZeroHello's 3dot menu

PR #655, opened 2016-11-19
Fix KeyError Exception when a site without "title" in content.json

linkerlin:
Fix KeyError Exception when a site without "title" in content.json


linkerlin:
If a site set it's title to None, a KeyError will raise.

HelloZeroNet:
If there is not title, then it can't be displayed in the list, so it's a requirement.

PR #669, opened 2016-11-24
Russian translation for sidebar

iShift:


HelloZeroNet:
some titles are overflow, fixed some in the latest revision, but there is some I cant

iShift:
Ok,thanks!

PR #687, opened 2016-12-01
get working dir from python script's real path

rcmorano:
since, for example in the docker container, it was trying to grab the dirname from the 'zeronet.py' relative path which has no path


HelloZeroNet:
Thanks!

PR #691, opened 2016-12-03
Translations for more plugins.

ysc3839:
I want to know if I should translate the `self.response` words? Such as `self.response(to, {"error": "Not a merger site"})`.


HelloZeroNet:
Thanks!
I think we should not translate network/log messages

PR #703, opened 2016-12-14
Just with this change it's possible to start it with absolute path

cryptid11:
like `python ~/Documents/ZeroNet-master/zeronet.py #startzeronet`. Anyway I don't test this in windows and maybe it's problematic there...
Anyway, maybe somebody know if it works in windows, or a more elegant way to get the same result.

### DIFF
--- a/src/Plugin/PluginManager.py
+++ b/src/Plugin/PluginManager.py
@@ -2,6 +2,7 @@ import logging
 import os
 import sys
 from collections import defaultdict
+from os.path import dirname, abspath
 
 from Debug import Debug
 from Config import config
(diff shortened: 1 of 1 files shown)


cryptid11:
Uh, I just discover the existence of Travis CI, so it seems it works even on windows :dancer:  :dancer:  :dancer:

HelloZeroNet:
I dont think its necessary, because we change the working path to zeronet's directory on startup: https://github.com/HelloZeroNet/ZeroNet/blob/master/zeronet.py#L14

cryptid11:
Uhg :/ yes, I got confused, at least I learn is possible something like that (idk of os.chdir), sorry for lost time....

PR #705, opened 2016-12-15
Update README.md to include work-around for LSOpenURLsWithRole() errors

justinwiley:
I downloaded the Mac bundle, unzipped, moved into directory.

I ran `open ZeroNet.app`.  An application icon briefly appeared on the dock, and then disappeared.

LSOpenURLsWithRole() failed with error -10810 for the file /Users/ju/Downloads/tmp/ZeroBundle/ZeroNet.app.

After googling around, it turns out you have to make the file executable via:

`chmod +x ZeroNet.app`

It would probably help others and save time to have this in the docs.


HelloZeroNet:
Thanks, probably it's depends on what unpacker application you have used.
It would be nice if someone could help me how to make a proper .dmg from it.

Issue #298, opened 2016-03-03
Please accept IDN .bit Names


HamishMacEwan:
Hi,

Truly amazed at what you have achieved and the things it can do right now and in the future.  Could you give some consideration to implementing acceptance of IDN .bit names,

ɥsıɯɐɥ.bit

for example.  

http://127.0.0.1:43110/ɥsıɯɐɥ.bit responds with:

```
Err: UnicodeDecodeError: 'ascii' codec can't decode byte 0xc9 in position 1: ordinal not in range(128) in UiServer.py line 76 > UiRequest.py line 90 > UiRequest.py line 498
```

Thanks again for returning control to the edge, peer to peer, autonomous networks, so cool.

Hamish.

HelloZeroNet:
I'm not fan of non-ascii domain names: they hard to type and possible phishing attacks via similar characters

HamishMacEwan:
That is a poor excuse.

If ZeroNet is to serve a global audience it will need to do better than Morse code in terms of character set.

They are not hard to type by people who use them, and there are sufficient phishing opportunities in 1l O0 etc and indeed people don't even need similarity to be fooled.

You might also like to consider how squatted NMC names are, opening up the full Unicode space may well defeat that.

To complain about phishing and unicode while exploiting torrent and bitcoin protocols and their attendant criminality is lame.

If the goal of ZeroNet was to give users choice, to deny implementation of a service NMC already offers is contrary to that ethos and disappointing in the extreme.

sinkuu:
ASCII homograph attack can be prevented if people are careful, while IDN homograph attack cannot however careful they are :(
I don't know how to distinguish `сор.bit` from `cop.bit`, except try googling URL before clicking.

Anyway if ZeroNet will support IDN, it needs at least blacklist like this: http://kb.mozillazine.org/Network.IDN.blacklist_chars

HamishMacEwan:
I'm sure you can distinguish xn--n1aeb.bit from cop.bit, and since ZeroNet doesn't display the unicode version on http://127.0.0.1:43110/1Name2NXVi1RDPDgf5617UoW7xA6YrhM9F there isn't much opportunity.

Given the much easier and effective phishing attacks, taking a position of protection against this kind of esoteric quirk is patronising.  I'm not saying it can't be done, but attempting to guard against such attacks by denying choice isn't worth the effort.

sinkuu:
> since ZeroNet doesn't display the unicode version

[...]

Issue #131, opened 2015-06-27
Allow for unicode characters in domain names.


bloff:
http://127.0.0.1:43110/lýc.bit/

Fails with python exception message:
Err: UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 2: ordinal not in range(128) in UiServer.py line 71 > UiRequest.py line 64 > UiRequest.py line 389

Just curious: any plans to move to python 3? Any library dependencies which are not available there?

bloff:
There is also an error with the encoded domain-name form:
http://127.0.0.1:43110/xn--lc-0ka.bit

Fails with python exception message:
Err: AttributeError: 'NoneType' object has no attribute 'group' in UiServer.py line 71 > UiRequest.py line 56 > UiRequest.py line 148 > UiRequestPlugin.py line 20 > UiRequest.py line 237 > UiRequestPlugin.py line 32

bloff:
Also worth saying, when I goto http://zero/xn--lc-0ka.bit (using zeronet chrome extension), I get a weird message:
Not Found: data/xn--lc-0ka.bit/index.html

I already tried to actually create such a file in my website, to no avail.

HelloZeroNet:
Thanks for reporting, we going to check the possibilities. Utf-8 domain names [not supported in Namecoin](http://wiki.namecoin.info/?title=Domain_Name_Specification_2.0#Valid_Domains), the domain-name encoding could be a workaround for this problem.

Python3 is not planned yet, because most of linux distribution still comes with python2 and the gevent liblary is not supported in Python3.

Idealcoder:
I'd like to point out a security concern with allowing UTF8 domain names. It would make it very easy to create a spoof version of a website for phishing, and use a very similar looking domain address with a different character. So I'm not sure if it is a good idea.

deavmi:
@Idealcoder True.

deavmi:
Keep things simple. Are there any domains on the DNS (not namecoin's DNS) that have Unicode characters other than the standard english ones in their domain name?

HamishMacEwan:
IDN and new gTLDs have meant the era of ASCII only domain names in the DNS is over, except that they use punycode which is ASCII.  These are the xn-- strings appearing from the OP.

[...]

PR #735, opened 2017-01-04
Fix encoding issues + IDNA support

MuxZeroNet:
Please test it on your own operating system and on your own locale.

### DIFF
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -242,8 +242,14 @@ class UiRequest(object):
                 idna_encoded = raw_domain.encode('idna')
                 if idna_encoded != raw_domain:
                     redirect_to = '/' + idna_encoded + the_rest
+                    if self.env.get('QUERY_STRING'):
+                        redirect_to += '?' + self.env.get('QUERY_STRING')
+
                     self.log.debug("IDNA domain: " + idna_encoded)
-                    return self.actionRedirect(redirect_to)
+                    self.log.debug('Location repr: ' + repr(redirect_to))
+                    return self.actionRedirect(str(redirect_to)) # gevent bug
+                    # https://github.com/socketio/socket.io/issues/2446
+                    # https://github.com/ajenti/ajenti/issues/838
 
             # Otherwise, Bad url
             return False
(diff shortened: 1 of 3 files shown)


src/Ui/UiRequest.py
```diff
@@ -40,11 +41,17 @@ def __init__(self, server, get, env, start_response):
...
+
     # Call the request handler function base on path
     def route(self, path):
         if config.ui_restrict and self.env['REMOTE_ADDR'] not in config.ui_restrict:  # Restict Ui access by ip
             return self.error403(details=False)
 
+        self.log.debug("-- path repr: " + repr(path))
+        path = path.decode('utf-8')
+        self.log.debug("-- path decoded repr: " + repr(path))
+
```
MuxZeroNet: If you are annoyed by encoding issues of Python 2, try using `repr(...)` on strings when printing out debug log.

MuxZeroNet:
Solves issues #131 and #298 but has homograph attacks. It may be a good idea to show a warning to show the encoded domain name, the domain public key, and whether known homographs are found in the domain name.

MuxZeroNet:
I decided not to make ZeroNet support IDNA domain names simply because it will make browsing IDNA sites more inconvenient.

Suppose someone registered for a fancy domain name `☉net.bit` (the Sun symbol), and a bad guy registered for a domain name `⊙net.bit` (the circled dot operator) for phishing.
```python
>>> u'☉' == u'⊙'
False
>>> print repr(u'☉'),repr(u'⊙')
u'\u2609' u'\u2299'
```

Most fonts are not optimized for differentiating these homographs, so we want ZeroNet to display them in puny coded form.
```python
>>> u'⊙net.bit'.encode('idna'), u'☉net.bit'.encode('idna')
('xn--net-vr2a.bit', 'xn--net-gn5a.bit')
```
To not to get to the phishing site, one has to memorize the full puny coded form of the innocent site's domain name. In this case, one will choose to use public keys to access these sites instead.

I will create a separate pull request only to fix encoding issues.

HelloZeroNet:
Yeah, I'm also uncertain about utf8 domain names, no one can tell the difference between Оnet, ⵔnet, Onet, Οnet or Onet (All uses different "0" character)

MuxZeroNet:
Closed. IDNA domains will not be supported due to security concerns. Discussion about string encoding bugs has been moved here: #765

Issue #742, opened 2017-01-06
ZeroNet still seeds a site when no inner iframe was not loaded.


MuxZeroNet:

MuxZeroNet:
Attacks works in Firefox. 
```
Host: 127.0.0.1:43110
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:50.0) Gecko/20100101 Firefox/50.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.8,zh-CN;q=0.5,zh;q=0.3
Accept-Encoding: gzip, deflate
Origin: null
DNT: 1
Connection: keep-alive
```
WTF? AJAX without referrer header, and X-Requested-With header.

MuxZeroNet:
WebSocket requests also have a "valid" Accept header:
```
Host: 127.0.0.1:43110
User-Agent: Mozilla/5.0 [...] Gecko/20100101 Firefox/50.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Sec-WebSocket-Version: 13
origin: http://127.0.0.1:43110
Connection: keep-alive, Upgrade
Upgrade: websocket
[Irrelevant headers omitted]
```

MuxZeroNet:
OPTIONS requests made by Ajax and Fetch APIs need some attention:
```js
var headers = new Headers();
var options = { method: 'GET', headers: headers };

headers.set('Accept', '*/*,fake/mime');
headers.set('X-Requested-With', 'FakeXMLHttpRequest');

fetch('/1CiDoBP8RiWziqiBGEd8tQMy66A6fmnw2V/big/sample-loader/files/list.json?fetch', options)
.then(function(response) {
  console.log("Done.");
});
```

```
OPTIONS /1CiDoBP8RiWziqiBGEd8tQMy66A6fmnw2V/big/sample-loader/files/list.json?fetch
Access-Control-Request-Method: GET
Access-Control-Request-Headers: x-requested-with
```

MuxZeroNet:
Finally, some proposals:
- Cross origin AJAX, WebSocket and Fetch requests have `origin` headers, but same origin requests can also have `origin` headers. Blocking all requests with `origin` headers will break same origin AJAX requests, but it is useful when AJAX request is strictly disallowed.
[...]

PR #743, opened 2017-01-06
Add X-Frame-Options header

### LINKED ISSUE #711: ZeroNet allows for arbitrary sites to be added to the client
ZeroNet allows for arbitrary websites to added to the client at random, without authentication or any checking that the user would actually want this.

For example, a website (on ZeroNet or elsewhere) could embed hidden image tags like this:
``

This triggers the user's client to automatically add the site.

I think there needs to be some sort of confirmation process that happens.

MuxZeroNet:
**Partially** fixed #711

Save this file below as `zeronet.html`:
```html
Escape and Seed
<iframe id="an-iframe" src="about:blank"></iframe>
<script>
document.getElementById("an-iframe").src = "http://127.0.0.1:43110/13Lo9wpxbogHdaTQ9rGZruu9MJZnigzU3B/"; // EmojiDay site
// or any site that haven't benn saved on your computer
</script>
```
Start `python -m SimpleHTTPServer` and go to `http://127.0.0.1:8000/zeronet.html`
The script inside the `iframe` will **escape** and redirect you to seed EmojiDay site.

With `X-Frame-Options: SAMEORIGIN` header, framing will be denied by the browser.
```
GET http://127.0.0.1:8000/zeronet.html
GET http://127.0.0.1:43110/13Lo9wpxbogHdaTQ9rGZruu9MJZnigzU3B/

Load denied by X-Frame-Options:
http://127.0.0.1:43110/13Lo9wpxbogHdaTQ9rGZruu9MJZnigzU3B/ does not permit cross-origin framing.
```

However, ZeroNet **still seeds** that site even when the inner iframe is not loaded by the browser. (Issue #742)

### DIFF
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -144,6 +144,7 @@ class UiRequest(object):
         headers.append(("Keep-Alive", "max=25, timeout=30"))
         if content_type != "text/html":
             headers.append(("Access-Control-Allow-Origin", "*"))  # Allow json access on non-html files
+        headers.append(("X-Frame-Options", "SAMEORIGIN"))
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
Thanks!

PR #748, opened 2017-01-08
Fix Unhandled Exception SOCKS5Error

MuxZeroNet:
When ZeroNet tries to route a LAN address through Tor's SOCKS5 proxy, Tor refuses to do so and sends back an error message. This error message should be handled.


HelloZeroNet:
upnpopening is disabled in tor mode and why is it better raising IGDError instead of SOCKS5Error (I think it would be usefull to keep UpnpPunch.py standalone from other libs)

DaniellMesquita:
@tangdou1

Can this pull request be closed?

PR #750, opened 2017-01-08
Detect suspicious Ajax requests

MuxZeroNet:


l5h5t7:
I just wonder if it affects public proxy?

MuxZeroNet:
Interesting, but this issue will be approached differently.

DaniellMesquita:
@HelloZeroNet @shortcutme

PR #751, opened 2017-01-09
Only allow GET HEAD OPTIONS

MuxZeroNet:
Only allow GET HEAD OPTIONS to load wrapper and site media


HelloZeroNet:
I think a `if self.env['REQUEST_METHOD'] != 'GET':` restriction will be enough for the `actionWrapper` (we catch OPTIONS earlier and POST to non-wrapper files are ok.

Issue #725, opened 2016-12-29
FileServer Trying to open port using UpnpPunch... - Unhandled exception


kasv:
$ pwd
~/downloads/ZeroBundle/ZeroNet/src/util

$ ~/downloads/ZeroBundle/Python/python ./UpnpPunch.py
Opening port...
DEBUG:root:Trying to open port 15443.
DEBUG:root:Found local ips: ['192.168.0.8']
DEBUG:root:Trying using local ip: 192.168.0.8
Traceback (most recent call last):
  File "./UpnpPunch.py", line 342, in <module>
    print ask_to_open_port(15443, "ZeroNet", retries=3, protos=["TCP"])
  File "./UpnpPunch.py", line 320, in ask_to_open_port
    protos=protos)
  File "./UpnpPunch.py", line 299, in _communicate_with_igd
    _orchestrate_soap_request(local_ip, port, fn, desc, protos)
  File "./UpnpPunch.py", line 274, in _orchestrate_soap_request
    idg_data = _collect_idg_data(ip)
  File "./UpnpPunch.py", line 258, in _collect_idg_data
    _retrieve_igd_profile(idg_data['location']))
  File "./UpnpPunch.py", line 104, in _parse_igd_profile
    dom = parseString(profile_xml)
  File "/home/kasv/downloads/ZeroBundle/lib/python2.7/xml/dom/minidom.py", line 1928, in parseString
    return expatbuilder.parseString(string)
  File "/home/kasv/downloads/ZeroBundle/lib/python2.7/xml/dom/expatbuilder.py", line 940, in parseString
    return builder.parseString(string)
  File "/home/kasv/downloads/ZeroBundle/lib/python2.7/xml/dom/expatbuilder.py", line 223, in parseString
    parser.Parse(string, True)
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 2, column 442

See attachment
[answer.xml.txt](https://github.com/HelloZeroNet/ZeroNet/files/677114/answer.xml.txt)

HelloZeroNet:
cc @sirMackk

sirMackk:
Super big thanks for attaching the SOAP response @kasv . I'll look into this in the next few days.

sirMackk:
[...]

PR #757, opened 2017-01-11
Decodes IGD response to utf-8; Makes function arguments immutable;

sirMackk:
This is a very likely fix for #725. Didn't get a confirmation yet, but I wanted to have this PR ready now.

I made the ol' mistake where I thought that urlopen().read() returns a text string, but it's a byte string, and `xml.minidom.parseString` expects a text string. Easily solved by decoding the byte string into a text string.

The other small change is to make the default function arguments that are _mutable_ lists into immutable tuples. Using a mutable object as a default function argument is an invitation to all sorts of pain.


HelloZeroNet:
are the `,` necessary at the `('TCP', 'UDP',)` ?

```
>>> ("TCP", "UDP") == ("TCP", "UDP",)
True
```

sirMackk:
Absolutely not! It's just a style thing - in my opinion, the `,` at the end instantly identifies the parenthesis as tuple. But if you don't like it, I'd be more than happy to get rid of the `,`'s :).

HelloZeroNet:
For me it's looks like an unfinished function call, so i would prefer without it. (I know its necessary for tuples with only one element)

sirMackk:
Thanks for the feedback, I never thought of it that way. I removed the commas and rebased my branch.

PR #767, opened 2017-01-13
Prevent torrc being overwritten after updates.

ysc3839:


HelloZeroNet:
By that we would loose the ability to update torrc later, so instead I would prefer configurable `tor_exe`, so if anyone wants to make modifications, then create a copy of the directory and change that settings.

Issue #627, opened 2016-11-04
Set stack size to let zeronet run on router


cxgreat2014:
RT-AC54U /opt/home/admin/ZeroNet-master]# opkg install libopenssl
Upgrading libopenssl on root from 1.0.2h-1 to 1.0.2j-1...
Downloading http://pkg.entware.net/binaries/mipsel/libopenssl_1.0.2j-1_mipselsf.ipk.
Configuring libopenssl.
[RT-AC54U /opt/home/admin/ZeroNet-master]# opkg install python-openssl
Upgrading python-openssl on root from 2.7.11-9 to 2.7.12-2...
Downloading http://pkg.entware.net/binaries/mipsel/python-openssl_2.7.12-2_mipselsf.ipk.
Configuring python-openssl.
[RT-AC54U /opt/home/admin/ZeroNet-master]# ./zeronet.py
- Starting ZeroNet...
[23:49:28] - OpenSSL load failed: File not found, falling back to slow bitcoin verify
[23:49:30] - Version: 0.4.1 r1536, Python 2.7.11 (default, Aug 15 2016, 19:04:16)
[GCC 5.3.0], Gevent: 1.1.2
[23:49:35] - Creating FileServer....
[23:49:36] TorManager Tor controller connect error: error: can't start new thread in TorManager.py line 154 > _socket2.py line 218 > _socketcommon.py line 268 > resolver_thread.py line 64 > pool.py line 300 > threadpool.py line 175 > threadpool.py line 136 > threadpool.py line 116 > threadpool.py line 145
[23:49:36] - Creating UiServer....
[23:49:36] SiteManager Deleting orphan site from content.db: 1Name2NXVi1RDPDgf5617UoW7xA6YrhM9F
Traceback (most recent call last):
  File "./zeronet.py", line 16, in main
    main.start()
  File "./src/main.py", line 420, in start
    actions.call(config.action, action_kwargs)
  File "./src/main.py", line 120, in call
    func(**kwargs)
  File "./src/main.py", line 130, in main
    ui_server = UiServer()
  File "./src/Ui/UiServer.py", line 61, in __init__
    self.sites = SiteManager.site_manager.list()
  File "./src/Site/SiteManager.py", line 130, in list
    self.load()
  File "plugins/Zeroname/SiteManagerPlugin.py", line 17, in load
    super(SiteManagerPlugin, self).load(*args, **kwargs)
[...]

PR #779, opened 2017-01-19
set stack_size for router #627

cxgreat2014:
#627

### DIFF
--- a/zeronet.py
+++ b/zeronet.py
@@ -3,6 +3,9 @@
 # Included modules
 import os
 import sys
+# For router issue #627
+from thread import stack_size
+stack_size(32768)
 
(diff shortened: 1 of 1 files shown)


cxgreat2014:
Um..If it can't start thread, it will show error to me, but works fine...

cxgreat2014:
image
without any error
image

ysc3839:
I mean if the thread works well on current OS. It's no need to change the stack size.

cxgreat2014:
Yes, you are right...But how to check it?

ysc3839:
#780

cxgreat2014:
better solution #780

PR #780, opened 2017-01-19
Fix thread stack size on some embedded devices.

ysc3839:
Code from: https://github.com/wishinlife/SyncY/blob/master/syncy.py#L337-L355

### DIFF
--- a/src/Config.py
+++ b/src/Config.py
@@ -60,6 +60,26 @@ class Config(object):
         else:
             fix_float_decimals = False
 
+        import thread
+        stack_size = 32768
+        while True:
+            try:
+                thread.stack_size(stack_size)
+                thread.start_new_thread(lambda: 0, ())
+                break
+            except thread.error:
+                thread.stack_size(0)
+                break
+            except RuntimeError:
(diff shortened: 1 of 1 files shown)


ysc3839:
I don't know if it's good to check it default. I think we should add a option.

src/Config.py
```diff
@@ -60,6 +60,26 @@ def createArguments(self):
...
+                break
+            except thread.error:
+                thread.stack_size(0)
+                break
+            except RuntimeError:
+                thread.stack_size(0)
+                break
+            except ValueError:
+                if stack_size < 512 * 1024:
+                    stack_size *= 2
```
cxgreat2014: ？？？乘了2之后没有使用这个值呀，有啥意义？

cxgreat2014: 哦哦哦，while True，之后进行下一次了，手机看代码没看全，sorry

ysc3839:
This code is from [SyncY](https://github.com/wishinlife/SyncY), which is licensed under an 'unopen' license. I don't know if it's good to use it's code on a GPL project.

PR #782, opened 2017-01-20
Add openssl support for optware and entware router

cxgreat2014:
Add "/opt/lib/libcrypto.so.1.0.0" support for optware and entware router


ysc3839:
I tested. There's two versions of libcrypto in my router. /lib/libcrypt-0.9.33.2.so(version 01000115F) and /lib/libcrypto.so.1.0.0(version 0100020AF).

cxgreat2014:
@ysc3839 if `$ python -c "import ctypes.util; print ctypes.util.find_library('crypto')"
libcrypto.so.1.0.0` work on your device, we needn't to do other change

ysc3839:
There's two versions of libcrypto. One is from the firmware and the other one is from Entware. The one from Entware is newer and faster.

Issue #756, opened 2017-01-10
Tor version parser broken?


adrelanos:
`zeronet --tor always` does not work for me.

> [22:17:05] TorManager Tor controller connect error: AttributeError: 'NoneType' object has no attribute 'group' in TorManager.py line 160

----

> `250-VERSION Tor="0.2.9.8 (git-a0df013ea241b026)"`

https://github.com/HelloZeroNet/ZeroNet/blob/master/src/Tor/TorManager.py#L160

HelloZeroNet:
well, i don't think if thats the problem:
```
>>> re.search('Tor="([0-9\.]+)', '250-VERSION Tor="0.2.9.8 (git-a0df013ea241b026)"').group(1)
'0.2.9.8'
```

Can you please add a `self.log.debug("Tor protocol: %s" % res_protocol)` before that line?

adrelanos:
Took the liberty to make that `self.log.info` so I don't have to enable `--debug` if that is okay.

```
[23:00:51] TorManager Tor protocol: 250-PROTOCOLINFO 1
250-AUTH METHODS=NULL

[23:00:51] TorManager Tor controller connect error: AttributeError: 'NoneType' object has no attribute 'group' in TorManager.py line 161
```

An issue with line handling?

-----

Could also be an issue with our [control-port-filter-python](https://github.com/Whonix/control-port-filter-python).

It receives:
```
PROTOCOLINFO
```
It replies:
```
250-PROTOCOLINFO 1
250-AUTH METHODS=NULL
250-VERSION Tor="0.2.9.8 (git-a0df013ea241b026)"
250 OK
```

MuxZeroNet:
Could you please find your `TorManager.py` file and add this line under [line 158](https://github.com/HelloZeroNet/ZeroNet/blob/master/src/Tor/TorManager.py#L158)?
```python
print "------", repr(res_protocol)
```

HelloZeroNet:
@MuxZeroNet he already did that, I think the problem is at https://github.com/HelloZeroNet/ZeroNet/blob/master/src/Tor/TorManager.py#L243, have to put that line to a `while not data:` loop, I will try to reproduce the problem

adrelanos:
Another option, if yo want the Tor version number, send `GETINFO version`, not ask for `protocolinfo`.

Yet another option would be porting to python-stem, which I created a separate ticket for just now:
https://github.com/HelloZeroNet/ZeroNet/issues/758

[...]

PR #784, opened 2017-01-20
TorManager.send(): read until repsonse terminator.

fred-a-kemp:
Tor command responses terminate with "250 OK\r\n" so we have to read
until that sequence is encountered.

The previous implementation is racy: after sending a command it would
accept whatever that is found on the socket as its response, no matter
if it is correctly terminated or not.

This commit fixes: https://github.com/HelloZeroNet/ZeroNet/issues/756


HelloZeroNet:
I can't test it, but thanks!

PR #791, opened 2017-01-24
Downvote when peer violates localhost restriction

MuxZeroNet:


Plasmmer:
Can you explain how this is intended to work?

Plasmmer:
@MuxZeroNet

PR #832, opened 2017-02-13
[WIP] Porting TorManager to Stem

MuxZeroNet:
Use Stem if we have it

### DIFF
--- a/src/Tor/TorManager.py
+++ b/src/Tor/TorManager.py
@@ -21,6 +21,10 @@ except:
 from util import helper
 from Debug import Debug
 
+try:
+    from stem.control import Controller
+except:
+    pass
 
 class TorManager:
     def __init__(self, fileserver_ip=None, fileserver_port=None):
@@ -40,10 +44,9 @@ class TorManager:
             self.status = "Disabled"
         else:
             self.enabled = True
-            try:
-                from stem.control import Controller
+            if sys.modules.get('stem'):
                 self.use_stem = True
-            except:
+            else:
                 self.use_stem = False
             self.status = "Waiting"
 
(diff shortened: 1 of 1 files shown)


MuxZeroNet:
```bash
python zeronet.py --tor always --tor_proxy 127.0.0.1:9150 --tor_controller 127.0.0.1:9151 --disable_udp
```
```
TorManager Authenticate using Stem... 127.0.0.1:9151
TorManager Tor stem connect error: SOCKS5Error: 0x01: General SOCKS server failure in TorManager.py line 213 > control.py line 998 > socket.py line 372 > socket.py line 243 > socket.py line 398 > socks.py line 681 > socks.py line 383 > socks.py line 458
TorManager self-bundled Tor fallback: Tor proxy port 127.0.0.1:9150 check error: No connection
```

HelloZeroNet:
Thanks, but I think it would be better as separate plugin, so you can avoid lots of ifs and does not makes the core source code more complicated. (it's easy check DonationMessage for example)

you can add to plugin's __init__.py like this:
```
try:
    import from stem.control import Controller 
    import StemPlugin
except:
   pass
```

and then easily override connect/addOnion/etc methods of TorController in StemPlugin.py

MuxZeroNet:
@HelloZeroNet Stem tends to route Tor Control Port information through the Tor Socks5 proxy. Any workaround for this?

HelloZeroNet:
Probably this is what needs to be modified: https://github.com/HelloZeroNet/ZeroNet/blob/master/src/util/SocksProxy.py#L6

MuxZeroNet:
How do I write the first lines of the plugin?
```python
@PluginManager.registerTo("TorManager")
class TorManagerPlugin(object):
```

PR #834, opened 2017-02-17
[WIP] StemPort Plugin

MuxZeroNet:
Closes the previous Pull Request #832.

### DIFF
--- a/src/Tor/TorManager.py
+++ b/src/Tor/TorManager.py
@@ -20,9 +20,11 @@ except:
     from gevent.lock import RLock
 from util import helper
 from Debug import Debug
+from Plugin import PluginManager
 
 
-class TorManager:
+@PluginManager.acceptPlugins
+class TorManager(object):
     def __init__(self, fileserver_ip=None, fileserver_port=None):
         self.privatekeys = {}  # Onion: Privatekey
         self.site_onions = {}  # Site address: Onion
@@ -58,7 +60,7 @@ class TorManager:
                 assert self.connect(), "No connection"
                 self.log.debug("Tor proxy port %s check ok" % config.tor_proxy)
             except Exception, err:
-                self.log.debug("Tor proxy port %s check error: %s" % (config.tor_proxy, err))
+                self.log.info("Starting self-bundled Tor, due to Tor proxy port %s check error: %s" % (config.tor_proxy, err))
                 self.enabled = False
                 # Change to self-bundled Tor ports
                 from lib.PySocks import socks
@@ -148,12 +150,15 @@ class TorManager:
         self.site_onions = {}
         self.privatekeys = {}
 
+        return self.connectController()
+
+    def connectController(self):
         if "socket_noproxy" in dir(socket):  # Socket proxy-patched, use non-proxy one
             conn = socket.socket_noproxy(socket.AF_INET, socket.SOCK_STREAM)
         else:
             conn = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
 
-        self.log.debug("Connecting to %s:%s" % (self.ip, self.port))
+        self.log.info("Connecting to Tor Controller %s:%s" % (self.ip, self.port))
         try:
             with self.lock:
                 conn.connect((self.ip, self.port))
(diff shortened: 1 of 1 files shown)


MuxZeroNet:
I am not able to make it work due to gevent problems.
```
LoopExit: ('This operation would block forever', <Hub at 0x7f001f0c0e10 epoll pending=0 ref=0 fileno=6>)
```

Gevent developer says "This is (almost always) not a problem in gevent." gevent/gevent#933

According to paramiko/paramiko#633, the `socket` module is monkey-patched to use gevent and therefore causes a `LoopExit` error.

HelloZeroNet:
If stem using socket lib that is not ready for gevent, then it will not work

MuxZeroNet:
I patched the socket module by calling `monkey.patch_all()`. Stem finally works but I got the following warning:
```
plugins/StemPort/StemPortPlugin.py:15:
RuntimeWarning: Patching more than once will result in the union of all True parameters being patched
```

MuxZeroNet:
http://www.gevent.org/gevent.monkey.html
```
patch_all(socket=True, dns=True, time=True, select=True, thread=True, os=True, 
    ssl=True, httplib=False, subprocess=True, sys=False, aggressive=True, Event=False, 
    builtins=True, signal=True)
```

> Changed in version 1.1: Issue a warning if this function is called **multiple times** with **different arguments**. The second and subsequent calls will only add more patches, they **can never remove** existing patches by setting an argument to False.

I think the problem is, I called `patch_all` with the wrong arguments. @HelloZeroNet What do you think the parameters should be used here in order to be compatible with the other `patch_all` calls in ZeroNet?

MuxZeroNet:
Now it seems to be working. Stem created hidden services. I got exit node IPs. Site updating and pushing seemed to work well. I got timeout when trying to push sites to onion peers, buy it might be a network latency issue.

MuxZeroNet:
`createSocket` got overridden by mistake.

MuxZeroNet:
> **All checks have failed**
> 1 failing check

Broken WebSocket connection results in a failure of [Travis CI Tests](https://travis-ci.org/HelloZeroNet/ZeroNet/builds/204786638). Is this related to StemPort plugin?

```
plugins/CryptMessage/Test/TestCrypt.py::TestCrypt::testPublickey PASSED
plugins/CryptMessage/Test/TestCrypt.py::TestCrypt::testEcies PASSED
plugins/CryptMessage/Test/TestCrypt.py::TestCrypt::testEciesUtf8 PASSED
plugins/CryptMessage/Test/TestCrypt.py::TestCrypt::testEciesAes **FAILED**
plugins/CryptMessage/Test/TestCrypt.py::TestCrypt::testAes PASSED
plugins/CryptMessage/Test/TestCrypt.py::TestCrypt::testAesUtf8 PASSED
```

```
        # Decrypt using AES
        aes_iv, aes_encrypted = CryptMessage.split(ecies_encrypted.decode("base64"))
        ui_websocket.actionAesDecrypt(0, aes_iv.encode("base64"), aes_encrypted.encode("base64"), aes_key)

>       assert ui_websocket.ws.result == "hello"
E       assert None == 'hello'
E        +  where None = <src.Test.conftest.WsMock instance at 0x7fb0d19f09e0>.result
E        +    where <src.Test.conftest.WsMock instance at 0x7fb0d19f09e0> = <Plugin.PluginManager.UiWebsocket object at 0x7fb0d19d2350>.ws

plugins/CryptMessage/Test/TestCrypt.py:71: AssertionError
```

HelloZeroNet:
It was temporary problem, re-running fixed it.

MuxZeroNet:
I have been using it for a while. No DNS leaks, is able to receive content from and push content to Tor Network peers, and it creates onion services quickly. Could anyone test this in Whonix? @adrelanos

In making this plugin, I assume you are always using the latest version of Tor and Tor Browser.

adrelanos:
How do I test this? Just run from your branch MuxZeroNet/patch-13 with `--tor always`? Would this enable the plugin?

MuxZeroNet:
First download `MuxZeroNet/patch-13` branch, and go to `plugins/` folder to enable the `StemPort` plugin. (Rename `disabled-StemPort` to `StemPort`)

You can use these parameters. (I use Tor Browser)
```
python zeronet.py --tor always --tor_proxy 127.0.0.1:9150 --tor_controller 127.0.0.1:9151
```

Make sure you have `stem` installed in your system. In my environment, I just downloaded the Git HEAD version of stem, and put the `stem` folder inside `ZeroNet/src/` directory.

adrelanos:
Works for me in Whonix.

When using `--tor always`, I guess `--tor_proxy 127.0.0.1:9150 --tor_controller 127.0.0.1:9151` will be the default and do not need to be manually set?

MuxZeroNet:
The default values are 127.0.0.1:9050 (for tor_proxy) and 127.0.0.1:9051 (for tor_controller).

If it fails to connect to 127.0.0.1:9051 it starts its own Tor and tries 127.0.0.1:49051

PR #839, opened 2017-02-21
Make the cmd launcher "silent" and not stick around(fix for win10)

beigexperience:
The codepage changer was outputing a short msg, also waiting for the exe, this way it should start it and go away.


HelloZeroNet:
Thanks!

PR #842, opened 2017-02-25
Rename SiteStorage.list to SiteStorage.walk

mishfit:
also, add SiteStorage.list as non-recursive
directory listing (`os.listdir`)

I've got a use case for listing the contents of a directory (non-recursively) and the nomenclature works better with 'list' (which is what it's actually doing) while the pre-existing 'list' function is actually performing a 'walk'.

I've renamed 'list' to 'walk' (to make more sense) and added 'list' for a single directory.

### DIFF
--- a/plugins/Mute/MutePlugin.py
+++ b/plugins/Mute/MutePlugin.py
@@ -36,7 +36,7 @@ class UiWebsocketPlugin(object):
             if not site:
                 continue
             dir_inner_path = helper.getDirname(row["inner_path"])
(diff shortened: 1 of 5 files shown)


HelloZeroNet:
I'm ok with this, but I think a `def actionFileList(self, to, inner_path, walk=True):` would be better, than a separate function

mishfit:
I would ordinarily agree, however the original `list` excludes directories in its resultset whereas the new list includes directories (something I actually want)

PR #850, opened 2017-03-05
Added Dutch translation file

sexybiggetje:
Adds Dutch translation file. I was unable to get it in the language menu, so tested it in another language file. Do I need to do more?


sexybiggetje:
I've added a proper translation to ZeroHello as well in https://github.com/HelloZeroNet/ZeroHello/pull/44

PR #869, opened 2017-03-23
Root content.json "signers"  field as a list

MRoci:
Changed this assignment to have the "signers" field in root's content.json as a list and not as a dictionary to mantain uniformity with "includes" "signers" field and for not having a dict with empty values.

### DIFF
--- a/src/Content/ContentManager.py
+++ b/src/Content/ContentManager.py
@@ -631,7 +631,7 @@ class ContentManager(object):
         valid_signers = []
         if inner_path == "content.json":  # Root content.json
             if "content.json" in self.contents and "signers" in self.contents["content.json"]:
-                valid_signers += self.contents["content.json"]["signers"].keys()
+                valid_signers += self.contents["content.json"]["signers"][:]
         else:
             rules = self.getRules(inner_path, content)
             if rules and "signers" in rules:


MRoci:
@HelloZeroNet any update on this issue? Is it correct what i've done or i've misunderstood the assignement?

HelloZeroNet:
You are right it should be list, but acutally the multisig is disable atm. (https://github.com/MRoci/ZeroNet/blob/4a981e88ada8d26ef673639feb9d4db442818b9e/src/Content/ContentManager.py#L648) It would be a great addition, but needs more testing

MRoci:
Thank you, i've seen this issue because if you try to add  multiple addresses to the signers filed it lets you sign the root content.json with different addresses.

imachug:
@shortcutme Any news on this?

HelloZeroNet:
I think we can change it to list without supporting dict, since the multisig is not enabled yet, so there won't be any backward compatibility issue.

imachug:
@HelloZeroNet So could you commit this?

HelloZeroNet:
Yeah change `valid_signers += self.contents["content.json"]["signers"].keys() 
` it simply to `valid_signers += self.contents["content.json"]["signers"][:] ` and I will accept it.

MRoci:
@HelloZeroNet  done

PR #870, opened 2017-03-24
add FreeBSD installation instructions to README-zh-cn.md

l5h5t7:


HelloZeroNet:
The plugins/Mute/languages/zh.json file has some merging junk left, pelase remove that file (and plugins/Mute/languages/zh-tw.json ) from commit

PR #896, opened 2017-04-10
Make sure data types are correct / expected

MuxZeroNet:
This pull request is a partial fix of various type assumption glitches in ZeroNet.

These corrections are made as response to a very simple fuzz testing I conducted against FileServer. (I will write about the fuzz testing soon.) That 15441 port is open to everyone in the network, so it's better to be conservative.

I also make it distinguish between detected errors and unexpected errors. By making a distinction between detected errors and unexpected errors, logging noise is reduced, and more important error logs can be seen easier.

### DIFF
--- a/src/Connection/Connection.py
+++ b/src/Connection/Connection.py
@@ -10,6 +10,16 @@ from util import StreamingMsgpack
 from Crypt import CryptConnection
 
 
+def ensure_type(instance, desired_type, allow_none=False):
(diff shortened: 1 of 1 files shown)


MuxZeroNet:
More detail here: https://github.com/MuxZeroNet/Fuzzing/tree/master/Fuzzing%20Log/Round%201

PR #912, opened 2017-04-21
Allow empty fileQuery filters

mishfit:
Currently `actionFileQuery` requires a string value for the query (filter) parameter. To run a query without "filtering" an empty string is required.

This PR makes `query` optional.


mishfit:
@shortcutme can I get eyes on this?

HelloZeroNet:
I will arrive back from Canada later this week, so I can check it then

mishfit:
Awesome, hope it's been fun.

Issue #914, opened 2017-04-23
Pass hash to iframe


imachug:
When you open http://127.0.0.1:43110/... and then change URL to http://127.0.0.1:43110/...#abc, the iframe changes its `location.hash` to `#abc`. That's OK.

When you directly open http://127.0.0.1:43110/...#abc, the iframe's `location.hash` is still empty! Changing to http://127.0.0.1:43110/...#def works again.

So, we could check `location.hash` in wrapper and pass it to inner iframe.

imachug:
I solved the problem. For anybody who will read this later: **call `.cmd("wrapperInnerLoaded", [])` when you are ready to catch hashchange**

PR #915, opened 2017-04-23
Pass #hash to inner iframe on load

imachug:
See #914

### DIFF
--- a/src/Ui/template/wrapper.html
+++ b/src/Ui/template/wrapper.html
@@ -60,7 +60,7 @@ if (window.opener && window.stop) window.stop()
 <!-- Site info -->
 <script>
 document.getElementById("inner-iframe").src = "about:blank"
-document.getElementById("inner-iframe").src = "{file_url}{query_string}"
+document.getElementById("inner-iframe").src = "{file_url}{query_string}" + window.location.hash
 address = "{address}"
 wrapper_nonce = "{wrapper_nonce}"
 wrapper_key = "{wrapper_key}"


mishfit:
While I like the idea of pushing the hash to the client I think it shouldn't be in the template for the wrapper, but instead somewhere in the [javascript portion of the logic for the wrapper](https://github.com/HelloZeroNet/ZeroNet/blob/master/src/Ui/media/Wrapper.coffee#L372-L380):
```javascript
# Iframe loaded
	onLoad: (e) =>
		@inner_loaded = true
		if not @inner_ready then @sendInner {"cmd": "wrapperReady"} # Inner frame loaded before wrapper
		#if not @site_error then @loading.hideScreen() # Hide loading screen
                //# ***NOTE***
                // somewhere in here the hash should be pushed to the inner_frame
                //***
		if @ws.ws.readyState == 1 and not @site_info # Ws opened
			@reloadSiteInfo()
		else if @site_info and @site_info.content?.title?
			window.document.title = @site_info.content.title+" - ZeroNet"
			@log "Setting title to", window.document.title
```
This would keep the logic in the same file as [this](https://github.com/HelloZeroNet/ZeroNet/blob/master/src/Ui/media/Wrapper.coffee#L29-L33) (which handles hash changes):
```javascript
window.onhashchange = (e) => # On hash change
			@log "Hashchange", window.location.hash
			if window.location.hash
				src = $("#inner-iframe").attr("src").replace(/#.*/, "")+window.location.hash
				$("#inner-iframe").attr("src", src)
```

HelloZeroNet:
This is intentionally: like @mishfit pointed out you have to call the wrapperReady api request, because if you have dynamic page you need to load resources to render the page before you can scroll to the anchor

imachug:
BTW, sometimes the inner iframe reloades on hashchange. What about changing:
```javascript
src = $("#inner-iframe").attr("src").replace(/#.*/, "")+window.location.hash
$("#inner-iframe").attr("src", src)
```
to:
```javascript
$("#inner-iframe").attr("src", window.location.hash)
```

Is there any reason why ```src``` is changed instead of hash?

imachug:
Well, I think, that's my fault. I didn't know that `wrapperInnerLoaded` is a valid command itself. For me, [ZeroFrame API Reference on readthedocs.io](https://zeronet.readthedocs.io/en/latest/site_development/zeroframe_api_reference/) was the best reference, but `wrapperInnerLoaded` section appeared *after* I found this issue.

PR #938, opened 2017-05-19
Weak password warning

MuxZeroNet:
Weak password warning

### DIFF
--- a/src/Ui/UiWebsocket.py
+++ b/src/Ui/UiWebsocket.py
@@ -48,48 +48,8 @@ class UiWebsocket(object):
             file_server = sys.modules["main"].file_server
             if file_server.port_opened is None or file_server.tor_manager.start_onions is None:
                 self.site.page_requested = False  # Not ready yet, check next time
-            elif file_server.port_opened is True:
-                self.site.notifications.append([
-                    "done",
-                    _["Congratulation, your port <b>{0}</b> is opened.<br>You are full member of ZeroNet network!"].format(config.fileserver_port),
-                    10000
-                ])
-            elif config.tor == "always" and file_server.tor_manager.start_onions:
-                self.site.notifications.append([
-                    "done",
-                    _(u"""
-                    {_[Tor mode active, every connection using Onion route.]}<br>
-                    {_[Successfully started Tor onion hidden services.]}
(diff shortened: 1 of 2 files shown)


HelloZeroNet:
Thanks, some things:
 - Removal of `self.cmd("error", "Internal error: %s" % Debug.formatException(err, "html"))` would make the error finding more difficult, so maybe we could show it depending on `if "Multiuser" in PluginManager.plugin_manager.plugin_names:`
 - Instead of print it would be better to use `logging.warning`
 - Name `sendHomepageNotes` to `sendHomepageNotifications` to be more clear?

src/Ui/UiWebsocket.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, ws, site, server, user, request):
         self.async_commands = ("fileGet", "fileList", "dirList")
 
     # Start listener loop
+    # (Uncaught exceptions are silently suppressed. Please print them out.)
```
HelloZeroNet: Can you please explain what you mean?

src/Ui/UiWebsocket.py
```diff
@@ -48,52 +49,21 @@ def start(self):
...
+                    self.addHomepageNotifications()
+                except Exception, err:
+                    self.log.error("Uncaught Exception: " + Debug.formatException(err))
 
         for notification in self.site.notifications:  # Send pending notification messages
+            # send via WebSocket
             self.cmd("notification", notification)
+            # just in case, log them to terminal
+            if notification[0] == "error":
+                self.log.error("\n\033[93mWARNING:\033[0m %s\n" % self.dedent(notification[1]))
```
HelloZeroNet: Adding some formatting/coloring to the output could be a good idea. But I think it has to done on the logging module level, because this way it messes up the output for non-compatible terminals and the log file.

MuxZeroNet:
I mean if an uncaught exception is thrown in the frame of `start` method, the routine gets terminated, but no traceback is printed.

HelloZeroNet:
Thanks!
Added logging for ws.receive() errors: https://github.com/HelloZeroNet/ZeroNet/commit/79005780772350db31042747749b129132f941dd

Issue #936, opened 2017-05-17
Bug: coffeescript wrong compilation paths with custom data_dir


grez911:
I have the following conditions:
OS: Arch Linux
Zeronet is installed in `/opt/zeronet/`
Directory with zites is `/var/lib/zeronet/`
My newly created site is in
`/var/lib/zeronet/13KHcN8FvmY9UgZ2DWAc2323f28qE5LZic/`

Config /etc/zeronet.conf:
```
[global]
data_dir = /var/lib/zeronet
log_dir = /var/log/zeronet
```

When I run zeronet with the following command:
`python2 /opt/zeronet/zeronet.py --config_file /etc/zeronet.conf --debug`

and go to my zite in browser, these alerts are popping:
```
/var/lib/zeronet/13KHcN8FvmY9UgZ2DWAc2323f28qE5LZic/js/utils/Class.coffee compile error: File not found: /opt/zeronet/var/lib/zeronet/13KHcN8FvmY9UgZ2DWAc2323f28qE5LZic/js/utils/Class.coffee
/var/lib/zeronet/13KHcN8FvmY9UgZ2DWAc2323f28qE5LZic/js/utils/ZeroFrame.coffee compile error: File not found: /opt/zeronet/var/lib/zeronet/13KHcN8FvmY9UgZ2DWAc2323f28qE5LZic/js/utils/ZeroFrame.coffee
```

in logs I have:

```
[20:48:37] - Running: "/usr/bin/coffee" --no-header -p "var/lib/zeronet/13KHcN8FvmY9UgZ2DWAc2323f28qE5LZic/js/utils/ZeroFrame.coffee" (Done in 0.09s)
[20:48:37] - /var/lib/zeronet/13KHcN8FvmY9UgZ2DWAc2323f28qE5LZic/js/utils/ZeroFrame.coffee Compile error: File not found: /opt/zeronet/var/lib/zeronet/13KHcN8FvmY9UgZ2DWAc2323f28qE5LZic/js/utils/ZeroFrame.coffee
```

grez911:
Changing this string: https://github.com/HelloZeroNet/ZeroNet/blob/master/src/Debug/DebugMedia.py#L88
from
`file_path_escaped = helper.shellquote(os.path.join(*file_path.split("/")))`
to
`file_path_escaped = helper.shellquote("/" + os.path.join(*file_path.split("/")))`
helped me. I hope someone will fix this in a more proper way.

[...]

PR #942, opened 2017-05-22
fix issue #936

grez911:
os.path.join(*file_path.split("/")) construction drops leading slash from string thus making absolute path relative. It leads into problems in case if data_dir differs from directory where is zeronet installed.


HelloZeroNet:
Thanks!

PR #954, opened 2017-05-26
added missing import for sys,ssl module

cryptofreedom:
fixed the missing import error

### DIFF
--- a/src/util/SslPatch.py
+++ b/src/util/SslPatch.py
@@ -3,7 +3,9 @@
 
 import logging
 import os
+import ssl
 
+import sys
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
Thanks for reporting, you are right the sys module was missing. 
The ssl module wasn't used, changed the variable name to avoid confusion and added other improvements:
https://github.com/HelloZeroNet/ZeroNet/commit/e291555e604f9db1a273a39db40642aca5a3e45a#diff-8483a7f5b31679cec82cabfd6c3a7c2d

Issue #941, opened 2017-05-22
Cannot encode Base64 value: 0\n


kavamaken:
![zeronet1](https://cloud.githubusercontent.com/assets/5235877/26319013/85731b54-3f47-11e7-9ef3-26aae43e7185.png)
![zeronet2](https://cloud.githubusercontent.com/assets/5235877/26319011/856df142-3f47-11e7-8acd-43889d24733b.png)
![zeronet3](https://cloud.githubusercontent.com/assets/5235877/26319012/856f1e6e-3f47-11e7-9106-2e8acc53188d.png)
Any page wont startup.
version is 0.5.5-2081 at windows.

HelloZeroNet:
Are you running ZeroNet in debug mode?
Have you modified any default running parameters? (eg. using zeronet.conf)

kavamaken:
Yes, I use zeronet.conf to turn off tor.
The message appears regardless of debug mode is on or off.

imachug:
@shortcutme, Could you test #974 which fixes this issue?

HelloZeroNet:
Fixed in latest rev: https://github.com/HelloZeroNet/ZeroNet/commit/1f83b6691b7864a35c1466ee23a14628f28ec974

PR #974, opened 2017-06-12
Replace tools/coffeescript with pip package

imachug:
Replaces BAT/WSH/JS *solution* with pip package.

On my machine WSH script just gave `Compiler error: Unable to Base64 encode...`. That's probably because:
```javascript
typeof "ABCDE"[0] == "undefined"
```

...Don't know why it worked before. Maybe because ZeroNet saw my own CoffeeScript compiler?


imachug:
Fixes #941

HelloZeroNet:
Thanks! 
I try to avoid external dependencies if possible and was able to fix this issue: https://github.com/HelloZeroNet/ZeroNet/commit/1f83b6691b7864a35c1466ee23a14628f28ec974

imachug:
@HelloZeroNet It works. Thanks!

PR #975, opened 2017-06-12
Partly implement multisig

purplesyringa:
Signs with keys are now saved. `signs_required` is checked.

### DIFF
--- a/src/Content/ContentManager.py
+++ b/src/Content/ContentManager.py
@@ -731,7 +731,6 @@ class ContentManager(object):
 
         sign_content = json.dumps(new_content, sort_keys=True)
         sign = CryptBitcoin.sign(sign_content, privatekey)
-        # new_content["signs"] = content.get("signs", {}) # TODO: Multisig
         if sign:  # If signing is successful (not an old address)
             new_content["signs"] = old_signs_content or {}
             new_content["signs"][privatekey_address] = sign
--- a/src/Content/ContentManager.py
+++ b/src/Content/ContentManager.py
@@ -247,7 +247,6 @@ class ContentManager(object):
                         deleted += include_deleted  # Add changed files
 
             # Save some memory
-            new_content["signs"] = None
             if "cert_sign" in new_content:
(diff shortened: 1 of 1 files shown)


imachug:
By the way, @HelloZeroNet.

Say, we have the following `content.json`:
```json
{
    ...
    "signers": [
        "A",
        "B"
    ],
    "signs": {
        "A": "Sign-for-version-1",
        "B": "Sign-for-version-1"
    },
    "signers_required": 2
    ...
}
```

Then, `A` adds new post:
```json
{
    ...
    "signers": [
        "A",
        "B"
    ],
    "signs": {
        "A": "Sign-for-version-2",
        "B": "Sign-for-version-1"
    },
    "signers_required": 2
    ...
}
```
...and tries to publish it. But he can't! All peers reject this `content.json` because only 1 sign (`A`) is valid. How can `B` get this `content.json` and sign it with his `B` key?

HelloZeroNet:
Unfortunetly I think we can't avoid it... A has to send the content.json directly to B before it can publish to other peers.

rllola:
There is no way to doing it with Zeronet ? You necessarily need to send it manually ? This might be annoying for users to do it everytime.

krixano:
One way that would help this is if the content.json was in version control. Then, B could pull the latest commit made by A and sign and publish. Still a bit annoying I suppose.

Thunder33345:
maybe for it to hold two versions of the site?
a signer pending version? where only content.json is downloaded and shared, when B gets it he can decide to sign or not while staying anonymous

DaniellMesquita:
@krixano

> One way that would help this is if the content.json was in version control. Then, B could pull the latest commit made by A and sign and publish. Still a bit annoying I suppose.

Not so annoying if it implements archival features, likes IPFS haves.

@imachug
Could PeerMessage plugin help it?
What do you think about a archive plugin?

purplesyringa:
I guess I'll close this PR because it doesn't address many problems and doesn't add any features.

PR #978, opened 2017-06-12
Add Spanish Translation

Pupiloho:
First revision, orthography checked.


imachug:
@Pupiloho, Please don't make four branches (`patch-1`, `patch-2`, `patch-3`, `patch-4`), add commits to one branch (`patch-sp-translation` or whatever) instead.

PR #981, opened 2017-06-13
Allow fileRules for root content.json

imachug:
Currently `fileRules('content.json')` throws `Pop from empty list`. This patch fixes the exception and gives valid result.


HelloZeroNet:
Thanks!

PR #984, opened 2017-06-15
Don't ask user for privatekey if file can be signed with auth_privatekey

imachug:
In ZeroNet core, both `privatekey` and `auth_privatekey` are used to check if user can sign `content.json`, but there is no such check in `Sidebar` plugin.

Here we also check if `auth_address` can sign `content.json`.


HelloZeroNet:
Thanks!

PR #1017, opened 2017-07-14
Add flake8 to testing

cclauss:
Are these failures valid?

Also did consistent indenting

### DIFF
--- a/src/util/SslPatch.py
+++ b/src/util/SslPatch.py
@@ -6,6 +6,7 @@ import os
 import sys
 import ctypes
 import ctypes.util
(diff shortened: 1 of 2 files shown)


src/util/SslPatch.py
```diff
@@ -6,6 +6,7 @@
 import sys
 import ctypes
 import ctypes.util
+import ssl
```
cclauss: @shortcutme is this the right way to resolve the undefined name errors at https://travis-ci.org/cclauss/itinerant-tester/jobs/259296035#L314-L322

PR #1018, opened 2017-07-14
Import ssl

cclauss:
See #1017


cclauss:
This works. Moved into #1017

PR #1028, opened 2017-07-17
Revamped Dockerfile 🐋

flibustier:
List of changes:
- Using Alpine instead of Ubuntu (the image's size is now *77 MB* instead of *421 MB*)
- Remove `MAINTAINER` because it's deprecated
- Using `COPY` instead of `ADD` because it's deprecated
- `WORKDIR` instead of `cd directory` following best practices


HelloZeroNet:
Thanks!

PR #1062, opened 2017-07-31
data_unpack --> data_unpacked

cclauss:
@shortcutme Is this the right way to fix these two undefined name errors?


HelloZeroNet:
Yes, thanks!

Issue #1016, opened 2017-07-13
ZeroNet is possibly under attack


Plasmmer:
I seriously believe that ZeroNet is suffering a DDOS attack.
I saw 1732 seeds in ZeroName.
In less than 4 minutes ZeroName is with 1608 seeds.
@HelloZeroNet, what do you think about preventing custom clients that can download a site but doesn't seed?
Also these great volume of "users" are seeding in majority ZeroName against ZeroHello. It is normal?
And when I tried to publish a comment about it in ZeroMe, it was stuck on publishing to 4 peers.

HelloZeroNet:
There is a bit more peer than usual, but I don't see any suspicious thing.
The peer number is just an estimation. If you running your client for longer time you will see more due peer exchange. Also, more  frequently updated sites = more connections = more peer exchange = more peers.

It's an opensource protocol/app, so you can't do much about making custom modifications, but I don't see any problem with that. If someone does not have spare bandwidth to seed sites, then he/she can use a proxy or simply put it on paused state.

Publishing to 4 peers should be fine.

Plasmmer:
http://127.0.0.1:43110/Talk.ZeroNetwork.bit/?Topic:69_1EZTWytasXcKdiUFnN4KHqAfvg1nwkTkKv
@Nofish: this NEEDS to be fixed with URGENCY. Thanks @kai0: now I know what is hapenning here.
Here I'm persecuted by the owners here, and they spy in my house anything I do in the Internet. When I created a site, in less than 1 minute it gained 5 peers without any promoting!
I properly signed it and tried to publish to a lot of proxies like I ever did in the last year, but now the proxies didn't received my updates. I know these 5 fixed peers are the mans that are persecuting me, and they made my site inavailable to the following peers. In the last year no, but now it was discovered by they. It needs to be fixed, because is clearly a security vulnerability.
Here are the prints, as a timeline:
http://prntscr.com/fx4482
http://prntscr.com/fx44d0
Without promoting, my site has 2 peers connected. One minute later, I saw it unavailable in the proxies.
http://prntscr.com/fx46pt
Stucked in the publishing (it can be normal, I know).
http://prntscr.com/fx470a
Stuck loading on proxies. I tried to publish more to reach proxies, everytime in this timeline.
http://prntscr.com/fx482w
Loaded only a few things.
http://prntscr.com/fx49kt
Loaded more things, but JS don't loaded.
http://prntscr.com/fx4afv
[...]

PR #1063, opened 2017-07-31
import ssl

cclauss:
@HelloZeroNet is this the right way to resolve the undefined name errors at https://travis-ci.org/cclauss/itinerant-tester/jobs/259296035#L314-L322 ?

### DIFF
--- a/src/util/SslPatch.py
+++ b/src/util/SslPatch.py
@@ -6,6 +6,7 @@ import os
 import sys
 import ctypes
 import ctypes.util
+import ssl
 
 from Config import config
 


Plasmmer:
@cclauss, what do you think about #1016?

HelloZeroNet:
Thanks, I fixed it using the unpatched ssl lib: https://github.com/HelloZeroNet/ZeroNet/commit/b8d68e2589468fb6fe97d599b1a3a39e173a4214

__import__("ssl") and import ssl is not the same because of gevent monkey patching.

Issue #80, opened 2015-04-12
P2P port open checking


HelloZeroNet:
Does not relies on canyouseeme.org, also possible to use to check i2p/tor hidden service.

MuxZeroNet:
Hi there. Any progress on this? This can be used in public gateways.

HelloZeroNet:
No progress on it, but i'm not sure what you mean.

In my theory it would work like this:
 - Collect some IPs via Trackers/Peer database/etc.
 - Ask them to connect back to the given port and return the result

Based on this information the client would be possible to find out if their public ip/hidden service is works well or not.

grez911:
@HelloZeroNet I made a pull request to close this issue, please, review it.

grez911:
@HelloZeroNet All checks have passed.

grez911:
Not closing because there is need to add tor hidden service check also.

HelloZeroNet:
I think Tor hidden service checking is not necessary. If the hidden service has registered successfully, then it should work regardless your network settings. 
It also has 1-5 minute cooldown time, so you can't do it on startup.

grez911:
@HelloZeroNet Then closing, I think.

PR #1076, opened 2017-08-15
Added P2P portcheck

grez911:
Issue: https://github.com/HelloZeroNet/ZeroNet/issues/80

### DIFF
--- a/src/File/FileRequest.py
+++ b/src/File/FileRequest.py
@@ -3,6 +3,7 @@ import os
 import time
 import json
 import itertools
+import socket
 
 # Third party modules
 import gevent
@@ -13,6 +14,7 @@ from util import RateLimit
 from util import StreamingMsgpack
 from util import helper
 from Plugin import PluginManager
+from contextlib import closing
 
 FILE_BUFF = 1024 * 512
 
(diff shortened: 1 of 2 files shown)


HelloZeroNet:
Thanks, I think it would be better to use object as return of the actionCheckport request, use config.homepage instead of fixed site address and i'm not sure if socket.connect_ex does not leaks real IP address in tor mode.

grez911:
> i'm not sure if socket.connect_ex does not leaks real IP address in tor mode.

@HelloZeroNet You do a [socket patch](https://github.com/HelloZeroNet/ZeroNet/blob/de550d7cae940630855a9466e88695f2b3a6a950/src/main.py#L137) in main.py if --tor always enabled, I think this will redirect all socket connections to Tor, but I will try to catch packets with Wireshark to be completely sure.

grez911:
@HelloZeroNet, I did it! There are two pcap-files in [this archive](https://github.com/HelloZeroNet/ZeroNet/files/1232142/cap.zip) each of which contain packets recorded on the receiving end of the suspicious `socket.connect_ex` command.
One file contains capture for Tor only mode and another for default zeronet start without any parameters.
Just a little explanation, what I did. I opened port 4444 on my remote VPS with `nc -k -l -p 4444` command and in the other console on the VPS started dumping packets to file with `tcpdump dst port 4444 -w out.cap` command.
Small changes have been made in this pull request to force zeronet connect to 81.4.104.218:4444 instead of a random peer of HelloZeroNet page. Here the scheme:
```
+---------+ 93.84.53.139    Clearnet    81.4.104.218:4444 +-----+
| zeronet |---------------------------------------------->| VPS |
+---------+                                         ^     +-----+
      |                                             |
      | Tor    +-----------+ 176.126.252.12         |
      +------->| Exit node |------------------------+
               +-----------+
```
So you can open records with Wireshark and make sure that there is absolutely no records about my real IP (93.84.53.139) when Tor always enabled.

P. S. Object return and config.homepage I will add to the pull request later, please, wait.

grez911:
@HelloZeroNet, all done. Is there anything else?

HelloZeroNet:
Thanks!

Issue #430, opened 2016-04-22
didn't gives error when users.json invalid


zeronetscript:
if manually edit users.json to invalid json, zeronet start with a blank page, without any warning

border0464111:
hummmm, Error should be visible when content is signed. both from the webUI and CLI.

grez911:
@HelloZeroNet Please, review the pull request.

PR #1091, opened 2017-08-22
Added exception handling for broken user.json

grez911:
https://github.com/HelloZeroNet/ZeroNet/issues/430


HelloZeroNet:
Thanks!

Issue #688, opened 2016-12-02
Multi-language interface is unavailable on public proxys


l5h5t7:
The configuration of language is stored in `zeronet.conf` so user can't change it. All users can only use the language which public proxy owner configured.
Allow users change language simply is not feasible because all users will be affected. So it needs a separate setting form uesrs store in cookie/localstorage and auto-detect via browser's language setting.

grez911:
@HelloZeroNet Can you please help with advice how to implement it correctly?
I have a few questions:
1. It is better to make changes in project core than in Multiuser plugin because I don't see something bad if language settings will be stored in browser cache both for single-user and multi-user usage. And this will improve consistency of the project. Do you agree with me?
2. Changes must be done in `Site/SiteStorage.py` and `Ui/UiWebsocket.py` files: `Translate` class must be initialized with `lang` variable in each invocation, there are must be no default calls without parameters. Right?
3. How and where better to get cookies from the browser?

HelloZeroNet:
At the time of the plugin release I was about to add this feature, but then decided not to add it as it would reduce the performance and add code complexity. (having per-proxy language is also reduces centralization)
Probably cookie would be better as simple http requests and websocket messages also needs to be translated.

grez911:
@HelloZeroNet Please review this pull request. [ZeroHello](https://github.com/HelloZeroNet/ZeroHello) code must be changed too to store language setting in cookies. Maybe I will do it later.

PR #1097, opened 2017-08-25
Added an extraction of the language setting from cookies

grez911:
https://github.com/HelloZeroNet/ZeroNet/issues/688

### DIFF
--- a/src/Ui/UiServer.py
+++ b/src/Ui/UiServer.py
@@ -13,7 +13,7 @@ from UiRequest import UiRequest
 from Site import SiteManager
 from Config import config
 from Debug import Debug
-
+import Translate
 
 # Skip websocket handler if not necessary
 class UiWSGIHandler(WSGIHandler):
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
The problem is the Translate object is shared between users, so if you change it to someone, then it will change to everyone who currently connected via websocket.

grez911:
@HelloZeroNet So the translation must be comletely moved to the frontend -- rewritten in javascript. Do you think it will be the clearest solution too?

HelloZeroNet:
i'm not sure how would parametized translations (eg.: `Change it to {auth_type}/{auth_user_name}@{domain}` or `Content publish queued for {0:.0f} seconds.`) would work that way

grez911:
@HelloZeroNet What if the websocket will send messages in English and then we translate it via simple substitution in JS that will not touch parametrized output at all. Something like this:
```
>> str = "Change it to {auth_type}/{auth_user_name}@{domain}"
>> str = str.replace("Change it to", "Változtatás")
>> print(str)
"Változtatás {auth_type}/{auth_user_name}@{domain}"

>> str = "Content publish queued for {0:.0f} seconds."
>> str = str.replace("Content publish queued for", "Tartalom publikálása elhalasztva")
>> str = str.replace("seconds", "másodperccel")
>> print(str)
"Tartalom publikálása elhalasztva {0:.0f} másodperccel."
```

What do you think?

PR #1120, opened 2017-09-21
Add commands for scripts

imachug:
Looks like `zeronet.py` is designed for *a)* those who don't use browser, *b)* scripts. This patch adds some more script commands (`getDataDir`, `getLogDir`, `getRootDir`) because paths can vary on different systems.


HelloZeroNet:
I think a getConfig action would be better that returns all the information at once. (json formatted)

imachug:
@HelloZeroNet Then take a look at #1123, please.

PR #1122, opened 2017-09-22
Allow siteClone for non-admin sites

imachug:
Adds notification "Clone site **1HeLLo...**? `Clone`" for non-admin sites. For sites with ADMIN permission, leave everything as is.

Rejects `{"error": "Not a site: something"}` if `address` is not valid. If source site wasn't downloaded yet, stop cloning, but don't tell the client about it to avoid exposing site existense.


HelloZeroNet:
Thanks, looking great!

PR #1123, opened 2017-09-22
[WIP] Move server info formatting to Config.py; use it via getConfig console command

imachug:
Also return `data_dir`, `log_dir` and `src_dir` (for scripts).

### DIFF
--- a/src/Config.py
+++ b/src/Config.py
@@ -423,7 +423,11 @@ class Config(object):
             "rev": self.rev,
             "language": self.language,
             "debug": self.debug,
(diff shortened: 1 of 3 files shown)


imachug:
This is not yet finished, don't merge please.

HelloZeroNet:
I think the server info exposed to the websocket API should be different, because adding the client location on the fs could be privacy problem.

imachug:
@HelloZeroNet, I've just understood it and marked this PR as [WIP].

PR #1165, opened 2017-10-26
DbQuery and Newsfeed bug

imachug:
@shortcutme, I understand that SQL is difficult, but using regular expressions to parse SQL... Here are some more hacks:

1. Add (brackets) around NewsFeed's `body LIKE ? OR title LIKE ?`.
2. WHERE without AND is still WHERE


imachug:
@HelloZeroNet ?

HelloZeroNet:
Thanks!

PR #1172, opened 2017-10-30
Update README.md

ajmeese7:
Added proper hyphenation and changed OSX to OS X


HelloZeroNet:
Well actually OS X is macOS now :)

ajmeese7:
@HelloZeroNet True true. Will change :thumbsup:

PR #1201, opened 2017-12-07
python 2.7 required

ubaidsworld:
It works with the python version 2.7 currently.


imachug:
Does it make sense? It is already stated that Python 2 is required: `* Start with python2 zeronet.py`. And versions like 2.3 are outdated much, so nobody would really use them.

imachug:
@HelloZeroNet Can we close this as ZeroNet is going to switch to py3?

PR #1230, opened 2017-12-30
Handle :param being an array

imachug:
I am not sure if this is so important, but using `:params` in `NewsFeed` queries proved it is useful.

So now the following code will work:

```javascript
page.cmd("dbQuery", [
    "SELECT * FROM posts WHERE posts.reply_to IN :replies AND posts.reply_to_json IN :replies_json"
], {
    replies: ["1", "2"],
    replies_json: ["3", "4"]
});
```

Code from this PR transforms above SQL query to:
```sql
SELECT * FROM posts WHERE posts.reply_to IN :replies__0, :replies__1 AND posts.reply_to_json IN :replies_json__0, :replies_json__1
```
...and placeholders to:
```
{
    replies__0: "1",
    replies__1: "2",
    replies_json__0: "3",
    replies_json__1: "4"
}
```

---

I know you can use `?` right now, but that doesn't work with complex queries.

### DIFF
--- a/src/Db/DbCursor.py
+++ b/src/Db/DbCursor.py
@@ -47,6 +47,21 @@ class DbCursor:
                 keysvalues = "(%s) VALUES (%s)" % (keys, values)
                 query = re.sub("(.*)[?]", "\\1%s" % keysvalues, query)  # Replace the last ?
                 params = tuple(params.values())
+        elif isinstance(params, dict) and ":" in query:
+            new_params = dict()
+            values = []
+            for key, value in params.items():
+                if type(value) is list:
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
Hi, thanks, it's a nice addition! There is a small bug: the `re.sub(r":" + re.escape(key) + r"([)\s])", ` does not matches if the query ends with the parameter. 

Eg.: `assert db.execute("SELECT COUNT(*) AS num FROM test WHERE test_id IN :test_id", {"test_id": [1, 2, 3]}).fetchone()["num"] == 3`

PR #1254, opened 2018-01-22
Gratipay shut down so removing

AceLewis:
Gratipay shut down so people can't donate to the project through it
anymore.


HelloZeroNet:
Thanks!

PR #1261, opened 2018-01-27
FileQuery + CORS

imachug:
Hello! I know `fileQuery` is quite an old thing, but it is useful to query some CORS sites that don't have any database.

So this PR adds two options:
1. First, it adds `fileQuery` to `corsFuncWrapper` list, so you can `fileQuery("cors-something...")` now.
2. The second change, which isn't connected with CORS, `fileQuery("file", "posts.slug=home")` is valid now - so you can use a string as value for searching.

### DIFF
--- a/plugins/Cors/CorsPlugin.py
+++ b/plugins/Cors/CorsPlugin.py
@@ -50,8 +50,8 @@ class UiWebsocketPlugin(object):
     def actionOptionalFileInfo(self, to, inner_path, *args, **kwargs):
         return self.corsFuncWrapper("actionOptionalFileInfo", to, inner_path, *args, **kwargs)
 
-    def actionFileQuery(self, to, inner_path, *args, **kwargs):
-        return self.corsFuncWrapper("actionFileQuery", to, inner_path, *args, **kwargs)
+    def actionFileQuery(self, to, dir_inner_path, *args, **kwargs):
+        return self.corsFuncWrapper("actionFileQuery", to, dir_inner_path, *args, **kwargs)
 
     def actionCorsPermission(self, to, address):
         site = self.server.sites.get(address)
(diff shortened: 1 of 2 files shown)


imachug:
@HelloZeroNet Any news?

HelloZeroNet:
The problem is FileQuery will get slow after a certain amount of file, so I not sure if we should encourage using it.

imachug:
I know, but there is no way to query data from ZeroWiki when I add it as CORS.

Kaffiene does use ZeroWiki's database, but it uses JavaScript SQLite.

HelloZeroNet:
Then adding cross-site database query support sounds better for me if it would solve the problem.

imachug:
@HelloZeroNet I can PR that - should it be support for `dbschema.json`:

```
maps: {
    "cors-.*/...": ...
}
```

or should it be a command `corsDbQuery("address", "query", params)`?

HelloZeroNet:
Working on it... I will change the (currently undocumented) "As" command to allow this.
It should not take long :)

imachug:
Thank you!

imachug:
@HelloZeroNet Let me check.

imachug:
Wait - this doesn't work:

```
page.cmd("as", ["138R53t3ZW7KDfSfxVpWUsMXgwUnsDNXLP", "dbQuery", "SELECT * FROM json WHERE json_id = :param", {param: param}])
```

HelloZeroNet:
Try this way: `page.cmd("as", ["138R53t3ZW7KDfSfxVpWUsMXgwUnsDNXLP", "dbQuery", ["SELECT * FROM json WHERE json_id = :param", {param: param}]])`

imachug:
Yeah, it works - thank you for implementing this. Though I think that should be documented somewhere.

HelloZeroNet:
Just added it:

https://zeronet.readthedocs.io/en/latest/site_development/zeroframe_api_reference/#as-address-cmd-arguments

Issue #1302, opened 2018-02-25
Please update installation guide


beenotung:
If I clone this projec then run the zeronet.py, it doesn't work.

Even after I run `sudo pip install -r requirements.txt`, It doesn't works.

```
beenotung:ZeroNet$ ./zeronet.py 
- Starting ZeroNet...
ERROR:root:Unhandled exception: No module named gevent
Traceback (most recent call last):
  File "./zeronet.py", line 18, in main
    import main
  File "/home/beenotung/local/opt/ZeroNet/src/main.py", line 9, in <module>
    import gevent
ImportError: No module named gevent
Traceback (most recent call last):
  File "./zeronet.py", line 91, in <module>
    main()
  File "./zeronet.py", line 64, in main
    traceback.print_exc(file=open(config.log_dir + "/error.log", "a"))
AttributeError: 'Config' object has no attribute 'log_dir'

```

Maybe related to https://github.com/HelloZeroNet/ZeroNet/issues/1056

beenotung:
Solution:
```
sudo pip2 install -r requirements.txt
```

Note that it's using python 2.
I suggest to include a install script in the root folder, so people not familiar with python (2 v.s. 3) can get started easily.

beenotung:
created this pull request: https://github.com/HelloZeroNet/ZeroNet/pull/1303

jebuscraisp:
Hi @beenotung , am curious about this output , is it not connecting to the tor network?

beenotung:
I wonder why the pull reuest is not accepted yet.
It works.

PR #1303, opened 2018-02-25
added guide to install from source code

beenotung:
for this issue: https://github.com/HelloZeroNet/ZeroNet/issues/1302

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,15 @@ Decentralized websites using Bitcoin crypto and the BitTorrent network - https:/
 
 It downloads the latest version of ZeroNet then starts it automatically.
 
+#### Manual install for General Linux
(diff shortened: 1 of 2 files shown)


blurHY:
This can help newbie users.One-click to install&run zeronet

blurHY:
Newbies may don't know what python is.
They just want to start using zeronet as quickly as possible.
We should attract new users by making it easier to use.
Install script and step-by-step tutorial are both needed.
It's not a priority yet,but you can merge this i think.

HelloZeroNet:
For newbies I would recommend to use the bundle package as it does not need anything to do, just unpack and run.

ghost (approved):

PR #1304, opened 2018-02-26
easy ChatCollector collector

CAOLINAN:
less code about collecting sites data every hour


HelloZeroNet:
You got it wrong: it's collecting global data every 5 minute, site data every hour

CAOLINAN:
Sorry, I'm wrong,need to study hard:(

PR #1328, opened 2018-03-13
update Chinese translation (simplified) for new functions

tangdou1:
update zh.json for incoming "Open site directory" function


HelloZeroNet:
Thanks!

PR #1367, opened 2018-03-29
Add a prototype.json

tangdou1:
Add a prototape.json for other languages to check update in SideBar


tangdou1:
Thanks. Hope HelloZeroNet would accept it.

krixano:
Why did you close this if you want it accepted?

HelloZeroNet:
Probably it would be better to generate it automatically as it could be a pain to keep it up to date.

tangdou1:
Yes, it would be a pain. However, I do not have these techniques to generate it automatically. How about close it？

PR #1370, opened 2018-03-31
Update Chinese translation

tangdou1:


tangdou1:
The "Missing files" has been replaced with "Needs to be updated" in code.
So I delete this line "Missing files": "丢失的文件", add "Needs to be updated": "需要更新".

PR #1378, opened 2018-04-06
Fix mergerSiteAdd typo

imachug:
Security fix. Sites could be downloaded to proxies via running `mergerSiteAdd(site_address)` in console while on ZeroMe.


HelloZeroNet:
Thanks

Issue #1467, opened 2018-07-01
Make sidebar donate link optional?


vitorio:
### Step 1: Please describe your environment

  * ZeroNet version: 0.6.3
  * Operating system: OS X 10.11, Windows 10
  * Web browser: Safari, Chrome, Edge
  * Tor status: not available on Mac/available on Windows
  * Opened port: yes

### Step 2: Describe the problem:

As a site owner, I'd like to be able to remove the "Donate" link in the sidebar for my site.

There doesn't appear to be a way to do this through the ZeroFrame API, perhaps it could be added as an optional flag in content.json?  Something like:

```json
"donate": false
```

or

```json
"hide_donate_links": true
```

?

anoadragon453:
The ability to remove it and/or change it would be nice.

Thunder33345:
well the bitcoin address is the site address, the site private key
the same key you used to access/edit the site
thus you already have a bitcoin address believe it or not

anoadragon453:
Yes, but some people prefer to use their main btc address which may 
already be verified by a WoT.

Thunder33345:
the site's address is always site owner, thus anyone can reasonably expect that if i donate to `1HeLLo4uzjaLetFx6NH3PMwFP3qbRbTf3D` the owner of that site can be able to withdraw it

and there's nothing stopping you from donating to any site address
https://www.blockchain.com/en/btc/address/1HeLLo4uzjaLetFx6NH3PMwFP3qbRbTf3D

krixano:
Right, but some people don't want to have to import every current and new zite. They just want to use one for everything.

vitorio:
So extending my original proposal, then, perhaps:

By default, the donate link exists, and it shows the site address against a Bitcoin URL handler.

If `content.json` contains a key of: `"donate": false` or `"donate": ""`, it does not display the site address or link.

If `content.json` contains a key of: `"donate": "1..."`, where that value passes a Bitcoin address validation, it treats it as a Bitcoin address and shows it against a Bitcoin URL handler.

If `content.json` contains a key of `"donate": "http://..."` or `"donate": "https://..."`, where those values are strings beginning with "http://" or "https://", it treats it as a URL and shows only the "Donate" button linking to that.

If `content.json` contains a string with any other contents, it displays it as a block of text, suitable for site owners to provide alternative donation instructions.

Thoughts?

Thunder33345:
actually that's not bad proposal, adding customizability is a +

[...]

PR #1479, opened 2018-07-07
Support removing donation link / custom donation text

vitorio:
Hi, as discussed in issue #1467, I'd like to be able to remove the "Donate" link in the sidebar for my site, and other users thought it might be nice to be able to provide a custom one.

This PR is the simplest implementation.  If a key `donate` is present in `content.json`, either as `"donate": false` or `"donate": ""`, the donation link is not rendered.  If a key `donate` is present in `content.json` and it's a string, it's escaped and rendered, allowing anyone to write out how they'd like to accept donations.  Otherwise, the existing "Donate" link is rendered as it is currently.

This PR does not try to interpret the contents of the string, nor does it provide a UI for editing the donation link string.  This provides just the basic rendering based on the presence or absence of a `donate` key.

### DIFF
--- a/plugins/Sidebar/SidebarPlugin.py
+++ b/plugins/Sidebar/SidebarPlugin.py
@@ -409,16 +409,30 @@ class UiWebsocketPlugin(object):
             </li>
         """))
 
-        site_address = self.site.address
-        body.append(_(u"""
-            <li>
-             <label>{_[Site address]}</label><br>
-             <div class='flex'>
-              <span class='input text disabled'>{site_address}</span>
-              <a href='bitcoin:{site_address}' class='button' id='button-donate'>{_[Donate]}</a>
-             </div>
-            </li>
-        """))
+        donate_key = site.content_manager.contents.get("content.json", {}).get("donate", True)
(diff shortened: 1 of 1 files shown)


imachug:
What about showing both `Site address` and `Donate` if `donate` is present? I mean, I usually check site's address by opening sidebar.

vitorio:
@imachug Looking at the commit history of the sidebar, it looks to me like the Site Address was always attached to the Donate link, I can't see a commit where there's just the Site Address by itself.

That said, that's a pretty reasonable idea, so I've updated the commit.  "Site address" is always shown, and just the "Donate" link itself is affected as previously described:

No `donate` key in `content.json` (default):
screen shot 2018-07-07 at 11 00 41 am

`"donate": false` or `"donate": ""` in `content.json`:
screen shot 2018-07-07 at 10 59 17 am

`"donate": "some string"` in `content.json`:
screen shot 2018-07-07 at 11 00 11 am

imachug:
Thank you for your changes! It looks much better now.

HelloZeroNet:
Thanks!

Issue #1486, opened 2018-07-16
Rev 3535 broke a bunch of sites


rllola:
The latest release broke a bunch of site. My site is not able to load the script and I got this error in the console : 
```
Loading failed for the <script> with source “http://127.0.0.1:43110/mixtape.bit/dist/all.js”.
```

rllola:
I am not able to download the my javascript file. 
If I go here ; http://127.0.0.1:43110/mixtape.bit/dist/all.js

I will see the error which lead me to a BigFilePlugin, It seems that because my `all.js` file is more than 1MB it is using BIgFilePlugin wrapper around it.

Lots of us use heavy javascript libraries in our site. It might be an issue to try to stay under 1MB. 

Should we raise the limit ?

slrslr:
@rllola i think it is fixed by now (after rev3537)? If so, please close the issue?

PR #1487, opened 2018-07-16
Fix rev 3535 big javascript file

rllola:
Related to this issue : 
https://github.com/HelloZeroNet/ZeroNet/issues/1486

`openBigFile` will return false if it is not a big file but just javscript file. Added a condition to check that it is really a big file.


HelloZeroNet:
Thanks!

Issue #1493, opened 2018-07-24
Manage site permissions


rllola:
I gave my site `ADMIN` permission and now I would like to revoke it. How can I do that ? I haven't found a command to do it...

rllola:
There is a function that allow us to do that https://github.com/HelloZeroNet/ZeroNet/blob/master/src/Ui/UiWebsocket.py#L815

But no api so we can call it.

krixano:
It might be good to have a webpage where we can manage all of the permissions of all sizes easily.

rllola:
It could be added in `config` ?

rllola:
Added the remove permission to the api : https://github.com/HelloZeroNet/ZeroNet/pull/1494

rllola:
Just in case someone do the same mistake as me and is wondering how to remove a permission you can use `siteCmd` and in it you have `permissionRemove`. So you can do something like this.

```
python zeronet.py siteCmd 15t2dFCcxamJnPkp5yTFfxGReeFphvmnE4 permissionRemove ADMIN
```

It remove admin permission.

PR #1494, opened 2018-07-25
Remove a permission

rllola:
Allow to remove permissions using zeroframe api

### DIFF
--- a/src/Ui/media/all.js
+++ b/src/Ui/media/all.js
@@ -312,33 +312,33 @@ if (window.getComputedStyle(document.body).transform) {
  * to offer multiple easing options
  *
  * TERMS OF USE - jQuery Easing
- * 
- * Open source under the BSD License. 
- * 
+ *
+ * Open source under the BSD License.
+ *
  * Copyright © 2008 George McGinley Smith
  * All rights reserved.
- * 
- * Redistribution and use in source and binary forms, with or without modification, 
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
  * are permitted provided that the following conditions are met:
- * 
- * Redistributions of source code must retain the above copyright notice, this list of 
+ *
+ * Redistributions of source code must retain the above copyright notice, this list of
  * conditions and the following disclaimer.
- * Redistributions in binary form must reproduce the above copyright notice, this list 
- * of conditions and the following disclaimer in the documentation and/or other materials 
+ * Redistributions in binary form must reproduce the above copyright notice, this list
+ * of conditions and the following disclaimer in the documentation and/or other materials
  * provided with the distribution.
- * 
- * Neither the name of the author nor the names of contributors may be used to endorse 
+ *
+ * Neither the name of the author nor the names of contributors may be used to endorse
  * or promote products derived from this software without specific prior written permission.
- * 
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY 
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
  * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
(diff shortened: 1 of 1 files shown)


rllola:
@HelloZeroNet Do you think it is a good PR or not safe ?

If we don't want it I shall close it.

ghost:
If you prompt the user before doing anything, then I don't see why it wouldn't be safe... Also, not having this makes ZeroNet more unsafe - being able to remove permissions from sites is a good thing. If we don't want it in ZeroFrame then it should definitely be implemented elsewhere (like an additional built-in page to the zeronet client).

HelloZeroNet:
Thanks for the PR, but looks like something is missing as there is no actionPermissionRemove function in the Wrapper class and please edit the .coffee file as all.js will be overwritten by the compiler.

rllola:
Yeah I am not sure what I as doing... But it will not call this ? 
https://github.com/HelloZeroNet/ZeroNet/blob/py3/src/Ui/UiWebsocket.py#L834-L839

However it seems to work only if you have admin permission....

purplesyringa:
You're editing *JavaScript* code, not Python code -- they are not connected directly. TBH I think that we shouldn't add a new `wrapperPermissionRemove` command, it's enough to make `permissionRemove` non-admin. What do you think?

rllola:
> You're editing JavaScript code, not Python code -- they are not connected directly. TBH I think that we shouldn't add a new wrapperPermissionRemove command, it's enough to make permissionRemove non-admin. What do you think?

Yeah, I know that. I was talking about the missing code inside the Wrapper class. For some reason I either though it would magically work or forgot to push the rest of the code. I changed computer since so it will stay a mystery.

This PR is linked to a specific need. A long time ago, I was asking `admin` permission for a site and  realized that was a bad idea and not the best way to accomplish what I wanted to do. So I wanted to remove this permission but the only way to do it is by command line. I don't think it is quite user friendly. See https://github.com/HelloZeroNet/ZeroNet/issues/1493

It could be good to have a way to remove it. It will pop up a notification like for granting permission. It will only be per site...

@imachug I am sorry I am not sure understand the last part and I am interested in hearing your feedback. Could you expand on that?

purplesyringa:
We already have `permissionRemove` command which is not actually used and is an ADMIN-only command. I see no point in adding another action, and I think that it'd be better if we used `permissionRemove` instead. We should make it not ADMIN-only, and maybe add some notification like "You successfully ~~unsubscribed from our emails~~ removed HelloWorld permission".

rllola:
>We already have `permissionRemove` command which is not actually used and is an ADMIN-only command. I see no point in adding another action, and I think that it'd be better if we used `permissionRemove` instead.

I still don't understand you. You mean you don't want to have it available through Zeroframe? You just want it as a command line?

rllola:
Closing this.

purplesyringa:
@rllola Nope. See, we have `siteInfo` command, and we don't need to define an additional `wrapperSiteInfo` command. Also, we had an ADMIN `siteClone` command before but then we made it non-admin (because of Git Center). Same here: instead of adding `wrapperPermissionRemove` we can just make `permissionRemove` non-ADMIN.

ghost:
@rllola  Uh... the ZeroFrame commands are called... *commands*, hence the `zeroframe.cmd()` function. I really don't understand what's not clear to you. `wrapperPermissionRemove` is **already** a *ZeroFrame* command, you just need to remove it from the ADMIN-only list. You don't need to create duplicate commands when they do the same thing.

rllola:
> @rllola Nope. See, we have `siteInfo` command, and we don't need to define an additional `wrapperSiteInfo` command. Also, we had an ADMIN `siteClone` command before but then we made it non-admin (because of Git Center). Same here: instead of adding `wrapperPermissionRemove` we can just make `permissionRemove` non-ADMIN.

Got ya! However if we want to prompt a notification "Do you want to remove this permission?" We might want a `wrapperPermissionRemove`? Or yeah we can just skip asking the question and notify that it has been removed... but I was going for the same model as adding permission.

I am preparing a new PR and small test site with this because this one is not relevant anymore. I will ping you when it is ready.

ghost:
Or you can just add the prompt to the existing permissionRemove command...

PR #1515, opened 2018-08-19
Add pipenv files.

ysc3839:
https://github.com/pypa/pipenv
https://docs.pipenv.org


blurHY:
There are two pipenv PRs here now. @shortcutme

Issue #1528, opened 2018-08-23
-noclone prefix for files that will not be clonned


DaniellMesquita:

DaniellMesquita:
@HelloZeroNet @shortcutme

HelloZeroNet:
What would be you usecase? You can add directories/files with -default prefix that will be renamed on clone. You can check it on ZeroBlog.

DaniellMesquita:
> What would be you usecase?

I'm creating Framework.js, and it will have a button in header where ZeroNet users can clone the site, to create a brand new zite based on FrameworkJS.
But the site have redundant files that can only be used on its own, but not in the clonned sites.
The ZeroNet's site clone feature is amazing, but really needs more improvements.
Not only -noclone prefix is needed, but also -noupgrade, as mentioned in this new issue: #1529

I need to launch FrameworkJS in order to begin other zite projects, so those prefixes for clonning are needed for Framework work in its full, and other projects that can be benefited from those new clonning prefixes.

DaniellMesquita:
@HelloZeroNet @shortcutme

> You can add directories/files with -default prefix that will be renamed on clone.

Thanks, but I know this and already using it.

> What would be you usecase?

I'm creating Framework.js, and it will have a button in header where ZeroNet users can clone the site, to create a brand new zite based on FrameworkJS.
But the site have redundant files that can only be used on its own, but not in the clonned sites.
The ZeroNet's site clone feature is amazing, but really needs more improvements.
Not only -noclone prefix is needed, but also -noupgrade, as mentioned in this new issue: #1529

I need to launch FrameworkJS in order to begin other zite projects, so those prefixes for clonning are needed for Framework work in its full, and other projects that can be benefited from those new clonning prefixes.

DaniellMesquita:
As mentioned in #1531:
Instead of using prefixes in file names (it will make names very ugly), site owner can add those infos in content.json.

HelloZeroNet:
> I'm creating Framework.js, and it will have a button in header where ZeroNet users can clone the site, to create a brand new zite based on FrameworkJS.

There is a root_inner_path argument for siteClone call, where you can specify which directory you want to use as the root of the new site. For example ZeroHello use this feature to create a new, empty site: https://github.com/HelloZeroNet/ZeroHello/blob/master/js/Head.coffee#L37

trenta3:
[...]

Issue #1531, opened 2018-08-23
New site clone prefixes


DaniellMesquita:
The ZeroNet's site clone feature is amazing, but really needs more improvements.

- [ ] -noclone (#1528)
- [ ] -noupgrade (#1529)
- [ ] -upgrade (#1530)

trenta3:
@DaniellMesquita Could you please specify more adequately how exactly one should specify them in `content.json`.
My proposal is to add some keys such as:
```
"default_files": [
  { "to": "config.ini", "from": "config.ini.example" },
  ...
]

"noclone_files": [
  "css/custom.css"
```

DaniellMesquita:
The form for a valid .json file, you knows better than me.

DaniellMesquita:
Well thinked, -default in .json is a very good idea against uggly file names.

PR #1542, opened 2018-08-28
Added noclone suffix for files that should not be cloned

trenta3:
This commit should solve issue #1528
i.e. this does not clone files or directories which have "-noclone" in their names.

I've tested the patch on a couple of sites but I couldn't understand how to adjust the asserts in the Test folder. This means tests for the patch are missing.


DaniellMesquita:
Thanks, really appreciated and will make FrameworkJS more close to be launched!
Could you also support the other suffixes?
https://github.com/HelloZeroNet/ZeroNet/issues/1531
@HelloZeroNet @shortcutme

HelloZeroNet:
Actually I think it should be recorded in content.json and not as prefixes.

trenta3:
I think we should support both options, for uniformity wrt the "-default" infix

DaniellMesquita:
@HelloZeroNet is right about recording it on content.json, but @trenta3 is also right about supporting both.

DaniellMesquita:
Don't forget about other suffixes: https://github.com/HelloZeroNet/ZeroNet/issues/1531.

DaniellMesquita:
@trenta3

trenta3:
This commit ended up defining a different syntax than the hypothized one.
In particular the "default_files" key is an object and not a list.
The keys are the file sources and the values are files' destinations.

I give an usage example:
```
"default_files": {
  "config.ini.example": "config.ini",
  ...
},
"noclone_files": [
  "uninmportant-file.txt",
  ...
]
```

Also note that this only works for files.

@HelloZeroNet I'm unsure how to set priority for noclone and default files setted in `content.json` under `src/Worker/WorkerManager.py#getPriorityBoost`.

trenta3:
@HelloZeroNet I'm also unsure on how to set up tests, since the `1TeST...` folder has to be signed by a key that I couldn't find in the repository...

DaniellMesquita:
@HelloZeroNet I'm unsure how to set priority for noclone and default files setted in content.json under src/Worker/WorkerManager.py#getPriorityBoost

@HelloZeroNet @shortcutme

Issue #1548, opened 2018-08-29
Feature Request: Add URL rewriting to sites


trenta3:
Allow each site to specify an URL rewriting scheme in `content.json`.
Proposed syntax:
```
"rewrite_rules": [
    { match: "files/.*", terminate: true },
    { match: "(.*)", replace: "index.html?url=$1", terminate: true },
]
```

The rewriting rules are:
* The URL which is checked to match against the regexp is the part after the site address, non including initial backslash.
* The matching rules are applied recursively until one with `terminate = true` is found. The first matched rule is the one which is applied at each step, each time starting from the first one.
* If the `replace` key is missing the returned string is not modified.

The final url should be then internally rendered and such page returned.
There should not be any html redirect.

If after 100 rewrites the url did not find a terminate condition, the site will log an error and the original request will be made.

PR #1549, opened 2018-08-29
Added basic URL rewriting to sites

trenta3:
Solves issue #1548 
Please note that the currently implemented form differs slightly from the specifications in the issue

Usage example: in content.json add the key rewrite_rules:
```
"rewrite_rules": [
  {
   "match": "index.html",
   "terminate": true
  },
  {
   "match": "files/.*",
   "terminate": true
  },
  {
   "match": "(.*)",
   "replace": "index.html",
   "replace_query_string": "url=$1",
   "terminate": true
  }
 ]
```

Now the requests are transformed as follows:
```
index.html?x=5 --> index.html?x=5
files/example.txt --> files/example.txt
example.html --> index.html?url=example.html
example.html?x=5 --> index.html?url=example.html
```

Caveats:
* `match` is mandatory, `terminate` defaults to false, `replace` and `replace_query_string` defaults to the global match
* The URL to be checked agains `match` is only the part after the site address, not including the initial backslash and not including the query string
* `replace` is the expression to replace the base part
* `replace_query_string` is the expression to replace the query string part
* There is no way as of now to preserve query string parameters in rewriting


trenta3:
Please note that i've rebased the PR and now all tests pass

trenta3:
Please do not merge it yet, since I would like to review the changes and also implement other possibilities:
* Custom return code (non-200)
* Matching on full url with querystring

@HelloZeroNet I'm dubious on wheter the changes have to go in the place where I put them or before line 290 in `UiRequest.py`

imachug:
Not sure, maybe we should make it a plugin to separate main core from additional features?

trenta3:
I would have done it yet, if I knew how to do it. I see that existing
plugins only patch some functions either at the beginning or at the end.

I also don't know where exactly to put the redirection function in the
existing code...

trenta3:
I think the rewrite has to occurr in the middle of a function

imachug:
You may find the information you're looking for at [DevCenter.bit](http://127.0.0.1:43110/DevCenter.bit), and you can have a look at my [BackgroundProcessing plugin](https://github.com/HelloZeroNet/Plugin-BackgroundProcessing) which adds new functionality.

trenta3:
I'm trying to port it as a plugin. I'm almost done, except it does not work for `.bit` domains, because I need the plugin `Zeroname` to load before mine. Is there any way I can do this?

imachug:
@trenta3 Can you show me the code where you use Zeroname?

imachug:
@trenta3 You can use:

```python
try:
    import Zeroname
except ImportError:
    # ZeroName plugin is disabled
```

trenta3:
@imachug I'm not using it directly. I just need that it loads before mine, since it patches `SiteManager` with `domainname -> address` resolution that I need to get to read the `content.json` files.

trenta3:
An example:
Suppose someone wants to visit `zeroid.bit/`.
I need to get its `content.json` file to see if it specifies some redirect.

Unfortunately i see, reading the `PluginManager.py`, that the plugins are loaded in alphabetical order, so it is very unfortunate that what I need is the `Zeroname` plugin :)

It would be nice to be able to add dependencies to the plugin system.

trenta3:
@imachug Thank you. Now it works. I add to be careful loading `Zeroname` befor `SiteManager`.

```python
try:
    import Zeroname
except ImportError:
    # ZeroName plugin is disabled

from Site import SiteManager
```

trenta3:
@HelloZeroNet I also had to load the `TranslateSite` plugin before `Zeroname` because otherwise it raises an exception on sites that are accessed as domain name:
```
UiWSGIHandler error: AttributeError: 'NoneType' object has no attribute 'content_manager' in UiServer.py line 40 > pywsgi.py line 923 > pywsgi.py line 907 > TranslateSitePlugin.py line 39
```

trenta3:
I think the feature is now mature enough to pass to the testing phase.
I've already tested it on some sites.

@imachug @DaniellMesquita @Thunder33345 @blurHY Please test it and report errors.

I've added documentation in the form of [a Markdown file](https://github.com/HelloZeroNet/ZeroNet/blob/64917273341f783734d594dbe76857b3662c7f84/docs/plugins/RewriteRequest.md).
Please read it and try this Plugin.

trenta3:
If someone tried, please update the code with my latest commit, since I had left in the code something that completely bypassed my own code :)

Now the plugin even supports conditional checks on file existance before applying a rewrite rule.
[Updated documentation](https://github.com/HelloZeroNet/ZeroNet/blob/0221df99cbb2069f618707b941f6ca865a121657/docs/plugins/RewriteRequest.md)

DaniellMesquita:
@trenta3 
What are the fields I put on content.json to test it?

trenta3:
Please click on the Link for documentation in my previous message. It has
detailed description and two usage examples.

The key is 'rewrite_rules' and should be a list of dictionaries with some
Keys. Read docs.

>

trenta3:
Also thank you for trying it. Please write about every problem you have or
lacking of proper documentation

>

trenta3:
@DaniellMesquita Long time has passed, did you try it? What are your findings?

PR #1606, opened 2018-09-14
Bug Fixs

tangdou1:
They do not match. Beacuse ip and port in (ip, port) are unicode type while ip in peer_blacklist is string type and port in  in peer_blacklist is int type.


HelloZeroNet:
Thanks for reporting I added a different fix in the latest rev as it could have other side-effects if we store the ip/port other than str/int: https://github.com/HelloZeroNet/ZeroNet/commit/1eacc26ef9d35988b6e45650bf54abd5bc11d603

PR #1619, opened 2018-09-17
IPV6 Supports in another way

tangdou1:
Stil test.


DaniellMesquita:
Why another pull request?

tangdou1:
> 
> 
> Why another pull request?

I want to make it more consistent with zeronet code style, so that HelloZeroNet would accept my pull request.

tangdou1:
I still like the smallest change to implete IPV6, I wont test it again.

PR #1635, opened 2018-09-21
Fix bootstrapper is blocked by itself

tangdou1:
Make the bootstrapper server can add its own ip and port in its own bootstrapper.db as a peer.


DaniellMesquita:
@HelloZeroNet @shortcutme

HelloZeroNet:
Thanks

PR #1683, opened 2018-10-06
Allow using other domain resolver plugins along with Zeroname

geekless:
Fixes several closely related bugs in Zeroname Plugin, those prevent correct work of other domain resolvers:

1. get(), isDomain(): when an address doesn't look like a valid domain, pass control to the next plugin in a chain.

3. Move the .bit-domain matching logic to a separate method isBitDomain().

2. get(), need(): use isBitDomain(), not isDomain() to check if domain resolving is needed in order to not interfere with other domain resolvers.

4. Also rewrite isAddress() to make it look similar to the new implementation of isDomain().

### DIFF
--- a/plugins/Zeroname/SiteManagerPlugin.py
+++ b/plugins/Zeroname/SiteManagerPlugin.py
@@ -21,15 +21,16 @@ class SiteManagerPlugin(object):
         if not self.get(config.bit_resolver):
             self.need(config.bit_resolver)  # Need ZeroName site
 
-    # Checks if its a valid address
+    # Checks if it's a valid address
     def isAddress(self, address):
-        if self.isDomain(address):
-            return True
-        else:
-            return super(SiteManagerPlugin, self).isAddress(address)
+        return self.isBitDomain(address) or super(SiteManagerPlugin, self).isAddress(address)
 
     # Return: True if the address is domain
     def isDomain(self, address):
+        return self.isBitDomain(address) or super(SiteManagerPlugin, self).isDomain(address)
+
+    # Return: True if the address is .bit domain
+    def isBitDomain(self, address):
         return re.match("(.*?)([A-Za-z0-9_-]+\.bit)$", address)
 
     # Resolve domain
@@ -54,7 +55,7 @@ class SiteManagerPlugin(object):
     # Return or create site and start download site files
     # Return: Site or None if dns resolve failed
     def need(self, address, *args, **kwargs):
-        if self.isDomain(address):  # Its looks like a domain
+        if self.isBitDomain(address):  # Its looks like a domain
             address_resolved = self.resolveDomain(address)
             if address_resolved:
                 address = address_resolved
@@ -67,7 +68,7 @@ class SiteManagerPlugin(object):
     def get(self, address):
         if not self.loaded:  # Not loaded yet
             self.load()
-        if self.isDomain(address):  # Its looks like a domain
+        if self.isBitDomain(address):  # Its looks like a domain
             address_resolved = self.resolveDomain(address)
             if address_resolved:  # Domain found
                 site = self.sites.get(address_resolved)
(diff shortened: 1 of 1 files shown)


DaniellMesquita:
It allows creating a plugin for Emercoin

DaniellMesquita:
# How it works
When accessing kaffie.zeroid, it points to the zite that haves the same identy address of user "kaffie".

# How subdomain works
Simply put those records on content.json:

>  "domain_records": {
>   "blog.kaffie.zeroid": "address",
>   "id.kaffie.zeroid": "address",
>   "search.kaffie.zeroid": "address"
>  },

@April93
Will you use it?

imachug:
Though, it will probably be `kaffie@zeroid.bit`, not `kaffie.zeroid.bit` because it'll be a subdomain, and not `kaffie.zeroid` because `zeroid` is not a valid cert provider.

DaniellMesquita:
@imachug 
> Though, it will probably be kaffie@zeroid.bit, not kaffie.zeroid.bit because it'll be a subdomain, and not kaffie.zeroid because zeroid is not a valid cert provider.

You didn't understood how NameID works

imachug:
Let me bring another example. `gitcenter.kaffie`. 1st, `kaffie.bit` is auth provider but not the site of auth provider. So somehow (how?) ZeroNet should guess it's `id.kaffie.bit`. 2nd, there isn't one database on KaffieID. So does it mean that NameID won't work with KaffieID?

DaniellMesquita:
Not possible to use kaffie ID, because the names aren't unique

Thunder33345:
the using names as domain thing will be hardcoded to only support zeroid and nothing more regardless if registry is unique or not

HelloZeroNet:
The zeroid provider name can be anything, it doesn't have to be live site.

geekless:
Not sure why we are discussing the IDName stuff in that pull request, but...

@imachug:

> Though, it will probably be kaffie@zeroid.bit, not kaffie.zeroid.bit because it'll be a subdomain, and not kaffie.zeroid because zeroid is not a valid cert provider.

The .bit suffix should be avoided, since it is related to the Namecoin address space. So `kaffie.zeroid.bit` is the impossible option. The same for `kaffie@zeroid.bit`.

My first thought was to use `.zeroid.bit` suffixes, but I gave up that idea, as it brings ambiguity to the domain space.

> not kaffie.zeroid because zeroid is not a valid cert provider.

There is no direct connection between the provider name and the top-level domain name. It should be a subject of agreement. If you recognize `.zeroid` domain names as provided by that specific provider, and I do the same, we can refer to those domains in unambiguous way.

@Thunder33345:

> the using names as domain thing will be hardcoded to only support zeroid and nothing more regardless if registry is unique or not

Any trusted ID provider, that exports its database in a machine-readable way, can be easily supported. I see no reasons why it shouldn't be done.

> say i have thunder33345.zeroid. how would i "point" my username to a site??

Your cert address **is** your site address. You don't have to do any additional "pointing".

> but that means now if i want to visit test.thunder33345.zeroid, i also have to download my main site, effectively forcing users to download it to resolve the subdomains which probably isnt a good idea either

Do you have any other ideas, how it should be done?

geekless:
@imachug:

Trying to recall, why I gave up `user@zeroid.bit` domains too.

The `.bit` suffix gives the false impression, that the domain belongs to the Namecoin space and can be misleading. It is better to distinguish domain types by their last part. It's more clear for human beings and it also simplifies the code.

By the way, the current implementation of Zeroname still isn't compatible with such domains, as it uses regexp `(.*?)([A-Za-z0-9_-]+\.bit)$`.


What is your opinion?

Thunder33345:
maybe zero.id?
any host like xko.id polar.id
i think using a unified TLD will make things easier when others want to integrate with their plugins
we could have either use 

- .zeroid

others would have to make their own TLD for each user registry which might get messy

- .zero.id

id as TLD anyone can use (servicename).id
probably need community consensus to merge/add their own service to the plugin

PR #1717, opened 2018-10-16
Make it compatible with pyinstaller for browser build

rllola:
We need absolute path for plugin so that pyinstaller can work. It is needed for the browser using PyQt5 to work (https://github.com/rllola/ZeronetBrowser). It should not affect how Zeronet works in general.

### DIFF
--- a/src/Plugin/PluginManager.py
+++ b/src/Plugin/PluginManager.py
@@ -19,7 +19,7 @@ class PluginManager:
         self.plugin_names = []  # Loaded plugin names
         self.after_load = []  # Execute functions after loaded plugins
 
-        sys.path.append(self.plugin_path)
+        sys.path.append(os.path.join(os.getcwd(), self.plugin_path))
         self.migratePlugins()
 
         if config.debug:  # Auto reload Plugins on file change


imachug:
Instead of `getcwd`, we should probably use `__file__`.

rllola:
Ok. What is the difference ?

imachug:
@rllola `os.getcwd()` is the current *working* directory, while `os.path.join(os.path.dirname(os.path.basename(__file__)), "..", "..")` will be the ZeroNet directory. I'll explain it.

*cwd* aka *current working directory* is the path from which the script was run. Usually it matches the directory where the script is.

Pretend `~/ZeroNet/zeronet.py` outputs the *cwd*. Then, here are some examples:

```
~$ cd ZeroNet
~/ZeroNet$ python zeronet.py
/home/ivanq/ZeroNet
~/ZeroNet$ cd ..
~$ python ZeroNet/zeronet.py
/home/ivanq
~/ZeroNet$ cd /
/$ python ~/ZeroNet/zeronet.py
/
```

So, the *cwd* matches the directory you are currently in. It usually matches the correct directory if you're executing the script via UI, though.

And for `os.path.join(os.path.dirname(os.path.basename(__file__)), "..", "..")`, it'll always work, since `__file__` is the current script that's being executed. Since we use it inside `src/Plugin/PluginManager.py`, `__file__` will be `something/src/Plugin/PluginManager.py`, so we need to go two levels up to reach `src`.

rllola:
Thank you for the details. It makes sense. 

I figured out that because in the `zeronet.py` we modify the working directory to be the one with the python script it would be the same : 
https://github.com/HelloZeroNet/ZeroNet/blob/master/zeronet.py#L15

Do you believe it might be a problem in the future ?

imachug:
I think we are fine with `os.getcwd` then, since a lot of other code probably depends on this.

PR #1731, opened 2018-10-20
Multiple changes

mrmajik45:
Unneccessary `;`

Missing vendor-prefixed CSS gradients for Old Webkit (Safari 4+, Chrome) (2)


HelloZeroNet:
Thanks, but the edited files are third-party libs, which I would not touch if not necessary.

PR #1766, opened 2018-11-05
Fix `from Config import config`

imachug:
Always run `from Config import config` inside `PluginManager.afterLoad`.


HelloZeroNet:
The Config class does not accept plugins , so putting it under "@PluginManager.afterLoad" is not necessary.
(and it's also got imported by pluginmanager so it makes not difference)

DaniellMesquita:
I think this removes the need of importing config on every plugin, opening possibility for plugins to have their own settings to appear on /Config

krixano:
@imachug

PR #1806, opened 2018-12-08
Define "Content-Encoding" header for encoded files

gqgs:
If the file is encoded returns the _encoding_ in the [Content-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding) header to allow proper decoding in the browser.

This should make it easier to distribute compressed files in the network.

### DIFF
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -180,6 +180,15 @@ class UiRequest(object):
 
         return content_type
 
+    # Get encoding by filename
+    def getContentEncoding(self, file_name):
+        encoding = mimetypes.guess_type(file_name)[1]
+
+        if encoding:
+            encoding = encoding.lower()
+
+        return encoding
+
     # Return: <dict> Posted variables
     def getPosted(self):
         if self.env['REQUEST_METHOD'] == "POST":
@@ -637,6 +646,7 @@ class UiRequest(object):
         if file_size is not None:
             # Try to figure out content type by extension
             content_type = self.getContentType(file_path)
+            encoding = self.getContentEncoding(file_path)
 
             range = self.env.get("HTTP_RANGE")
             range_start = None
@@ -650,6 +660,8 @@ class UiRequest(object):
                 extra_headers["Accept-Ranges"] = "bytes"
                 if header_length:
                     extra_headers["Content-Length"] = str(file_size)
+                if encoding:
+                    extra_headers["Content-Encoding"] = encoding
                 if range:
                     range_start = int(re.match(".*?([0-9]+)", range).group(1))
                     if re.match(".*?-([0-9]+)", range):


HelloZeroNet:
Can you please describe what's your usecase? This way it would return gzipped content-encoding for tar.gz files, but it's also returns application/tar as content-type which is not displayable by the browser.

So I don't see any use case when it's returns content-type that is displayable by the browser.

gqgs:
@HelloZeroNet The main use case is compressing static assets.
For instance, with this modification it would be possible for a owner to compress the `css.all`/`js.all` files and using appropriate files names (i.e. `css.all.gz`/`js.all.gz`) the content would be decoded and interpreted correctly in the browser.

For sites there are updated frequently this would significantly decrease the size of the payload that is delivered to users after each update.
Or am I missing something and a gzip compression for these cases is already happening somewhere else?

HelloZeroNet:
I have tried it and it do works this way. My only problem is that it would mess up distributing compressed files: eg if you have link that points to myapplication.tar.gz, then it would return myappliaction.tar as the content which is unfortunate.

But in ZeroNet you can reach  the packed files (.zip and .tar.gz) content as a directory, so I recommend to use `<script src="all.zip/all.js">`

gqgs:
>reach the packed files (.zip and .tar.gz) content

That hack does work. Thanks.
With that the result should be effectively the same.

Just one last thing I didn't understand is how defining this header would affect the distribution of compressed files.
I tested and fetching some `*.tar.gz` files as you mentioned and apparently everything behaved as expected.
For the browser it made no difference. I will just initialize a fileGet request / download dialog with the provided `*.tar.gz` filename.

The only difference I spotted is that, as expected, the `Content-Encoding: gzip` will now be returned in the request response.
```
Accept-Ranges: bytes
Cache-Control: no-cache, no-store, private, must-revalidate, max-age=0
Connection: Keep-Alive
Content-Encoding: gzip
Content-Length: 11835546
Content-Type: application/x-tar
Date: Sun, 09 Dec 2018 15:51:04 GMT
Keep-Alive: max=25, timeout=30
Version: HTTP/1.1
X-Frame-Options: SAMEORIGIN
```

In the master branch the content type being returned is the same.
```
Accept-Ranges: bytes
Cache-Control: no-cache, no-store, private, must-revalidate, max-age=0
Connection: Keep-Alive
Content-Length: 11835546
Content-Type: application/x-tar
Date: Sun, 09 Dec 2018 15:47:14 GMT
Keep-Alive: max=25, timeout=30
Version: HTTP/1.1
X-Frame-Options: SAMEORIGIN
```

It seems to me the reply above could led to some ambiguity since there is nothing to distinguish between compressed and uncompressed files, e.g.
```
>>> mimetypes.guess_type("file.tar")
('application/x-tar', None)
>>> mimetypes.guess_type("file.tar.gz")
('application/x-tar', 'gzip')
```

HelloZeroNet:
I did a test:
 - Created a test.tar.gz file (size: 15 087B)
 - Created a website with the content `<a href="test.tar.gz">test.tar.gz</a>`
 - Without patch applied clicking the link results download of a file sized 15 087B
 - With the patch applied clicking the link results download of a file sized 15 177B

What happened is the browser unpacked the .gz file during the download. It can be problematic as this way the file hash value will be different and it will be an invalid tar.gz file (as it's actually a .tar file now): 

```
$ tar xvpfz test.tar.gz
gzip: stdin: not in gzip format
```

gqgs:
That's interesting. I was not expecting the browser unpacking the file to affect what is being stored by the node.
Still the `Content-Type: application/x-tar` seems incorrect in this case.
Trying to get the mime type from the file magic number (and then fall backing to the filename if the previous method failed) might be a more accurate way in order to yield the correct type.
But then I'm not sure if the browser would try decompressing it again in this case.

Either way the `<script src="all.zip/all.js">` hack should work fine for my use case. Appreciated.

PR #1814, opened 2018-12-10
Added issue templates for suggestions

Thunder33345:
also please squash this PR, i made a mess for begin too lazy to use git and got me more attempts that should take to get it right

the feature request could probably have more work done in it though...


HelloZeroNet:
Looking great, I didnt know this was possible:
https://github.com/HelloZeroNet/ZeroNet/issues/new/choose

Thunder33345:
i am not sure of the suggestion template it's rather generic
maybe ill work something about it

PR #1819, opened 2018-12-12
Allow IP addresses via WebSocket API

### LINKED ISSUE #1799: Websocket connection with HTTP_HOST "127.0.0.1" denied access
Trying to make a websocket connection from Godot to ZeroNet fails by default as the `HTTP_HOST` env property of the connection is `127.0.0.1` instead of `127.0.0.1:43110`. This is likely an issue on Godot's websocket library side, but would adding `127.0.0.1` and `localhost` to the list of default allowed hosts be a problem in any way?

As a reminder, the current default values are `zero`, `127.0.0.1:zeronet_port` and `localhost:zeronet_port`.

anoadragon453:
Fixes #1799 

Allows a websocket connection originating from any IP address to connect. `allowed_hosts` is still in place for anything other than an IP address. This is implemented under the assumption that IP addresses are not subject to DNS rebinding attacks.

IP addresses have also been removed from the list of default allowed_hosts as they are inherently allowed by this new rule.

### DIFF
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -4,8 +4,7 @@ import os
 import mimetypes
 import json
 import cgi
-import ipaddress
-import sys
+import socket
 
 import gevent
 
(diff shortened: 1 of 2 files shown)


HelloZeroNet:
Thanks for the PR

The ipaddress module is only available on python 3.3+ default, so we need to include it to lib dir, use regexp/socket.inet_aton based validating or wait until zeronet is ported to py3.

Also it would be more efficient to add the ip address to self.server.allowed_hosts after the successful test to avoid later re-testing and save some cpu.

And please rename it to "isIp" to keep it consistent with eg. helper.isPrivateIp

anoadragon453:
Good feedback, thanks. I'll switch it to use socket.

HelloZeroNet:
The problem with that is it's ipv4 only, but we can change it to ipaddress module after we moved to py3

anoadragon453:
Perhaps I can't just `if python3: import ipaddress`?

HelloZeroNet:
we will drop support of python2 with the move, so I think it's not necessary

anoadragon453:
Ok, guess I can drop the check in the code as well.

anoadragon453:
Made the requested changes!

anoadragon453:
Waiting on this to release the Godot asset :D

anoadragon453:
Thank you! :heart:

Issue #1773, opened 2018-11-09
Necessary code changes to support Python3


HelloZeroNet:
### General
- [x] Run [2to3](https://docs.python.org/2/library/2to3.html) on every file to automatically apply language changes
- [ ] Add GPL compatible headers (?) to every source code
- [ ] Change % style variable formatting to .format and `_("...")` style (to also get automatic escaping)

#### main.py:
  - [x] Review ssl/old gevent patches
  - [ ] Move configuration dependent actions (stack size/socket patching) to separate function in Config.py

#### Content/ContentManager.py:
  - [ ] Too heavy: Move verification code to separate ContentVerify.py

#### Crypt/CryptBitcoin.py:
  - [x] Remove old style bitcoin signature code (BitcoinECC lib)

#### Crypt/CryptConnection.py:
  - [x] Remove dead ECC code

#### File/*
  - [ ] Move to Peer directory as PeerServer and PeerRequest to make it more concrete

#### Peer/PeerServer.py
  - [x] Move port checking code to separate PeerPortcheck.py

#### Ui/UiRequest.py
  - [ ] Requires major rebase, remove outdated /media handling

#### Ui/UiWebsocket.py
  - [ ] Move actions to UiWebsocketAction.py

### Dependencies
- [x] bencode
  - The current library is not available on PIP, possible top3 replacements: bencode, bencode-python3, python-bencode
  - After running some test:
    - The current lib is 2x faster on encode, but 2x slower on decode
    - Py3 is 2x faster on encode and 4x faster on decode using the bencode pip lib
- [x] BitcoinECC
  - We can remove this, as it's only used for the old (invalid) site addresses.
- [x] cssvendor
  - Not a third party lib, created to transform special css tags to vendor specific ones.
  - Should be compatible with python3
- [x] geventwebsocket
  - Python3 compatible according to docs
- [x] merkletools
  - Python3 compatible according to docs
- [x] opensslVerify
  - Not a third party lib, used to interact with openssl so/dll using ctypes.
  - Probably it's not Python3 compatible as it is.
  - [x] Need check how hard to translate it to python3, Result: I was able to port it to Python3
  - [ ] It could use some refactoring: The code is found at pastebin and the quality is not the best.
  - [ ] Openssl 1.1 compatibility?
  - Possible replacements:
    - https://pyopenssl.org/
      - Pros: Openssl 1.1 compatibility, Robust API
[...]

PR #1833, opened 2018-12-25
Make code compatible with Python 2.7 and Python >=3.6 (Syntax: fix exceptions)

ilyeSudo:
Workflow:
- Py2 > Py2&3 Compatible > Py3

Process:
- Feature-oriented PRs (Feature-by-Feature migration)

Steps (5):
- Syntax Fixes: lists/iterators handling, exceptions, octal literals, etc.
- Semantic Fixes: Unicode literals, division, hybrid comparisons, etc.
- Env switch: add py3 & cut/keep py2 support
- More Fixes: fix remained bugs (because it's painful, you guessed it...you are damn right!)
- Cleanup (optional): get rid of compatible code such as Try/Except imports, TODOs, etc.

Challenges:
- Contributors may still write py2 code which will cause me rebasing my branch on master forever. That's why we need the `python3_checks.sh` file in place.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ install:
  - pip list
 before_script:
   - openssl version -a
+  - bash tools/python3_check.sh
 script:
  - python -m pytest -x plugins/CryptMessage/Test
  - python -m pytest -x plugins/Bigfile/Test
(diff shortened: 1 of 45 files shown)


HelloZeroNet:
Hello thanks for the PR, but right now I'm working on moving to Python3 that will include other, general refactoring and some changes in used libraries. By this move the Python2 compatibility will be dropped (I think it safe to do as it will go EOL next year). https://github.com/HelloZeroNet/ZeroNet/issues/1773

ilyeSudo:
Alright. Best of luck then :ok_hand: don't hesitate to ask if you need any help!

Issue #382, opened 2016-03-20
Packaging


WeirdCarrotMonster:
I think that zeronet can be packaged via setuptools and uploaded to PyPI instead of using custom update scripts. You will only need to create setup.py file.

Pros:
- Official way to install python packages
- Easier to install (pip install zeronet, dependencies are downloaded automagically)
- Easier to update (pip install --upgrade zeronet)
- [Command line scripts](https://python-packaging.readthedocs.org/en/latest/command-line-scripts.html)

Cons:
- Some modifications to source code (mostly imports) may be required

If you think that such modifications can be useful, i can work on this issue and create PR later.

almet:
I would love to see this happening!

HelloZeroNet:
Currently ZeroBundle is the default install method it contains all of the liblaries the you need to run zeronet. (SSL+Python is a sad story: had lots of problems with different openssl, gevent and python versions that are incompatible with eachother) It also has built-in update method that downloads the latest version from github, unpacks it and restart itself.

almet:
- For SSL + Python, does anything prevents us to pin the versions (like what you're currently doing) in the `setup.py`?
- For the update mechanism, I believe we could do it also with the setup.py. Could you point to the code that's doing this? Thanks!

HelloZeroNet:
It's hard to pin version, because it's also depends on OpenSSL and Python version which is supplied by the OS. The updating is done by `update.py`

WeirdCarrotMonster:
The package itself can be installed in bundle, the only thing that changes is update method - instead of downloading master branch from github, script will use pip to install latest released version.

On the other hand, it won't work with plugins in their current implementation - but if you ask me, marking plugins as disabled by prefixing "disabled-" is bad idea. I don't know much about good practices for pluggable modules, so i'll look into it for now.

almet:
>I don't know much about good practices for pluggable modules, so i'll look into it for now.

In python packaging there is the [concept of entrypoints](http://stackoverflow.com/questions/774824/explain-python-entry-points) which can be installed in separate packages and listed by another package.

This is exactly done for such cases. Then we could enable/disable modules based on a configuration file, for instance (and default configuration associated).

WeirdCarrotMonster:
[Here is full example](http://stackoverflow.com/a/9615473/3315512).

We can agree on common entrypoint name (i.e. "zeronet") and expect function to return plugin object.

FedericoCeratto:
[...]

PR #1836, opened 2018-12-28
pipenv support

Serkan-devel:
The future of package management in Python

https://pipenv.readthedocs.io/en/latest/

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,18 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+gevent = ">=1.1.0"
+msgpack = ">=0.4.4"
+
+[scripts]
+app = "python zeronet.py"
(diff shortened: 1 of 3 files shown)


Serkan-devel:
Actually, I forgot some things

Serkan-devel:
I thought ZeroNet can be python packaged already, but there is a `setup.py`-file missing for doing so https://github.com/HelloZeroNet/ZeroNet/issues/382 .

That way, one could have pointed to the `requirements.txt`-file to manage all dependencies.

Sorry for letting you wait long for nothing

radfish:
Pipenv seems to be missing the fact that the top-level dependencies (without frozen version) are tracked inside setup.py. The package defined in setup.py is self-contained (installable without information about dependencies from any other files) and orthogonal to the frozen environment definition (optionally) kept in requirements.txt. Pipenv looked intriguing, but after looking into it, I don't see the value in adopting it. Setuptools (setup.py) and pip freeze is sufficient, imo.

PR #1848, opened 2019-01-06
Fix ResourceWarning: unclosed file, prevent resource leaks

BoboTiG:
Signed-off-by: Mickaël Schoentgen <contact@tiger-222.fr>


blurHY:
Good. Now I know why my zeronet spider always warn that there are too much open files

blurHY:
@shortcutme Review and merge this,please

BoboTiG:
I am still working on to fix tests. Should be OK in a couple of minutes :)

HelloZeroNet:
As far as i know its not necessary as python automatically closes the unreferenced files. Can you provide an example when it's not happening?

BoboTiG:
This is almost right. I [quote](https://stackoverflow.com/a/21060677/1117028) a CPython core dev:

> This ResourceWarning means that you opened a file, used it, but then forgot to close the file. Python closes it for you when it notices that the file object is dead, but this only occurs after some unknown time has elapsed. Thus in recent versions, Python also prints a ResourceWarning when it does that. It is a way for you to quickly identify where the unclosed files are, and properly close them. It might be important on some platforms which cannot have more than N files opened at the same time (e.g. 1024). Also, specifically on Windows, you cannot do some operations with a file if it's still open (e.g. deleting it).

I think it is always good to do things in a clean way, but you have the last word :)

imachug:
In *Python* -- yes, it may work like this. However, if you open a file via Python and try to delete it via Explorer, it won't let you do this. This has caused some problems before for me, I couldn't delete a site I recently created. Not sure if you can reproduce it, but this should be fixed.

imachug:
In *Python* -- yes, it may work like this. However, if you open a file via Python and try to delete it via Explorer, it won't let you do this. This has caused some problems before for me, I couldn't delete a site I recently created. Not sure if you can reproduce it, but this should be fixed.

HelloZeroNet:
Please update with the non-deletable file name (db file, optional file, content.json, *.json, or other resource) if this issue if this happens again

imachug:
I think it was a database that caused this.

Issue #1843, opened 2019-01-04
ZeroName updater should use "map" field for subdomains


JeremyRand:
**Is your feature request related to a problem? Please describe.**

In [Namecoin's domain names specification](https://github.com/namecoin/proposals/blob/master/ifa-0001.md#abstract-constructs), the JSON `"map"` field is used to designate subdomains.  Unfortunately, ZeroNet is using a nonstandard subdomain designation via setting the `"zeronet"` field to a JSON object.  This mismatch makes it substantially more difficult for Namecoin-related software (such as ncdns) to properly interoperate with ZeroNet (e.g. the draft ncdns feature that interoperates with ZeroNet's transproxy feature).

(Note that I discussed this subject with Tamas at Decentralized Web Summit 2018.)

**Describe the solution you'd like**

* The ZeroName updater should walk subdomains via the `"map"` JSON field.
* The ZeroName updater should accept `"zeronet"` fields whose value is a string instead of a JSON object.
* It's fine with me if the ZeroName updater continues to accept the existing nonstandard notation as a fallback, so that existing sites don't break due to the update.

**Describe alternatives you've considered**

* I considered requesting that the ZeroName updater discontinue support for the existing nonstandard notation, but I'm fine with pushing back that discussion for a future date, since it's higher priority for me that the standard notation work properly in ZeroNet, and breaking backwards compatibility would probably entail extra delay and work in making sure that everyone updates.
* Long-term I'd suggest that ZeroNet consider the possibility of using a JSON field in Namecoin that maps to a DNS record type (e.g. a `TXT` record), but the discussion about that would probably be more involved than the simple change that I'm requesting here, so I'm fine with pushing back that discussion for a future date.

rllola:
As far as I understand the issue today Zeronet register subdomains using this format : 
```
{
    "zeronet": {
        "": "19rXKeKptSdQ9qt7omwN82smehzTuuq6S9", 
        "staging": "1MzSDy3x8cFrt3P2YkXNN7ogYDFfKX6hnP"
    }
}
```
but the standard way would be : 
```
{
    "map": {
        "": { "zeronet":"19rXKeKptSdQ9qt7omwN82smehzTuuq6S9"}, 
        "staging": {"zeronet":"1MzSDy3x8cFrt3P2YkXNN7ogYDFfKX6hnP"}
    }
}
```

[...]

PR #1856, opened 2019-01-13
Support namecoin standard 'map' for subdomain name

rllola:
See #1843

### DIFF
--- a/plugins/Zeroname/updater/zeroname_updater.py
+++ b/plugins/Zeroname/updater/zeroname_updater.py
@@ -40,10 +40,15 @@ def processNameOp(domain, value, test=False):
         #    ....
         # }
             new_value[""] = data["zeronet"]
-        if len(newValue) > 0:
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
Thanks, please use other variable name instead of map (eg. data_map) as it's a built-in function in python and changing it could have some side-effects and rename newValue to new_value to make the variable names consistent within the function.

rllola:
Done.

I have also added a new case I forgot https://github.com/HelloZeroNet/ZeroNet/pull/1856/files#diff-ccde09c3a724251c612f29d91a7751d7R47

It is when you only have : 
```
{
   "zeronet":"19rXKeKptSdQ9qt7omwN82smehzTuuq6S9"
}
```

I choose to reformat the namecoin standard in the zeronet one and recall the function so I don't have to modify anything in the part where you save it. I hope this is fine. I prefer to make a minimal modification.

PR #1882, opened 2019-01-29
Fix atexit not working on SIGTERM

blurHY:
I am working on HorizonManager that manages the process of HorizonSpider and zeronet.
It would terminate zeronet by sending SIGTERM.
However, atexit won't work on SIGTERM.We have to manually call sys.exit to trigger atexit.

### DIFF
--- a/zeronet.py
+++ b/zeronet.py
@@ -11,7 +11,6 @@ def main():
 
     main = None
     try:
-        import atexit
         from signal import signal, SIGTERM
 
         signal(SIGTERM, lambda signum, stack_frame: sys.exit(0))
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
Looking great, but please remove the unnecessary import of atexit and use `import signal` `signal.signal(signal.SIGTERM, lambda signum, stack_frame: sys.exit(0))` to reduce global namespace variables

radfish:
Zeronet 0.6.5 just hangs on receiving SIGTERM from systemd.
Maybe sys.exit(0) from signal handler is not good? Perhaps something needs to be stopped from the main loop? I'll look into it if I find the time.

blurHY:
`sys.exit` would only raise a SystemExit exception. Non-daemon threads won't exit

PR #1934, opened 2019-03-24
Update SidebarPlugin.py for site having a lot of optional file

tangdou1:


DaniellMesquita:
How it works?

tangdou1:
> 
> 
> How it works?

When optional file is over 1000MB, showing unit as GB.

PR #1943, opened 2019-03-30
requirements: pyelliptic any version, websocket_client

radfish:
`pyelliptic` 1.5.8 works fine for me. Any particular reason it was frozen at 1.5.6?

`websocket` is likely not the package you are looking for -- the canonical websocket package is  `websocket_client`. `websocket_client` installs `/usr/lib/python3.7/site-packages/websocket/`

Btw, this looks like stale code:
```
$ grep -C2 -rnI 'from lib import websocket'
src/main.py-338-
src/main.py-339-    def getWebsocket(self, site):
src/main.py:340:        from lib import websocket
src/main.py-341-        ws = websocket.create_connection("ws://%s:%s/Websocket?wrapper_key=%s" % (config.ui_ip, config.ui_port, site.settings["wrapper_key"]))
src/main.py-342-        return ws
```
Unrelated: I also added links to distribution packages to the README. Distro packages are always preferred when they exist. Also, added installation into virtualenv because it's preferred and does not require root.

PS. I've updated zeronet-git Arch package on AUR to the Python 3 branch (it's supposed to be bleeding edge).
PPS. If I have some time I will try to package zeronet itself with setuptools (setup.py). It should be installable into site-packages and runnable with sources owned by root (like any other python application).

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,15 +63,46 @@ Decentralized websites using Bitcoin crypto and the BitTorrent network - https:/
 
 ## How to join
 
-* `sudo apt-get update`
-* `sudo apt-get install python3-pip`
-* `wget https://github.com/HelloZeroNet/ZeroNet/archive/py3.tar.gz`
-* `tar xvpfz py3.tar.gz`
-* `cd ZeroNet-py3`
-* Install required Python modules `sudo python3 -m pip install -r requirements.txt`
-* Start with `python3 zeronet.py`
-* Open http://127.0.0.1:43110/ in your browser
+### Install from package for your distribution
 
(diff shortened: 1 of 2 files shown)


HelloZeroNet:
The pyelliptic package only used for ECIES encrypt API commands (eg for ZeroMail), so it may look like that the client works fine, but the API calls will fail (like the travis-ci failed on this PR), because they removed the ECIES support.

radfish:
Ok. Got it. Removed the version unfreeze commit from this PR.

What's your plan for ECIES? Should we factor that code out of pyelliptic 1.5.6 into a new Python package and depend on that?

radfish:
Or https://pypi.org/project/eciespy/ ?

HelloZeroNet:
The cryptography module would be our best option, but after a quick check I see no out of the box ECIES support in it.
The openssl1.1 compatible bitmessage module would be a good alternative.

PR #1950, opened 2019-04-02
Include mime.types file and use it for WebUI Content-Types

ValdikSS:
Python 2.7 on Windows loads known MIME types from registry, which is
not always up-to-date. For example, it causes issue with webp images:
http://127.0.0.1:43110/Talk.ZeroNetwork.bit/?Topic:1554160326_131Seivv6aH8vnohPmYPcPwPfZ1a5ubMMs

The mime.types file has been merged from Fedora's /etc/mime.types and
https://s-randomfiles.s3.amazonaws.com/mime/allMimeTypes.txt

### DIFF
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -48,6 +48,15 @@ class UiRequest(object):
         self.user = None
         self.script_nonce = None  # Nonce for script tags in wrapper html
 
+        # Use bundled MIME types file
+        if not mimetypes.inited:
+            mimetypes.init(
+                [
+                    os.path.dirname(os.path.abspath(__file__)) + "/../lib/mime_types/mime.types",
+                    "../lib/mime_types/mime.types",
+                ]
+            )
+
     def learnHost(self, host):
         self.server.allowed_hosts.add(host)
         self.server.log.info("Added %s as allowed host" % host)
--- a/src/lib/mime_types/mime.types
+++ b/src/lib/mime_types/mime.types
@@ -1,0 +1,1497 @@
+application/A2L			a2l
+application/acad			dwg
+application/AML			aml
+application/andrew-inset			ez
+application/applixware			aw
+application/arj			arj
+application/ATF			atf
+application/ATFX			atfx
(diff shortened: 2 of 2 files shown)


HelloZeroNet:
I think it would be better more predictable not use mimetypes module, but add a simple module that parse handle mimetypes queries and parse the data file.

ValdikSS:
@HelloZeroNet, why? mimetypes is a small module, it's included in standard Python library and already bundled&used in ZeroNet.

imachug:
@HelloZeroNet Any progress on this?

HelloZeroNet:
I still think it would be easier/faster to include a small list of known popular filetypes without using the mimetypes module. For example: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types

filips123:
@HelloZeroNet What if program first checks some common MIME types from your list. If none of them matches, the program also checks `mimetypes` module (because it contains more types). If it still doesn't match, it uses `application/octet-stream`.

HelloZeroNet:
It's also possible, but I think its better if it's does not depends on system settings, but works the same way on for every client. So if there is any problem related to content-type, then we can find the problem and reproduce it easier.

ValdikSS:
@HelloZeroNet Are you aware that mimetypes module just loads my list and parses it, without using system mime types? Are you still want to replace it with a custom one? I don't understand the reason.

HelloZeroNet:
I usually try to avoid using packages if I can do the same with a few lines of code to avoid unwanted side effects. 
Eg for in this case `mimetypes.guess_type("any.html.xz")[0]` returns text/html, but probably we want to serve it gzip file as browsers won't be able to display the xz file content as a html
Other problem could be if any other module uses the mimetypes before our mimtes.init called, then it won't be inited from our file.

ValdikSS:
Unfortunately I don't have much spare time recently.
I propose the following: use this modification for Python 2 version, and for Python 3 version I (or somebody else) will make a proper mime types implementation.

PR #1953, opened 2019-04-03
ipv6 compatibility

tangdou1:


DaniellMesquita:
Compatibility with what?
Not clear to understand, as IPv6 compati ility (and its featres) is already a thing

PR #1954, opened 2019-04-03
Update SidebarPlugin.py for better showing

tangdou1:


HelloZeroNet:
We could avoid the code duplication if the size formatter function would be a separate function

tangdou1:
> 
> 
> We could avoid the code duplication if the size formatter function would be a separate function

Yes, that will be a good idea. However, I donot know how to implement it explicitly.

krzotr:
> > We could avoid the code duplication if the size formatter function would be a separate function
> 
> Yes, that will be a good idea. However, I donot know how to implement it explicitly.

I have removed code duplication.

Created function `helper.prettySize` and added tests

PR #1959, opened 2019-04-06
Change default value to 10MB

tangdou1:
Big File is an optional file which is bigger than 10MB, so the default value should be at least 10MB, which means the effect for setting this value to any number that is lower than 10 is the same as the one with 10MB.
When I open this toggle `Download and help distribute all files` in the side bar, all the small optional files (lower than 10MB) will be downloaded no matter the value of `Auto download big file size limit` is 1MB or 10MB. It will give a misunderstanding that why the optional file that is even bigger than 1MB  is also downloaded, when the `..file size limit` value is just 1MB. (Because the value is only valid for big file which is bigger than 10MB). For eliminating misconception, so the default value should be 10MB.


imachug:
@HelloZeroNet Can we *please* merge this PR that adds a single character?

PR #1962, opened 2019-04-06
Sidebar: use geoip db from system if exists

radfish:
Avoid downloading the GeoIP DB when it is already provided by the system.
The system one is kept up to date by the system package manager, while the downloaded one goes stale.

For example, on Arch Linux geoip2-database package already provides the Db.
I will add the package dependency to Arch package once this is merged.


radfish:
Thanks for merging. Added geoip2-database dep to Arch Linux zeronet-git pkg.

PR #1963, opened 2019-04-07
[setuptools packaging] remove some  dead code related to openssl, pyelliptic

radfish:

*  zeronet: no openssl.closeLibrary in pyelliptic
    Not in pyelliptic  (neither in 1.6.7 nor 2.0.1)

grep closeLibrary shows nothing: do you know what else to check?

 * zeronet: remove ref to opensslVerify
    This module is no longer used. Does not exist in lib/

### DIFF
--- a/zeronet.py
+++ b/zeronet.py
@@ -24,12 +24,6 @@ def main():
         if main.update_after_shutdown:  # Updater
             import gc
             import update
-            # Try cleanup openssl
-            try:
-                if "lib.pyelliptic" in sys.modules:
-                    sys.modules["lib.pyelliptic"].openssl.closeLibrary()
-            except Exception as err:
-                print("Error closing pyelliptic lib", err)
 
             # Close lock file
             sys.modules["main"].lock.close()
--- a/zeronet.py
+++ b/zeronet.py
@@ -26,11 +26,6 @@ def main():
             import update
             # Try cleanup openssl
             try:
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
TODO: To be able to update the dlls we have to force-close them somehow to make the files writeable

radfish:
Make the updater a separate process that is forked off?
Unless you specifically want some kind of "live" updating.

HelloZeroNet:
Not sure if that would work on every platform.

radfish:
On Sun, Apr 07, 2019 at 10:23:27AM -0700, ZeroNet wrote:
> Not sure if that would work on every platform.

Hm, dunno. Any obstacles in particular you're thinking of?

Isn't the only support needed is to launch processes? ZeroNet already
launches a process for ssl certificate generation. I mean, on update,
fork an updater process (`run_shell_cmd(python update,py <parent_pid>)`)
and shutdown, updater waits for `parent_pid` to exit, then overwrites
whatever files (within the ZeroNet distribution), then forks of another
process to relaunch ZeroNet (`run_shell_cmd(python zeronet.py)`) and
exits. Not a live update, of course, but any other problem with this?

But.. I shouldn't speak about this anyway, because I don't think I would
be the one to contribute to the updater, because I don't use it (I rely
on distribution packager instead).

PR #1964, opened 2019-04-07
[setuptools packaging] factor path to testdata, log error

radfish:

* test: refer to test data path via variable (refactoring only, no change to functionality)

* UiRequest: log error respones
    
    Useful during development. In some cases, things brake, but no error
    shows up anywhere. Add logging for that case.

### DIFF
--- a/src/Test/TestSite.py
+++ b/src/Test/TestSite.py
@@ -4,16 +4,17 @@ import os
 import pytest
 from Site import SiteManager
 
+TEST_DATA_PATH = "src/Test/testdata"
 
 @pytest.mark.usefixtures("resetSettings")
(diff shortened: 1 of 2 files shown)


radfish:
Removed the commit regarding adding Db free functions to Db package. Moving that commit to a different PR that actually depends on it.

radfish:
@HelloZeroNet good to go or any changes needed?

HelloZeroNet:
Looking good, but I think the formatting function should not contain logging (formatError)

radfish:
Agreed. Logging doesn't belong there -- should find a better place for it up the stack.
Removed that commit and rebased.

PR #1969, opened 2019-04-07
[setuptools packaging] access modules via imports

radfish:
This PR is part of preparatory patches for setuptools  (setup.py) packaging (which is ready, but I'm trying to minimize the  size of that diff by splitting all independent changes into separate  PRs).

For setuptools packaging we will need to change `import`s throughout the codebase. This PR replaces usages of `sys.modules["module_name"]` with `import module_name`, which is better (one reason is that it enforces encapsulation) and it simplifies the setuptools diff to contain only modifcations to `import` statements.

Note: it would be best to have all `import`s at the top of the files, but that is not currently possible due to some modules having side effects from importing. This is too large a patch to fix, so I'm leaving some imports on-demand.

This PR is purely refactoring. No change in functionality at all.

*  imports: refer to modules via import objs
    
    Replace the access via sys.modules (which is prone to violate
    encapsulation).
    
    For main, import strictly on demand (not at the top of each file),
    because importing main has side-effects (unfortunately), and importing
    it at time of loading the respective module breaks things (at the very
    least, it breaks tests, due to argparser arguments getting created
    twice -- once by test script and once by main).
    
    Unfortunately, not all usage of sys.modules can be replaced with imports
    at this point in time. Namely, TorManager.py and Site.py are not only
    accessing main via sys.modules but also check whether main is loaded or
    not. To fix this, a significant change to main is required -- to factor
    it such that the owner of the singletons can be imported without
    sideeffects.

 *   Db: remove the imports from the pkg __init__.py
 *    Site: remove imports from pkg __init__.py
    
 Have the importer import the modules and classes explicitly, to keep things simple.

Thank you for reviewing. Let's get this setuptools packaging merged.

### DIFF
--- a/plugins/Chart/ChartCollector.py
+++ b/plugins/Chart/ChartCollector.py
@@ -25,7 +25,8 @@ class ChartCollector(object):
 
     def getCollectors(self):
         collectors = {}
-        file_server = sys.modules["main"].file_server
+        import main
+        file_server = main.file_server
         sites = file_server.sites
         if not sites:
             return collectors
@@ -100,7 +101,8 @@ class ChartCollector(object):
         return site_collectors
 
     def getUniquePeers(self):
-        sites = sys.modules["main"].file_server.sites
+        import main
+        sites = main.file_server.sites
         return set(itertools.chain.from_iterable(
             [site.peers.keys() for site in sites.values()]
         ))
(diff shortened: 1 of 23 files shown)


radfish:
rebased (and fixed conflict)

radfish:
rebased (and fixed conflict) again.
I will keep rebasing, but how is the review coming along? Thanks.

HelloZeroNet:
Thanks for the PR. I was thinking about maybe it would be better to get rid of __import__.py magic, so "from Db import Db" would mean the module and not the class. So we could use "from Db import Db; Db.opened_dbs"
It also requires to modify some plugins import, but it would me more consistent.
What do you think?

radfish:
Cool. Sounds better. Let me try this out.

radfish:
Updated with the removal of imports in __init__.py

PR #1972, opened 2019-04-08
Use Local App Data when running from Program Files

filips123:
The program now checks if it is running from `Program Files` and then set `start_dir` to `%LOCALAPPDATA\ZeroNet`.

There are still some problems with this. See https://github.com/HelloZeroNet/ZeroHello/issues/116#issuecomment-480868018.

Should I also submit this to `py3` branch?

### DIFF
--- a/src/Config.py
+++ b/src/Config.py
@@ -66,10 +66,8 @@ class Config(object):
         else:
             start_dir = "."
 
-        try:
+        if not os.path.isdir(start_dir):
             os.makedirs(start_dir)
-        except OSError:
-            pass
  
         return start_dir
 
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
Not necessary, the changes will be merged when the py3 considered stable.

It would be better instead of "try:" use "os.path.isdir(start_dir)", so if there is any error then it can be displayed

filips123:
Ok, fixed!

What about that errors in https://github.com/HelloZeroNet/ZeroHello/issues/116#issuecomment-480868018?

HelloZeroNet:
The meek client is handled and started by tor.exe, we dont have any control over it.

filips123:
So what's then the issue? Maybe Tor needs writable directory? ZeroNet process is also not stopped.

HelloZeroNet:
Yes, tor also needs a writeable directory for the authentication file

filips123:
Ok, I installed to `%LOCALAPPDATA%`.

Can you make installation program that would correctly install and set up all directories?

PR #1973, opened 2019-04-10
[setuptools packaging] Sidebar: rename media-globe/ to media_globe/

radfish:

Rename folder so that it can be an importable package. This is not frivolous; this is necessary for loading data files as resources (coming up in a separate PR).


radfish:
Rebased, just in case.
(This PR is a prerequisite for next PR for setuptools.)

Issue #917, opened 2017-04-25
PyElliptic is officially deprecated. Consider using another crypto library.


MuxZeroNet:
Yann [announced](https://github.com/yann2192/pyelliptic/issues/50#issuecomment-297018932) that PyElliptic is officially **deprecated**.

Please consider using another wrapper or crypto library, such as [pyOpenSSL](https://pypi.python.org/pypi/pyOpenSSL), [cryptography](https://cryptography.io/en/latest/), [PyNaCl](https://pynacl.readthedocs.io/) or [pySodium](https://github.com/stef/pysodium). Here is a crypto library choice discussion among Ethereum devlopers: ethereum/pydevp2p#58

[pyOpenSSL](https://pypi.python.org/pypi/pyOpenSSL) - a OpenSSL wrapper library

[cryptography](https://cryptography.io/en/latest/) - the famous, actively maintained Python cryptography library. If used, it needs to be included in `requirements.txt`

[PyNaCl](https://pynacl.readthedocs.io/) and [pySodium](https://github.com/stef/pysodium) are both wrappers for libsodium. PyNaCl is mentioned and recommended in cryptography's documentation. The libsodium provides Curve25519 for public key encryption, Salsa20 + Poly1305 for secret key encryption, and Ed25519 for digital signatures.

HelloZeroNet:
Unfortunately Nacl is not compatible with the crypto we use, pyopenssl does not have functions we need (eg aes encryption). Cryptography probably could work, but has binary dependencies and requires compilation to every platform/openssl version, which could reduce the portability.
Pyelliptic still has some support by the bitmessage project

ofek:
[https://github.com/pyca/cryptography](https://github.com/pyca/cryptography)'s `pip install cryptography` only requires compilation on linux. Binary distributions are available for Windows and macOS.

MuxZeroNet:
On the Privacy mailing list, Peter Surda wrote:

> Hello and thank you for the info.
> 
> The PyElliptic code that's in PyBitmessage is a modified version of PyElliptic (i.e. a fork). I'm not aware of any issues with it, those that existed were fixed (e.g. OpenSSL 1.1 support, library locator, using a different hash algorithm). I tested it on a broad range of operating systems.
> 
> I'll keep it for the time being as it is.
> 
> Peter Surda  
> Bitmessage core developer

[...]

PR #1974, opened 2019-04-10
Get rid of pyelliptic

purplesyringa:
So I got tired of this dependency we are all talking about. This PR replaces `pyelliptic` with other libraries.

1. For AES, I use `pycryptodome` (see `plugins/CryptMessage/CryptMessagePlugin.py`). It's simple to use and should be simple to maintain:
```python
cipher = AES.new(key, AES.MODE_CBC, iv)
encrypted = cipher.encrypt(pad(text, AES.block_size))
```

2. For ECIES, I wrote my own library. I've tested it on your tests, and compared the binary data with some other data, and I'm sure it works correctly. The format is compatible with pyelliptic's format (though I think that this format is weird and seldom used). For key derivation (ECDH), I used some code from `pyelliptic` and fixed it to be compatible with modern OpenSSL (according to the docs). For the AES stage, I use `pycryptodome`. For MAC, I use the standard `hmac` module (preinstalled).

### DIFF
--- a/plugins/CryptMessage/CryptMessage.py
+++ b/plugins/CryptMessage/CryptMessage.py
@@ -6,13 +6,11 @@ import lib.pybitcointools as btctools
 ecc_cache = {}
 
 
-def eciesEncrypt(data, pubkey, ephemcurve=None, ciphername='aes-256-cbc'):
+def eciesEncrypt(data, pubkey, ciphername='aes-256-cbc'):
     import pyelliptic
     pubkey_openssl = toOpensslPublickey(base64.b64decode(pubkey))
     curve, pubkey_x, pubkey_y, i = pyelliptic.ECC._decode_pubkey(pubkey_openssl)
-    if ephemcurve is None:
-        ephemcurve = curve
-    ephem = pyelliptic.ECC(curve=ephemcurve)
+    ephem = pyelliptic.ECC(curve=curve)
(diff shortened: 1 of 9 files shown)


radfish:
Very nice!
There was an effort to not vendor dependencies, but this brings back vendoring again. What do you think about publishing your ECIES library to pypi?

Btw, since your library is based on pyelliptic's source code with fixes for OpenSSL v1.1, roughly what are the differences from the pyelliptic fork from PyBittmessage (https://github.com/radfish/pyelliptic/), which is also based on pyelliptic's source code with fixes for OpenSSL v1.1?

Also, mentioning #917

imachug:
My library is based on *one* 50-line file from pyelliptic, not the whole package. As for publishing to pypi, ok, I'll publish the package soon.

HelloZeroNet:
Thanks for the PR. Do you think if we could use the AES implementation from openssl? I just tested cryptodome and it's looks pretty heavy (25MB in 695files) and on my SBC it takes 10s to execute "from Cryptodome.Cipher import AES" which I like to avoid.
Maybe it's easier to use radfish's fork of pyelliptic that also include that function.

imachug:
However, it makes sense not to use a deprecated package, even if it was forked. So I would vote against using a fork, it would be better to make our own small lib using OpenSSL docs.

radfish:
@imachug how does your library for ECIES compare to eciespy package? (see recent comment in #917)

imachug:
eciespy forces the usage of sha256 which would break backward compatibility.

PR #1984, opened 2019-04-14
Add ECDSA actions to CryptMessage

imachug:


krixano:
This is very useful for an Id Provider. I'm already using it for this, but can't publish until this is merged.

HelloZeroNet:
Is hashing the data before sign/verification necessary?

imachug:
It is. Signing/verifying is an extensive operation, a really extensive one. So it makes sense to make data smaller. That's how everyone does it.

HelloZeroNet:
The CryptBitcoin.sign will hash it anyway (https://github.com/HelloZeroNet/ZeroNet/blob/py3/src/lib/pybitcointools/main.py#L520), so hashing the message is not necessary.

imachug:
Okay, will fix

PR #1987, opened 2019-04-16
Fix ECDSA on CryptMessage

imachug:


imachug:
@HelloZeroNet I've made another fix, please check it. Also, you should probably squash this PR.

plugins/CryptMessage/CryptMessagePlugin.py
```diff
@@ -117,12 +117,12 @@ def actionEcdsaSign(self, to, data, privatekey=None):
         if privatekey is None:  # Sign using user's privatekey
             privatekey = self.user.getAuthPrivatekey(self.site.address)
 
-        self.response(to, CryptBitcoin.sign(data, privatekey))
+        self.response(to, CryptBitcoin.sign(data.encode("utf8"), privatekey))
```
imachug: @HelloZeroNet I'm really sorry for all this sh\*t, but I have no idea how to make GitHub resolve the conflicts correctly. There should be *no* `.encode` call here. I think that when I was resolving conflicts, I removed this, but GitHub thinks the other way. Please fix this bug.

PR #1989, opened 2019-04-17
Ui, Translate: fix bytearray format string for Py 3.4

radfish:
Py 3.4 does not support bytearray format strings for % operator: b"%s" % s

Fixes on Python 3.4:
```
UiWSGIHandler error: TypeError: unsupported operand type(s) for %: 'bytes' and 'bytes' in UiServer.py line 41 > pywsgi.py line 924 > pywsgi.py line 908 > UiRequest.py line 302
```

### DIFF
--- a/plugins/TranslateSite/TranslateSitePlugin.py
+++ b/plugins/TranslateSite/TranslateSitePlugin.py
@@ -57,7 +57,7 @@ class UiRequestPlugin(object):
         if not lang_file_exist or inner_path not in content_json.get("translate", []):
             for part in file_generator:
                 if inner_path.endswith(".html"):
-                    yield part.replace(b"lang={lang}", b"lang=%s" % translate.lang.encode("utf8"))  # lang get parameter to .js file to avoid cache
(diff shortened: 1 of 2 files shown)


imachug:
Shouldn't we use .format? % is deprecated.

imachug:
Hey, @HelloZeroNet ! We shouldn't just fix legacy all the time, we should use `.format` instead of `%`. Moreover, `format` is supported on byte strings as well!

radfish:
@imachug Sorry, I didn't know % was deprecated. Let just convert all of them in a separate PR.

Funny though:
> Since str.format() is quite new, a lot of Python code still uses the % operator. However, because this old style of formatting will eventually be removed from the language, str.format() should generally be used.

Removed from the language? So.... brace for Py 4.0 incoming. Please have mercy not again. Lol.

HelloZeroNet:
As far as I know its not deprecated. If they would planning to remove it on short term probably they would not added % formatting support of bytes in Python 3.6

krixano:
You should switch to using f"" (format string literals) anyways because it's faster than both .format and % operator.

PR #1991, opened 2019-04-18
[setuptools packaging] load auxiliary files via resources

radfish:

This is THE LAST preparatory PR.

Load auxiliary files via Python importlib resource API. When the setuptools package is installed to `/usr/lib/python/site-packages` we should access files via the resource API to keep the path logic simple and avoid issues with paths.

The importlib resource API is builtin into Python >=3.7 as importlib.resources and for older pythons there is an official backport package importlib_resources (see Resources.py in this PR).

The importlib API requires the resource files to be part of packages, hence added the __init__.py files.

 *   Ui: load UI media and translations from resource
    
 Merging *.css/*.js into all.css/all.js in debug mode still reads the input files as files from the file system, not as a resources.
    
 * UiPassword: load login.html and icon as resource
 * Crypt: load openssl conf from resource
 * lib: pybitcointools: load worldlist as resource

Manually tested on Py 3.7 and Py 3.4.


HelloZeroNet:
Are you sire it's necessary to turn every directory to a module?
The documentation says you can include data files using `package_data={'mypkg': ['data/*.dat']}`
https://docs.python.org/3.4/distutils/setupscript.html#installing-package-data

radfish:
I am using package_data -- that has to be used.
The purpose of this PR is about how the files are accessed when the package is installed to /usr/lib. It's best to access them via the import system as resources (clear and simple) as opposed to constructing paths somehow derived from __file__ or some hack like that.

radfish:
@HelloZeroNet I rebased and retested to keep this PR fresh on top of latest py3 branch.

CI for py 3.7 failed in CryptMessage for some reason -- I cannot reproduce on localhost. Do you know what is the problem with the test?

Also, any changes you want me to make, or is it good to go? Thanks.

HelloZeroNet:
I have restarted the test and it passes now (I seen this error before and tried to find the problem, but looks like it's still not fixed)

Actually I'm still not sure why do we need a wrapper over reading files and turn every directory to module. The resources package also deliver the path from the .py file location: https://gitlab.com/python-devs/importlib_resources/blob/master/importlib_resources/_py3.py#L79-83

As far as I know the only benefit would be that it could be packed up as a zipped file, but I'm not sure if it's worth the added layer and the added 30+ module to the namespace. But if this is really a required modification, then I will do some testing on my SBC and if there is no significant overhead, then we can change the file access to this way.

radfish:
Ok, I certainly don't want to add anything that's unnecessary. I don't have a full understanding of resources, so maybe I am missing some other way. Here's the problem that can't get around:

    mkdir -p pkg/media
    touch pkg/media/file
    touch pkg/__init__.py
    python
    >>> import importlib.resources as r
    >>> import pkg.media as m
    >>> r.read_text(m, 'file')
    FileNotFoundError: Package has no location <module 'pkg.media' (namespace)>
but once you make media a package, it works:

     touch pkg/media/__init__.py
     >>> r.read_text(m, 'file')
     ''
I'm actually surprised that `import pkg.media` did not fail even though media was not a package. But in any case resource loading fails cause I'm asking it to load from something that is not a package. Are you saying that this should work under some special circumstance?

Btw, I also read about running from zip file, and I agree that it's not top priority, but it would be nice to be compatible, if this support comes for free as a sideeffect of using resources. But, I only care about being able to load the resource after package is installed -- can we achieve that with importlib.resources but without packages?

If the importlib method causes performance issue, then we can do it without importlib at all, with `os.path.join(os.path.dirname(zeronet.__file__), 'path/to/resource')`, but going via the import system seems cleaner.

Btw, I also run zeronet on ARM, so can test too.

radfish:
I roughly timed the startup on ARM with and without the patch: can't see a difference -- around 15s+/-3s. Not a precise test though.

In any case, would you rather do `os.path.join(os.path.dirname(zeronet.__file__), 'path/to/resource')`? If so, I'll close this PR and submit another one.

radfish:
Hello @HelloZeroNet what's your verdict on this? Do you want to implement alternative via `__file__` instead (see prev comment)?

HelloZeroNet:
I didn't had time to test it, but I will do tomorrow.

radfish:
Rebased. @HelloZeroNet

radfish:
@HelloZeroNet  rebased and re-tested

radfish:
@HelloZeroNet any chance at merging this so that we can move on and finish up the packaging at last?

radfish:
@HelloZeroNet rebased and re-tested again. Anything I can do to make the merge happen? I just want to close this chapter on setuptools packaging finally.

HelloZeroNet:
Thanks again for the effort.
I started to review/rearrange multiple times and really wanted to like the importlib.resources module, but I can't. 
I don't like the idea to have a wrapper over file reading, unnecessary `__import__.py` files and `__pycache__` directories in every subdirectory of the media and  `*(os.path.join(res_pkg.split('.') + [res_file]))` like hacks.
We could improve the resource files path handling, but I would prefer more simple solution, like:

```py
import os

MEDIA_URL = "uimedia/plugins/uiconfig"
MEDIA_PATH = os.path.dirname(__file__) + "/media"

def getMediaPath(url, prefix_url, prefix_path):
    url = url.lstrip("/")
    if url.startswith(prefix_url):
        return url.replace(prefix_url, prefix_path, 1)
    else:
        raise Exception("Url not in prefix url")


print(getMediaPath("/uimedia/plugins/uiconfig/js/all.js", MEDIA_URL, MEDIA_PATH))
```

radfish:
Ok. I'll put together a PR based on `__file__` approach.

PR #1993, opened 2019-04-19
Support web notifications

imachug:


HelloZeroNet:
Looking good, but I'm not sure if PushNotification permission is necessary, as the notifications only going to work when the browser tab is opened, because lack of the serviceworker service

imachug:
I am sure this is required. If you allow notifications for 127.0.0.1, then all the zites would be able to send notifications. It's better to have explict confirmation. Compare it to Clearnet: you have to allow noifications for each site.

HelloZeroNet:
I think on clearnet the notification confirmation is required, because if you accept it, then the sites also able to send you notification if you are not on the page. Bot for us it's not the case, because there are no service workers.

imachug:
Okay, done.

PR #2006, opened 2019-05-01
Add privToPub and pubToAddr commands

imachug:
I think that doing this stuff in browser is a bad idea -- the site size increases a lot, however, we have these functions in Python already -- so we can use ZeroFrame to access them.


HelloZeroNet:
Looking good, but please rename the functions to actionEccPrivToPub and actionEccPubToAddr to make it more unique

imachug:
Sorry, no idea why I missed your comment. Fixed now

PR #2015, opened 2019-05-15
Travis CI: Add Python 3.6 in allow_failures

cclauss:
Retry #1392 one year later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: python
 python:
   - 2.7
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
Thanks for the PR, actually we have a separate bracnh for py3 version (will become default soon) that has test against 3.4-3.7: https://github.com/HelloZeroNet/ZeroNet/blob/py3/.travis.yml

PR #2016, opened 2019-05-15
.travis.yml: The 'sudo' tag is now deprecated in Travis CI

cclauss:
[Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

"_If you currently specify __sudo:__ in your __.travis.yml__, we recommend removing that configuration_"

Also, removed Python 3.4 because has already reached its end of life:
https://devguide.python.org/devcycle/#end-of-life-branches

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 python:
+  - : 3.4
   - : 3.5
   - : 3.6
   - : 3.7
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
Py 3.4 compatibility is required because that the last version that supports windows xp, which is still popular in some countires.

HelloZeroNet:
There is some logging error appeared in pytest result, but after some testing it turned out it's because of pytest change (probably: https://github.com/pytest-dev/pytest/issues/4942)

I will see if we can fix it.

PR #2026, opened 2019-05-19
ECMAScript modules

imachug:
ECMAScript modules are rather new, but they are supported by Chrome and Firefox already, so I think it's important to support them as well.

This PR makes several changes:

1. All filenames ending with `.js` are forced to have `text/javascript` mimetype (fixes Chrome and Firefox)
2. JavaScript files are allowed to be loaded without Referer header (fixes Chrome)


filips123:
@imachug Are you sure it must be `text/javascript` mimetype? Because some people say it must be `text/javascript` and some other `application/javascript`.

PR #2029, opened 2019-05-23
Fix calling start.py to reopen browser

imachug:
Calling start.py/zeronet.py twice in a row wouldn't work, because ZeroNet would try to logrotate before locking lock.pid.


HelloZeroNet:
Probably it would be better solution to move the initLogging() function call after the lock checking part: https://github.com/HelloZeroNet/ZeroNet/blob/master/src/main.py#L64

imachug:
Ok, I wasn't sure whether the functions between these lines access logging somehow, will move initLogging then

imachug:
There we go.

Issue #1891, opened 2019-02-06
Function to clone hub easily from a merger site


slrslr:
Merger sites are hard to use if every new n00b user has to **[learn how to manually clone the hub site](http://127.0.0.1:43110/17gCoQDCAAoQkzKRiajFNUAhnVD2gSNmuC/?:13kifcv8J8noDEXdztXY4qkPPPKLK2PiAe:9e148d9a-4685-4224-bf2f-a4ca448b7bd2)**.

It may be handy if there is ZeroNet (API?) function allowing to clone a hub (creating new one) using button on a merger site.
The parameters of such a function would be:
- setting the hub site title
- site description
- max_size parameter per users data.json in data/users/content.json
- result of the query would be address of the new hub that user is invited to copy and save + can click to open the site in new tab.

[1DocsYf2tZVVMEMJFHiDsppmFicZCWkVv1/site_development/zeroframe_api_reference/](/1DocsYf2tZVVMEMJFHiDsppmFicZCWkVv1/site_development/zeroframe_api_reference/)

imachug:
Actually that's exactly what happens when you create a new repo on Git Center. You press `Clone`, wait a few seconds and you have a page where you can set repo title or description. What I do is just call `siteClone` method and put auto-redirect to `index.html`.

rllola:
The redirect after doing a `siteClone` is a bit annoying. @imachug Do you think it should be removed ?

imachug:
I'm afraid that would break backward compatibility, but we could add it as an option, e.g. `siteClone(address, redirect=False)`. In this case, siteClone should return the cloned site address.

rllola:
> I'm afraid that would break backward compatibility, but we could add it as an option, e.g. `siteClone(address, redirect=False)`. In this case, siteClone should return the cloned site address.

Ok seems like a good solution to me. :+1:

PR #2032, opened 2019-05-25
Added optional redirect value for siteClone

rllola:
Added an argument which allow the developer to not redirect to the newly cloned site.

See #1891;


imachug:
This is not enough. In case the redirect does *not* happen, we should also return the cloned site address -- otherwise, there's no way to get it.

rllola:
Yeah good point. I am on adding this.

rllola:
It now returns the address of the newly cloned site.

@HelloZeroNet ready for review.

HelloZeroNet:
Thanks, looking great!

imachug:
Cool, thanks for merging.

PR #2033, opened 2019-05-25
Better error message if private key not stored when assigning

### LINKED ISSUE #2022: Failed to sign via websocket api really caused by a not found privatekey

  * ZeroNet version: Version: 0.6.5 r3862
  * Operating system: Linux
  * Web browser: Firefox
  * Tor status: not available
  * Opened port: no
  * Special configuration: None

I have been using the websocket api to sign a merger site content.json and it keeps failing. However I determined that ZeroNet wasnt able to find the key. Instead of telling me that the private key was not found it says that it fails to sign.

Error message : 
```
[17:08:10] Site:15t2dF..mnE4 Site sign failed: data/users/content.json: SignError: Private key invalid! Valid signers: ['*************'], Private key address: ************ in UiWebsocket.py line 459 > ContentManager.py line 711
```

### Proposition

Check if the privatekey is different from `None` and show a more instructive error message.

NOTE: Obviously not a bug per se but could get a bit confusing. I would be happy to come with a PR for this.

rllola:
Return a more instructive message in case the privatekey is not found when attempting to sign.

**Current behavior**: It will fetch the the user private key and not the site private key. It will ultimately failed because we are trying to modify a merger-site file.

**Fix** : If failed to get the site private key, an instructive message will be shown to say that the key is not stored. It will no more try to fetch the user private key as a fallback to avoid incoherent error message.

Fix #2022

### DIFF
--- a/src/Ui/UiWebsocket.py
+++ b/src/Ui/UiWebsocket.py
@@ -461,7 +461,7 @@ class UiWebsocket(object):
 
         if privatekey == "stored":  # Get privatekey from sites.json
             privatekey = self.user.getSiteData(self.site.address).get("privatekey")
-            self.cmd("notification", ["error", _["Content signing failed"] + "<br><small>Private key for found in sites.json </small>"])
+            self.cmd("notification", ["error", _["Content signing failed"] + "<br><small>Private key not found in sites.json </small>"])
             self.response(to, {"error": "Site sign failed: Private key not stored."})
             self.log.error("Site sign failed: %s: Private key not stored in sites.json" % inner_path)
             return
--- a/src/Ui/UiWebsocket.py
+++ b/src/Ui/UiWebsocket.py
@@ -461,6 +461,10 @@ class UiWebsocket(object):
 
         if privatekey == "stored":  # Get privatekey from sites.json
             privatekey = self.user.getSiteData(self.site.address).get("privatekey")
+            self.cmd("notification", ["error", _["Content signing failed"] + "<br><small>Private key for found in sites.json </small>"])
+            self.response(to, {"error": "Site sign failed: Private key not stored."})
+            self.log.error("Site sign failed: %s: Private key not stored in sites.json" % inner_path)
+            return
         if not privatekey:  # Get privatekey from users.json auth_address
             privatekey = self.user.getAuthPrivatekey(self.site.address)
 


imachug:
1. Shouldn't the error handler be under `if not privatekey`?
2. It isn't that easy. Both `siteSign("content.json", "stored")` and `siteSign("data/users/authaddress/content.json", "stored")` should work, so showing the notification from UiWebsocket is a bad idea -- it should be the sidebar JS code that shows it.

rllola:
>     1. Shouldn't the error handler be under `if not privatekey`?

If I put it under it will fetch the private key of the _user_ and not the _site_ private key. The signing will then fails. That's what happened to me and couldn't figure out why because I was using the `stored` keyword but the _site_ private key wasn't stored and so return `None` so it then fetched my _user_ private key and attempt to sign the new `content.json` with it.

>    2. It isn't that easy. Both siteSign("content.json", "stored") and siteSign("data/users/authaddress/content.json", "stored") should work, so showing the notification from UiWebsocket is a bad idea -- it should be the sidebar JS code that shows it.

I asked myself the same question. But then we will be in the same case I describe above... So I figure that `siteSign("data/users/authaddress/content.json", "stored")` should not work and if you want to use the _user_ key don't use `stored`. And If we want to keep `siteSign("data/users/authaddress/content.json", "stored")` working, we would need to take a guess by looking at the `inner_path` and it feels hacky.

Another possibility would be to have 2 differents keyword : `storedSite` and `storedUser`. Just throwing ideas here.

imachug:
It looks like the sidebar code was heavily rewritten since the moment when signing user files was implemented. So, this issue isn't actually related. We can report the `privatekey="stored", not actually stored` case as an error then.

However, I've actually noticed a small bug in the new `Sign` button implementation. If you're signing a user file, you should first try your user privatekey and *then* site privatekey, not in the opposite order. @HelloZeroNet

HelloZeroNet:
it's looks good, but please fix the error message: `<small>Private key for found in sites.json </small>`


@imachug To do that probably it would be better and more clear to add a "Sign as current user" separate menu item to the sidebar to avoid the confusion.

imachug:
Not really. We've simplified the sign/verify buttons by merging them, and now you want to complicate it?

I think it is clear enough: I can't find any case when signing with user privatekey is worse than signing by site privatekey.

rllola:
> However, I've actually noticed a small bug in the new `Sign` button implementation. If you're signing a user file, you should first try your user privatekey and _then_ site privatekey, not in the opposite order. 

But then we have the same problem in the other order? I just think the idea of trying another key as a fallback is not a good practice.

rllola:
The only problem we might have is here : https://github.com/HelloZeroNet/ZeroNet/blob/py3/plugins/Sidebar/media/all.js#L1006-L1050 

If it has found the private key of the site it will try to sign it with it without regarding which `content.js` we are trying to sign. So it will fails if you are trying to sign as an user. However it will only affect only site owners. 

I would say we show 2 buttons for site owner only so they can choose exactly which key they want to sign their `content.json` with. It won't make it more complex but more explicit.

The lack of information in what ZeroNet was doing is exactly the issue we are trying to fix. I didn't expect it to try signing using my **user** private key when signing my **site** content.json and showing me this incoherent error message.

imachug:
> But then we have the same problem in the other order?

Not really, I can't make an example when signing with site key is better than signing with user key.

> I just think the idea of trying another key as a fallback is not a good practice.

This makes the UI universal, though -- you sign both your and site's files the same way.

---

@HelloZeroNet Can you stop merging everything please? We haven't finished the conversation, we've just broken something, discussed that it's broken and now you just merge it!

Would you also merge one of my PRs if I added malicious code with a `Fix typo` commit? I think you wouldn't -- then why did you merge this!?

HelloZeroNet:
@imachug You are right, the PR was broken. I just fixed it and made a change that will use the user privatekey to sign the content if possible:
https://github.com/HelloZeroNet/ZeroNet/commit/a2d29a453115fc6580b9f89f0b9ba9c8d3ded5a8
https://github.com/HelloZeroNet/ZeroNet/commit/7b41922c2da45848d1fa6ee0c545c560e3d688ff

imachug:
Yay, a lot better now. Thanks!

PR #2043, opened 2019-06-03
Show error message when db failed to rebuild

### LINKED ISSUE #1908: Database rebuild says successfull but log showed error
I am trying to rebuild the database and the UI message says that everything went well when in the log I have an error message.

screenshot from 2019-02-19 13-37-45
screenshot from 2019-02-19 13-38-13

rllola:
See issue #1908;

**Fix**
Return `False` when failed to rebuild and sidebar plugin verify the value in `actionDbReuild` and if it has failed to rebuild it will create a response with an error message.


HelloZeroNet:
Thanks

PR #2045, opened 2019-06-09
Always error rebuild database message fix

rllola:
Return True if it has been able to rebuild DB even if importing a file fails.


HelloZeroNet:
Thanks, I added a quick fix some days ago: https://github.com/HelloZeroNet/ZeroNet/commit/8cb629fb55e304ad200fac2a97835718062353f6#diff-0823b49a6bc2f350e7c6134f24e9440d

So it should work fine now

PR #2047, opened 2019-06-14
Try and catch block for dbRebuild

rllola:


HelloZeroNet:
You could add a test for error eg. for an invalid dbschema file (eg broken json), then except an error for dbrebuild() call.
Here is an example how to do that in pytest: https://github.com/HelloZeroNet/ZeroNet/blob/862e19a26390c28d92d548e17b4fff797661f34f/src/Test/TestNoparallel.py#L130

HelloZeroNet:
The CI has failed because:
```

./src/Site/SiteStorage.py:79:25: F821 undefined name 'logging'

                        logging.error(err)

                        ^

1     F821 undefined name 'logging'
```
Can you please fix it? (it's better to use `site.error` where its possible)

rllola:
Fixed the errors in the CI.

And I have added a test that will try to rebuild the db of the test site. I am not fully getting how the test works and how I can modify the dbschema.json file in the test. I tried to use `writeJson` but doesn't seem to work. I don't think creatings a test site as a fixture is the good approach but could it be ?

PR #2053, opened 2019-06-26
Support different cryptographies

purplesyringa:
This PR adds support for other cryptographies -- since now, not only Bitcoin is supported.

One implementation example is [Ark plugin](https://github.com/imachug/Plugin-Ark), which you can use immediately by checking out `py3-ark` branch of my fork (https://github.com/imachug/ZeroNet/tree/py3-ark). While Ark is similar to Bitcoin, it's a good example for multiple cryptographies support.

### Can I use both Bitcoin and Ark?

Yes.

### As a site owner, can I decide what cryptography to use?

Yes.

### As a ZeroMail user, can I switch to Ark?

No. When ZeroMail adds other cryptography support, the answer will change to "Yes".

### What cryptographies can possibly be supported?

All that computers can use.

## Tutorial for site owners

<details><summary>Click to expand</summary>

Since now, sites can use different cryptographies for signatures. When you create a site via ZeroHello or clone someone's site, you get the following question:

image

(of course, the shown cryptographies depend on what plugins you have)

Whatever cryptography you choose, you get an address. For example, APXEPGdqwc5MbasL4SZBHdGYRxcAPjAc6A is an example address for Ark that I got while cloning ZeroTalk.

If you're making a site without user data, it's ready. You can do everything you usually do with zites. When you give the link to the zite to people who have the required cryptography installed, they'll be able to use the zite.

If you're making a site with user data, and you want the users to be able to use Ark, you'll have to use the specific certificate provider (e.g. ArkID -- a zite I've made). Of course, Bitcoin-address zites can support Ark-address users and vice versa.

If you're making a certificate provider, your certificates will use the same cryptography as what your site uses. This means that ArkID must use Ark cryptography, etc.

</details>

## Tutorial for users

There is none -- you just install the plugin that adds the cryptography you want.

## Tutorial for those who want to add a new cryptography

<details><summary>Click to expand</summary>

The recommended way is to add it via a plugin, which is really easy.

First, you create a `plugins/[name]` directory for your plugin. Example: `plugins/Ark`, `plugins/Etherium`.

Second, you create a file for your cryptography implementation inside that directory. `Crypt[name].py` is what *should* be used. Example: `CryptArk.py`, `CryptEtherium.py`. You'll need to use `Crypt.Crypt` module a lot, so you'd better import it:

```python
from Crypt import Crypt
```

Here's what you'll have to implement:

```python
newPrivatekey() -> str
hdPrivatekey(seed: str, child: int) -> str
privatekeyToAddress(privatekey: str) -> str
sign(data: str, privatekey: str) -> str
verify(data: str, valid_address: List[str]|str, sign: str) -> bool
isAddress(address: str) -> bool
```

### `newPrivatekey`

Generates a new random private key. Luckily, `Crypt` module contains a `newSeed` function that returns a random 256-bit key as a `bytes` object. Example:

```python
def newPrivatekey():
    return "super!" + Crypt.newSeed().hex()
```

### `hdPrivatekey`

Generates a private key deterministically from a seed and an index (called `child`). The seed is a 256-bit hexadimical string (so it contains of 64 hexadimical digits), the child is a small integer.

An ideal implementation should follow the contracts below:

```python
assert hdPrivatekey(a, b) == hdPrivatekey(a, b)  # The private key id generated deterministically
assert b == c or hdPrivatekey(a, b) != hdPrivatekey(a, c)  # Different children result in different private keys
assert a == c or hdPrivatekey(a, b) != hdPrivatekey(c, b)  # Different seeds result in different private keys
```

Example:

```python
def hdPrivatekey(seed, child):
    return "super!" + hashlib.sha256(seed + str(child)).hexdigest()
```

### `privatekeyToAddress`

Given a privatekey, transforms it into an address. In case the given private key wasn't generated by the current cryptography implementation, it should raise `Crypt.CryptError`.

Example:

```python
def privatekeyToAddress(privatekey):
    if not privatekey.startswith("super!"):
        raise Crypt.CryptError("The privatekey must start with 'super!'")
    return hashlib.sha256(privatekey).hexdigest()[:12]
```

### `sign`

Signs a given string with a given privatekey and returns the signature as a string (usually base64-encoded).

### `verify`

Verifies that a given signature really signs the given data, and that the signature was generated by `valid_address`. (or, if it's a list, by any of valid addresses). Should raise `Crypt.CryptError` if the signature wasn't generated by the current cryptography.

### `isAddress`

Given an address, checks that it was generated by the current cryptography implementation.

Example:

```python
def isAddress(address):
	return len(address) == 12 and all((c in "0123456789abcdef" for c in address))
```

---

Finally, you should *register* the cryptography, so that ZeroNet knows where to find its implementation. Create `plugins/[name]/__init__.py` file with the following content:

```python
from Crypt import Crypt
from . import Crypt[name]
Crypt.registerCrypto("[name]", Crypt[name])
```

Example:

```python
from Crypt import Crypt
from . import CryptSuper
Crypt.registerCrypto("Super", CryptSuper)
```

</details>

## Docs

<details><summary>Click to expand</summary>

First things first -- ZeroNet creates unique `auth_address` and `auth_privatekey` for every site that you visit. If the site is a certificate provider, it can use these values to generate a certificate for the user.

By default, site's `auth_address` & `auth_privatekey` use the same cryptography as the site uses.

This feature can be used to create different certificate providers (see [ARKidcu9P7HTpig7htMsxTgEuUAp3h34qg.zip](https://github.com/HelloZeroNet/ZeroNet/files/3337969/ARKidcu9P7HTpig7htMsxTgEuUAp3h34qg.zip)).

### Modify `siteClone`

`siteClone` has received a new argument. The signature has been changed to the following:

```python
siteClone(address: str, root_inner_path: str="", target_address: str=None, redirect: bool=True, crypto: str="Bitcoin")
```

In case the cryptography is not passed and the user doens't have any other cryptographies, the behaviour isn't changed.

In case the cryptography is not passed, but the user has several cryptographies, they get the following question:

image

In case several cryptographies are passed (i.e. as a list), the list above is filtered to only include the allowed cryptographies.

In case the cryptography is passed and only one cryptography is allowed, the behaviour is the same as before this PR (the only difference is the cryptography actually used).

</details>

## Implementation details

<details><summary>Click to expand</summary>

All the core & plugins have been updated to use a new `Crypt.Crypt` module instead of `Crypt.CryptBitcoin`, and `Crypt.CryptBitcoin` has been updated to be no different from any other cryptography implementation.

First, we'll discuss how using `CryptBitcoin` is different from using `Crypt`.

### Rename `CryptBitcoin.privatekeyToAddress` -> `Crypt.privatekeyToAddress`

No changes except renaming.

### Rename `CryptBitcoin.sign` -> `Crypt.sign`

No changes except renaming.

### Rename `CryptBitcoin.verify` -> `Crypt.verify`

No changes except renaming.

### Modify `CryptBitcoin.newPrivatekey` -> `Crypt.newPrivatekey`

Old signature:

```python
newPrivatekey() -> str
```

New signature:

```python
newPrivatekey(crypto: str) -> str
```

### Modify `CryptBitcoin.hdPrivatekey` -> `Crypt.hdPrivatekey`

Old signature:

```python
hdPrivatekey(seed: str, child: int) -> str
```

New signature:

```python
hdPrivatekey(crypto: str, seed: str, child: int) -> str
```

</details>


filips123:
Is it also possible to use multiple cryptographies at the same time? Can you add support for this?

imachug:
Yeah, I'll add FAQ right now.

imachug:
Added simple FAQ, working on docs right now.

filips123:
Will system detect which cryptography is used and automatically switch to it?

Also, I think that you should now also create an Ethereum plugin because it is very widely used.

imachug:
> Will system detect which cryptography is used and automatically switch to it?

Yes, take a look at the tutorial I've just added.

> Also, I think that you should now also create an Ethereum plugin because it is very widely used.

Will do.

imachug:
The docs have been updated & finished.

imachug:
CI is failing on old Python versions because of f string literals, which were added in Python 3.6. I will replace them with .format calls soon.

HelloZeroNet:
The class filenames has the prefix of the directory, so CryptEcc.py would be better. 
Also renaming the WrongCryptoError error to CryptEccError and dropping some meaningful message would make it easier to debug.
Does the actionChangeCrypto necessary? I think it should generate and use the same cryptography as the site does otherwise we will run into sync issues. (eg posting content signed with crypto that is not supported by every client)

imachug:
> Does the actionChangeCrypto necessary? I think it should generate and use the same cryptography as the site does otherwise we will run into sync issues. (eg posting content signed with crypto that is not supported by every client)

While `actionChangeCrypto` might not be necessary if we assume "default crypto = site crypto", I disagree with using a single cryptography per site. For example, a forum should support different cryptographies.

So, my proporsal is: get rid of `changeCrypto`, assume the default crypto (when no certificate is used) matches the cryptography of the site.

imachug:
I also vote against `CryptEcc` because we should support classic cryptographies (i.e. non-elliptic) as well. We should come up with a better name. However, I can't really make up something better than "cryptography" -- and `CryptCryptography` sounds... weird :)

HelloZeroNet:
Yeah probably CryptEcc is not the best idea, but we could have a Crypt/Crypt.py and import it with `from Crypt import Crypt` as we have Site/Site.py or Peer/Peer.py

imachug:
Yeah, sounds great, will do that.

imachug:
Changelog:

1. `changeCrypto` was removed
2. A new `isAddress` method added to cryptography implementations
3. `Cryptography` renamed to `Crypt`
4. `WrongCryptoError` renamed to `CryptError`

Working on better error messages & docs right now.

imachug:
The docs have been updated.

Thunder33345:
what's the differences of ark vs bitcoin crypto?
and why should i as site owner consider using one or the other?
some sort of explanation seems to be lacking

imachug:
Hello, @Thunder33345! Great to see you here again.

> what's the differences of ark vs bitcoin crypto?

They both use the same elliptic curves, so there isn't much difference between them -- they work in a similar way and are equally secure. The reason why I used Ark was just because someone asked me to help him to replace Bitcoin with Ark, and I decided to support both. The purpose of this PR is to allow *other* cryptographies to be added too.

> and why should i as site owner consider using one or the other?

How the site private key is mapped to the public key and to the address depends on what cryptography to use. Thus, it's impossible to support many cryptographies at once, just because they would yield different addresses. Because of this, a site must use a single cryptography for content signing. However, users may use other cryptographies. For example, you can use ZeroID at some Ark forum.

Thunder33345:
ah so this purpose is just basically a move to diversify of algos begin used instead of relying only on bitcoin
that make sense i suppose
for reference i was more of wondering the purpose of this PR or why ark should be supported
so there wont be much differences but just more options, might be confusing for newer users though...

>Hello, @Thunder33345! Great to see you here again.
yeah just occasionally checking over gh, nice seeing py3 support begin worked on, maybe i should get back on matrix and 0net itself haha

imachug:
> ah so this purpose is just basically a move to diversify of algos begin used instead of relying only on bitcoin

Exactly.

> so there wont be much differences but just more options, might be confusing for newer users though...

Right. However, we can add some explanation to the "clone" window. Or, as the cryptographies are only added by plugins, we can make the plugin authors explain it :)

Thunder33345:
>Exactly.

ok this now make sense
probably reiterating the reason of addition is to be able to diversify cryptographic algorithms and not to only rely on BTC

i wonder what's the possibility of that "crypto" that's said to be resistant to quantum computing
maybe adding that will make someone happy haha(cryptographic algorithms to be added is probably out of scope of this thread though)

how about client corss compatibility?
say i have an outdated node on zerotalk, would the arkid users just be basically "invisible" to me right?

imachug:
> how about client backward compatibility?
> say i have an outdated node on zerotalk, would the arkid users just be basically "invisible" to me right?

Right. You see people who use any cryptography among the ones you have installed.

filips123:
@imachug @HelloZeroNet I saw a question about SegWit addresses on [ZeroTalk](http://127.0.0.1:43110/Talk.ZeroNetwork.bit/?Topic:1561983082_1EBYmm4sYxPoUVKAYFLfmYR6opizGGmM8v). Are SegWit addresses already supported by ZeroNet? If not, could they be supported by this PR?

imachug:
AFAIK segwit isn't supported now, but it would be easy to support it.

HelloZeroNet:
Yeah there is no support for segwit addresses yet

rllola:
Do we need to support segwit addresses ? We can make donation through the address site so it would be for this purpose ?

HelloZeroNet:
the only real benefit that I see is bech32 addresses are case-insensitive ones. So in the future `zero://bc1qw..` like access could work on browsers that support custom protocol handlers. It would not possible for current addresses, because the browsers converts all hosts to lower-case ones.

zeroexploding:
absurd! :nauseated_face: :no_entry:

PR #2059, opened 2019-06-30
Allow sites to lock pointer

### LINKED ISSUE #2057: Allow Pointer Lock API in iframe
[Pointer Lock API](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API) is blocked in sandboxed iframes. Can you allow it for usage?

imachug:
Add `allow-pointer-lock` to iframe sandbox.

Fixes #2057.

### DIFF
--- a/src/Ui/template/wrapper.html
+++ b/src/Ui/template/wrapper.html
@@ -70,7 +70,7 @@ else if (window.opener && window.opener.location.toString()) {
 
 
 <!-- Site Iframe -->
-<iframe src='about:blank' id='inner-iframe' sandbox="allow-forms allow-scripts allow-top-navigation allow-popups allow-modals allow-presentation {sandbox_permissions}" allowfullscreen="true" webkitallowfullscreen="true" mozallowfullscreen="true" oallowfullscreen="true" msallowfullscreen="true"></iframe>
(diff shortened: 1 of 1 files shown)


imachug:
Offtopic: CI is failing because of incorrect pytest configuration on Python 3.8, not sure what's going on.

rllola:
> Offtopic: CI is failing because of incorrect pytest configuration on Python 3.8, not sure what's going on.

Seems to be because of this : `pytest.py: error: unrecognized arguments: -rsxX`. It should be a new issue.

imachug:
What's interesting is that there weren't any commits to `py3` recently, and the CI on my other recent PRs *did* succeed...

rllola:
Yeah but pytest seems to have new commits for python 3.8 version and I think it is where it is coming from (https://github.com/pytest-dev/pytest)

imachug:
Yay, thanks!

Thunder33345:
shouldnt features like this better if it require permission/consent from user first to authorize it?

imachug:
I've been thinking about this as well, and this is what I came up with:

1. Browsers don't allow sites to lock pointer without any user interaction
2. Locking pointer results in a big alert shown

Notice that no permission is asked. However, the user would notice that something is wrong, and leave the site.

So, I think that doing the same in ZeroNet won't bring any problems.

Thunder33345:
i suppose so then users could always ctrl+w or whatever if it comes to that

PR #2060, opened 2019-07-01
Fix siteCloning subdirectory with -default files

imachug:
How to reproduce:

Put `ZeroHello/template-new/data/users/content.json-default` file, and add it to `includes` to `ZeroHello/template-new/content.json-default`.


HelloZeroNet:
Thanks!

PR #2064, opened 2019-07-03
Prevent ZeroNet from running if Python interpreter is optimized

beardog108:
ZeroNet (and dependencies) rely on assert statements in a few places, notably in [certificate pinning](https://github.com/HelloZeroNet/ZeroNet/blob/eb88dbbec8d88f58c5c4c060a61f74c9074e591e/src/Crypt/CryptConnection.py#L53)

My solution is to use assert to test if the Python interpreter is ran in optimized mode (-O) and exit if so. I doubt optimized mode really speeds up ZeroNet much anyway, so there is nothing of value lost. This just prevents edge case bugs and security issues resulting from the assert statements being ignored.

[See this article](https://hackernoon.com/10-common-security-gotchas-in-python-and-how-to-avoid-them-e19fbe265e03)

I also switched the Python version detection to exit with code 1 because error-exits should be non-zero.

@shortcutme

### DIFF
--- a/zeronet.py
+++ b/zeronet.py
@@ -8,8 +8,17 @@ import sys
 def main():
     if sys.version_info.major < 3:
         print("Error: Python 3.x is required")
-        sys.exit(0)
+        sys.exit(1)
 
+    # Test if Python is being ran in optimized mode (-O) because ZN relies on assert
+    try:
+        assert True is False # Intentionally get an AssertionError
+    except AssertionError:
+        pass
+    else:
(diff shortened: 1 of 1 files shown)


imachug:
Wouldn't it be better to replace assert with if in critical places?

beardog108:
That would be ideal, yes

HelloZeroNet:
Thanks for reporting I have replaced asserts with ifs in the code: https://github.com/HelloZeroNet/ZeroNet/commit/ff32f822ba65b1d7c8a13f394001304db9db641d

PR #2067, opened 2019-07-03
fix pytest version to 4.6.3 to avoid it to broke for python 3.8 dev

rllola:
I have open an issue to have a bit more information on what has been changed on the latest version. (https://github.com/pytest-dev/pytest/issues/5545)


HelloZeroNet:
Thanks

PR #2069, opened 2019-07-04
Guess content type correctly

imachug:
Fix e.g. vue.min.js being reported as text/plain instead of text/javascript.


HelloZeroNet:
Thanks

PR #2078, opened 2019-07-07
Move the BitTorrent related code from SiteAnnouncer.py

geekless:
Move the BitTorrent related code from SiteAnnouncer.py to a separate plugin.

The support of BitTorrent trackers can now be enabled/disabled by enabling/disabling the plugin.

config.disable_udp is not in use by the core now, but I haven't moved its definition to the plugin. Since it's a general-purpose network option, it can be used in other modules in the future. For example, in a DHT implementation, or in a UDP-based trasport protocol.


HelloZeroNet:
Looking good, but the Travis CI failed, because of the missing `import Debug`

geekless:
> Looking good, but the Travis CI failed, because of the missing `from Debug import Debug`

Fixed.

Issue #1986, opened 2019-04-14
Wrong fonts content-type when UiPassword is enabled


SkRiMiX:
### Step 1: Please describe your environment

  * ZeroNet version: **0.7.0 r4059**
  * Operating system: **Arch Linux**
  * Web browser: **Firefox 66**
  * Tor status: **enabled**
  * Opened port: **yes**
  * Special configuration: 

### Step 2: Describe the problem:
When UiPassword is enabled, some files, fonts are sent with `Content-Type: text/html; charset=utf-8`, getContentType doesn't get called on them. Also `Access-Control-Allow-Origin` header is missing.
#### Steps to reproduce:

  1. Open site `http://127.0.0.1:43110/14c5LUN73J7KKMznp9LvZWkxpZFWgE1sDz/` with UiPassword disabled.
  2. Content-type of `http://127.0.0.1:43110/14c5LUN73J7KKMznp9LvZWkxpZFWgE1sDz/fonts/material-icons/MaterialIcons-Regular.woff2` is `application/octet-stream`.
  3. Enable UiPassword and reload site.
  4. Content-type of `http://127.0.0.1:43110/14c5LUN73J7KKMznp9LvZWkxpZFWgE1sDz/fonts/material-icons/MaterialIcons-Regular.woff2` is `text/html; charset=utf-8` and browser blocks request because `Access-Control-Allow-Origin` header is missing.

#### Observed Results:

Fonts fail to load because of wrong content-type and missing `Access-Control-Allow-Origin` header.
I added `print('getContentType called: file_name={}, content_type={}'.format(file_name, content_type))` to getContentType and this is what I got:
UiPassword disabled:

> getContentType called: file_name=index.html, content_type=text/html
> getContentType called: file_name=all.css, content_type=text/css
> getContentType called: file_name=all.css, content_type=text/css
> getContentType called: file_name=all.js, content_type=application/javascript
> getContentType called: file_name=logo-white.png, content_type=image/png
> getContentType called: file_name=index.html, content_type=text/html
> getContentType called: file_name=fonts/material-icons/material-icons.css, content_type=text/css
> getContentType called: file_name=material-icons.css, content_type=text/css
> getContentType called: file_name=css/main.css, content_type=text/css
> getContentType called: file_name=main.css, content_type=text/css
> getContentType called: file_name=css/asciinema-player.css, content_type=text/css
> getContentType called: file_name=asciinema-player.css, content_type=text/css
[...]

Issue #1940, opened 2019-03-27
A bug about content-type?


tangdou1:
### Step 1: Please describe your environment

  * ZeroNet version: 3860
  * Operating system: `remote zeronet on Debian9` / `local zeronet on windows 7`
  * Web browser: _firefox66
  * Tor status: not available
  * Opened port: yes
  * Special configuration: _no___

#### Steps to reproduce:

  1. Open the zite `http://127.0.0.1:43110/14c5LUN73J7KKMznp9LvZWkxpZFWgE1sDz/` through `local` ZeroNet
  2. Press F12, you can find the content-type of `http://127.0.0.1:43110/14c5LUN73J7KKMznp9LvZWkxpZFWgE1sDz/fonts/material-icons/MaterialIcons-Regular.woff2` is `application/octet-stream`.
  3.  However, Open the zite `http://remote zeronet's ip:port/14c5LUN73J7KKMznp9LvZWkxpZFWgE1sDz/` through `remote` ZeroNet.
 4. Press F12, you can find the content-type of `http://remote zeronet's ip:port/14c5LUN73J7KKMznp9LvZWkxpZFWgE1sDz/fonts/material-icons/MaterialIcons-Regular.woff2` is `text/html; charset=utf-8`, which means the font source would be blocked by browser due to cors problem and the content-type is also wrong.

#### Observed Results:

 Some source cannot be read when using remote zeronet.

#### Expected Results:

the content-type should be consistent.

tangdou1:
maybe it is just a python version problem. (Debian 9 use python 2.7.13)

HelloZeroNet:
What happens if you open the `http://remote zeronet's ip:port/14c5LUN73J7KKMznp9LvZWkxpZFWgE1sDz/fonts/material-icons/MaterialIcons-Regular.woff` directly? Does the file content got displayed?

tangdou1:
> 
> 
> What happens if you open the `http://remote zeronet's ip:port/14c5LUN73J7KKMznp9LvZWkxpZFWgE1sDz/fonts/material-icons/MaterialIcons-Regular.woff` directly? Does the file content got displayed?

@HelloZeroNet 
It will be downloaded directly.

[MaterialIcons-Regular.zip](https://github.com/HelloZeroNet/ZeroNet/files/3016080/MaterialIcons-Regular.zip)

tangdou1:
[...]

PR #2080, opened 2019-07-07
Fix font route bug when ui-password plugin is enabled

tangdou1:
see issues:
https://github.com/HelloZeroNet/ZeroNet/issues/1986
https://github.com/HelloZeroNet/ZeroNet/issues/1940


HelloZeroNet:
The problem is this way anyone could list what sites are served on the host without entering the password.

PR #2081, opened 2019-07-07
Fix font route bug -py3

tangdou1:
see issues:
https://github.com/HelloZeroNet/ZeroNet/issues/1986
https://github.com/HelloZeroNet/ZeroNet/issues/1940


krixano:
@tangdou1 Is this a finished fix for those issues? Why did you close it? We really need a fix for this stuff.

PR #2089, opened 2019-07-10
Fix a websocket connection issue when running in the proxy mode

geekless:
Steps to reproduce:

1. Run Zeronet with --ui_trans_proxy in a docker container or other isolated environment.
2. Set 127.0.0.1:43110 as a HTTP Proxy in a browser.
3. Navigate to http://talk.zeronetwork.bit/
4. The site stops loading, since the websocket connection isn't functional.

The bug consists of 2 related issues:

1. Connect to the proper address.

   When running in the proxy mode, Zeronet sends its env["SERVER_NAME"] and env["SERVER_PORT"] as the websocket destination address in the wrappper HTML code (variable `server_url`). That looks wrong for me, since the values of those variables have nothing to do with the actual address accessible by the browser. ZeroNet can run in a container, the connection also can be tunneled over SSH and so on. In all those cases, the internal SERVER_NAME makes no sense as the destination at all.

   The code for sending and handling `server_url` is deleted. The browser should connect to the websocket in the same way as it connects to the ZeroNet UI in general.

2. Send the proper response for CONNECT requests.

   A browser tries to connect to the correct address, but it still fails. A browser, when connecting via a HTTP proxy to a webdocket, uses the [HTTP tunneling protocol](https://en.wikipedia.org/wiki/HTTP_tunnel), and ZeroNet lacks support for it. To fix this, it's enoght to send the "200 OK" response on the CONNECT method request.

### DIFF
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -81,6 +81,10 @@ class UiRequest(object):
         if config.ui_restrict and self.env['REMOTE_ADDR'] not in config.ui_restrict:
             return self.error403(details=False)
 
+        if self.env['REQUEST_METHOD'] == "CONNECT":
+            self.start_response("200 OK", [])
+            return ""
+
         # Check if host allowed to do request
         if not self.isHostAllowed(self.env.get("HTTP_HOST")):
             return self.error403("Invalid host: %s" % self.env.get("HTTP_HOST"), details=False)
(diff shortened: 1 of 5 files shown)


geekless:
We still have some issues here.

`UiRequest` doesn't allow what it does:

> http://zero/1HeLLo4uzjaLetFx6NH3PMwFP3qbRbTf3D/ returns "Forbidden"

And allows what it doesn't:

> CONNECT accepts just any connection. So Firefox tryes to open SSL connections to *.firefox.com and sends encripted garbage there.

But this should be fixed in a different pull request, I guess, as it affects other parts of `UiRequest`.

PR #2096, opened 2019-07-15
Add missing @helper.encodeResponse

geekless:
Add missing @helper.encodeResponse in StatsPlugin.py and BootstrapperPlugin.py


HelloZeroNet:
Thanks!

PR #2108, opened 2019-07-25
Update README.md

slrslr:
Without skipping certificate check, an "wget" error may be shown on Debian 9 ARM:

ERROR: The certificate of 'github.com' is not trusted.
ERROR: The certificate of 'github.com' hasn't got a known issuer.


HelloZeroNet:
You have to install ca-certificates package to have proper ssl with wget. Adding the --no-check-certificate would allow mitm the data transfer, so it's not a good idea.

PR #2114, opened 2019-07-30
Save content.json of site even if limit size is reached

### LINKED ISSUE #2107: Py3 : not able to able to update site

  * ZeroNet version: rev4112
  * Operating system: ubuntu 18.04
  * Web browser: Firefox
  * Tor status: not available
  * Opened port: yes/no
  * Special configuration: No

Trying to update a site but I have a bad file error.

#### Steps to reproduce:

  1. Run py3 version of Zeronet
  2. Try to update 

#### Observed Results:

No update happening. It is working though on python2 version.

Log (*I have remove the peer ip) :

```
[2019-07-25 16:45:28,670] DEBUG Site:1ChMNj..2RTL Queried listModifications from: [<Peer:>, <Peer:>, <Peer:>, <Peer:>, <Peer:>] in 0.800s since 1538157251
[2019-07-25 16:45:28,772] DEBUG Site:1ChMNj..2RTL content.json loadContent same json file, skipping
[2019-07-25 16:45:28,772] DEBUG Site:1ChMNj..2RTL Bad files: {'content.json': 3}
[2019-07-25 16:45:28,774] DEBUG Site:1ChMNj..2RTL Retry 1 bad files
[2019-07-25 16:45:28,775] DEBUG Site:1ChMNj..2RTL New downloadContent pool: len: 1, only if bad: True
[2019-07-25 16:45:28,776] DEBUG Site:1ChMNj..2RTL Ended downloadContent pool len: 1, skipped: 0
[2019-07-25 16:45:28,777] DEBUG WorkerManager:1ChMNj..2RTL Starting workers (Added new task), tasks: 1, peers: 0, workers: 1
[2019-07-25 16:45:28,778] DEBUG Site:1ChMNj..2RTL Recent peers 8 of 44 (need: 10)
[2019-07-25 16:45:28,778] DEBUG WorkerManager:1ChMNj..2RTL Added worker: 212.24.103.49:29535, workers: 2/5
[2019-07-25 16:45:28,779] DEBUG WorkerManager:1ChMNj..2RTL Added worker: 82.130.48.43:16571, workers: 3/5
[2019-07-25 16:45:28,779] DEBUG WorkerManager:1ChMNj..2RTL Added worker: 177.139.86.28:31031, workers: 4/5
[2019-07-25 16:45:28,780] DEBUG WorkerManager:1ChMNj..2RTL Added worker: 91.121.106.25:27485, workers: 5/5
[2019-07-25 16:45:28,780] DEBUG WorkerManager:1ChMNj..2RTL Added worker: 188.116.183.41:26552, workers: 6/5
[2019-07-25 16:45:28,851] DEBUG Site:1ChMNj..2RTL Can't update content.json
[2019-07-25 16:45:28,854] WARNING Site:1ChMNj..2RTL content.json: verify sign error: VerifyError: Content too large 15638244B > 10485760B, aborting task... in ContentManager.py line 991 > ContentManager.py line 869
[2019-07-25 16:45:28,855] DEBUG WorkerManager:1ChMNj..2RTL 212.24.103.49:29535: Verify failed: content.json, error: Content too large 15638244B > 10485760B, aborting task..., failed peers: 0, workers: 0
[2019-07-25 16:45:28,856] DEBUG Site:1ChMNj..2RTL Download content.json failed, check_modifications: False
```

#### Expected Results:

Site to update

rllola:
fix #2107; Still save the content.json received even if site size limit is reached but dont download files; Allow better distribution of latest version of content.json

### DIFF
--- a/src/Content/ContentManager.py
+++ b/src/Content/ContentManager.py
@@ -859,14 +859,15 @@ class ContentManager(object):
             raise VerifyError("Wrong inner_path: %s" % content["inner_path"])
 
         # Check total site size limit
-        if site_size > site_size_limit:
-            if inner_path == "content.json" and self.site.settings["size"] == 0:
-                # First content.json download, save site size to display warning
-                self.site.settings["size"] = site_size
-            task = self.site.worker_manager.findTask(inner_path)
-            if task:  # Dont try to download from other peers
-                self.site.worker_manager.failTask(task)
-            raise VerifyError("Content too large %sB > %sB, aborting task..." % (site_size, site_size_limit))
+        # Should not be checked here
+        #if site_size > site_size_limit:
+        #    if inner_path == "content.json" and self.site.settings["size"] == 0:
+        #        # First content.json download, save site size to display warning
+        #        self.site.settings["size"] = site_size
+        #    task = self.site.worker_manager.findTask(inner_path)
+        #    if task:  # Dont try to download from other peers
+        #        self.site.worker_manager.failTask(task)
+        #    raise VerifyError("Content too large %sB > %sB, aborting task..." % (site_size, site_size_limit))
 
         # Verify valid filenames
         for file_relative_path in list(content.get("files", {}).keys()) + list(content.get("files_optional", {}).keys()):
--- a/src/Content/ContentManager.py
+++ b/src/Content/ContentManager.py
@@ -869,6 +869,17 @@ class ContentManager(object):
         #        self.site.worker_manager.failTask(task)
         #    raise VerifyError("Content too large %sB > %sB, aborting task..." % (site_size, site_size_limit))
 
+        # If our content.json file bigger than the size limit throw error
+        if inner_path == "content.json":
+            content_size_file = len(json.dumps(content, indent=1))
+            if content_size_file > site_size_limit:
+                # Save site size to display warning
+                self.site.settings["size"] = site_size
+                task = self.site.worker_manager.findTask(inner_path)
+                if task:  # Dont try to download from other peers
+                    self.site.worker_manager.failTask(task)
+                raise VerifyError("Content too large %s B > %s B, aborting task..." % (site_size, site_size_limit))
+
         # Verify valid filenames
         for file_relative_path in list(content.get("files", {}).keys()) + list(content.get("files_optional", {}).keys()):
             if not self.isValidRelativePath(file_relative_path):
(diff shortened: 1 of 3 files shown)


imachug:
What if the site owner (or, in case of multiuser sites, any user) distributes a 1 TB content.json?

rllola:
Good question. We were already downloading it but not writing to disk so I guess even without this PR we still would be downloading 1TB... My guess is that there is already a size limit for file (at least I think I saw this).

One of the main advantage of this is that we would not distribute outdated `content.json` because the client haven't accepted to raise size limit (however we lose the old `content.json` but it should not impact the use of the site). Also it will show again the message on ZeroHello that the limit need to be raised which is not showing anymore (took me sometime to understand why I couldnt update my site and test it on py3 branch).

HelloZeroNet:
As it could be good to add some test how does it works if the site or an user content exceed the allowed site size.

rllola:
For sure we need some. I will add a couple. Better safe than sorry.

@HelloZeroNet Can you confirm that the `content.json` is not saved to disk if the site size is bigger than the limit ? Just want to be sure this is indeed the current way of doing.

rllola:
I have added a test and the huge `content.json` file is downloading on the other client. So I need to fix that.

rllola:
@HelloZeroNet @imachug This can be reviewed.

HelloZeroNet:
Looking good, but we have to figure out what should happen with huge content.json files. Maybe allow up until current size limit + 10%?

Also please remove the commented code if it's no longer necessary.

rllola:
Is it interesting to have a hard limit for content.json size ? Is it a good thing to allow contnt.json file bigger than 10MB ? I think this can be a hard limit and enforce when signing. I don't see the need for a file that big.

HelloZeroNet:
If you have user submitted content, then it can be a problem as if you site limit is reached, then it will not allow you to send updates.

rllola:
With this if the site content is bigger than 10MB it will offer you to update the limit. However I don't think it will do it if the `content.json` file size is over 10MB because it won't save such a big file.

Yeah there is a different between the content size being over 10MB and the `content.json` file being more than 10MB. I am talking about the latest case.

rllola:
@HelloZeroNet What should I do with this ? I believe it is still relevant. Should I update the test ? I am not sure I understood what you meant in the last comment. I might have missed something.

HelloZeroNet:
We can improve it later, so thanks for the PR

Issue #2128, opened 2019-08-07
Execution order of ZeroNet plugins


filips123:
In which order do ZeroNet plugins execute when registering to core classes?

I tested this and it seems to be in reverse alphabetical order. Is that right?

---

Because of that, I have a problem with developing the ZeroNet plugin for Ethereum Name Service. It will be called `ENS` so it will be executed somewhere in the middle. This is a problem because plugins that are executed before it won't have resolved domain so they will probably crash. This isn't a problem for NameCoin plugin because it is called `Zeroname` so it is executed first.

Is it possible to fix this? A quick solution would be to rename my plugin to `ZeroENS` but there should be better handling of plugin order.

---

There needs to be support to specify priority and execution order of plugins.

There need to be a few priority classes reserved for a specific type of plugin:

* High - For DNS plugins and other plugins that need to be executed first
* Medium - Normal plugins
* Low - Network plugins that need to be executed last

So I would recommend documenting those three priority classes (maybe more if needed) and represent them with numbers:

* High - 100
* Medion - 50
* Low - 0

Here, the highest number would execute first. Plugins with the same number should be executed in reverse alphabetical order (as currently).

This would allow plugins to categorize into groups, but still allow exceptional plugins to be executed before or after majority or other plugins. Normal plugins should be using one of those three numbers but they would also be able to use other numbers if needed.

rllola:
I think it is right because it is going to check all the folders in the `plugins` folder which doesn't start with `disabled-`.

As far as I know this behavior is going to be modified as we are going to explicitly ask to activate plugins inside the config file. You might be able to specify the order then.

filips123:
> As far as I know this behavior is going to be modified as we are going to explicitly ask to activate plugins inside the config file. You might be able to specify the order then.

Yes, but it would be better to also allow plugins themselves to define when they should be executed. So something like `@PluginManager.registerTo('SiteManager', priority=50)` where priority should be an optional argument (and set to 50 by default) to maintain backwards compatibility.

HelloZeroNet:
The plugins are loaded alphabetically, but if you subclass the same class multiple times, then the last will be called as it's overrides the previous function (then you can call the previous function using super())

There could be some cases where the order matters, but I think for domain resolver it should note be an issue.

rllola:
Do you have the source code of your plugin somewhere ? Maybe we can brainstorm to help you fix your problem.

[...]

PR #2135, opened 2019-08-10
Added super call for resolveDomain function

rllola:
Related to #2128

### DIFF
--- a/plugins/Zeroname/SiteManagerPlugin.py
+++ b/plugins/Zeroname/SiteManagerPlugin.py
@@ -50,7 +50,7 @@ class SiteManagerPlugin(object):
                 (len(self.db_domains), time.time() - s, self.db_domains_modified, site_zeroname_modified)
             )
             self.db_domains_modified = site_zeroname_modified
-        return self.db_domains.get(domain)
+        return self.db_domains.get(domain) or super(SiteManagerPlugin, self).resolveDomain(address)
 
     # Return or create site and start download site files
     # Return: Site or None if dns resolve failed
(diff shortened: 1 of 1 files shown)


HelloZeroNet:
The problem is there is no resolveDomain function in the SiteManager, so it will drop exception on non-existent domains.

I think it's better to override the get() and need() function instead in the other plugin and not the resolveDomain one.

Other solution would be adding isDomain(address) and resolveDomain(address) to SiteManager.py that is always return False by default and name resolving plugins could override these functions. (we have to rename resolveDomain to resolveBitDomain in Zeroname plugin)

rllola:
The other plugin has a `get()` and `need()` function and it is failing.

> Other solution would be adding isDomain(address) and resolveDomain(address) to SiteManager.py that is always return False by default and name resolving plugins could override these functions. (we have to rename resolveDomain to resolveBitDomain in Zeroname plugin)

I can try that but isn't it going to fail when others call `resolveDomain` ?

rllola:
What I understand is happening is that `actioinSiteMedia` from `ZeroName` plugin is called it resolve the domain name (with `resolveDomain`) :
https://github.com/HelloZeroNet/ZeroNet/blob/py3/plugins/Zeroname/UiRequestPlugin.py#L15-L22

Then it is call `actioinSiteMedia` from `TranslateSite` which is called and then the `ENS` plugin. So ENS plugin hasn't resolve the domain name and `TranslateSite` is left with `example.eth` instead of a proper address.

rllola:
Now that I think about it ordering the plugins might sound like a better idea but it would leave too much freedom and therefore might implies lots of support...

Otherwise is there any way to decouple resolving the domain from the `actionSiteMedia` call... ?

rllola:
Closing this as modification has been added in https://github.com/HelloZeroNet/ZeroNet/commit/2bdd07360854bb7920c52efc28913211e4d47ab1

PR #2141, opened 2019-08-12
Allow all valid filenames to be added to content.json

### LINKED ISSUE #2136: Files with "!" in the name are not added to content.json

  * ZeroNet version: 0.7.0 rev4172
  * Operating system: W7x64
  * Web browser: N/A
  * Tor status: N/A
  * Opened port: N/A
  * Special configuration: No

Files that contain `!` (exclamation mark) in the file name are not added to `content.json`.

Note: This seems to apply to a few other characters too, including but not limited to `'` (apostrophe) and `&` (ampersand)

#### Steps to reproduce:

  1. Create blank site
  2. Create a few files with and without an exclamation mark in the name
  3. Set site name and description, then sign and publish the site

#### Observed Results:

When setting the site name and description, the directory is scanned for files, and they are added to the `content.json` index. Files containing an exclamation mark are not added to the file and thus are not shared with others.

The debug log contains lines like these (some information has been redacted by me):

    [2019-08-11 ███] ERROR    Site:███ - [ERROR] Invalid filename: Test !.txt
    [2019-08-11 ███] INFO     Site:███ - [SKIPPED] Test !.txt

#### Expected Results:

No seemingly arbitrarily chosen file name rules in place. Any file name made up of a valid file name character sequence should be added to the `content.json` file.

### Related problems

Files/directories that start with `.` or end in `-old` or `-new` are hardcoded to be excluded too.

This puts a huge dampener on sharing git repositories or creating the `.well-known` folder. (See [RFC 5785](https://tools.ietf.org/html/rfc5785))

### Notes

The offending location for the character selection seems to be https://github.com/HelloZeroNet/ZeroNet/blob/5a746769d01db58d54ff121e784fbf87374a89f3/src/Content/ContentManager.py#L603

joshiemoore:
Fix #2136 by allowing all valid filenames to be added to content.json when a site is published.

I basically used the new regex expressions that the creator of the issue suggested, modified to allow files in subdirectories to also be added to content.json.

### DIFF
--- a/src/Content/ContentManager.py
+++ b/src/Content/ContentManager.py
@@ -600,10 +600,10 @@ class ContentManager(object):
         elif len(relative_path) > 255:
             return False
         else:
-            return re.match(r"^[^\x00-\x1F\x22\x2A\x3A\x3C\x3E\x3F\x5C\x7C]+$", relative_path)
+            return re.match(r"^[^\x00-\x1F\"*:<>?\\|]+$", relative_path)
 
     def sanitizePath(self, inner_path):
-        return re.sub("[\x00-\x1F\x22\x2A\x3A\x3C\x3E\x3F\x5C\x7C]", "", inner_path)
+        return re.sub("[\x00-\x1F\"*:<>?\\|]", "", inner_path)
 
     # Hash files in directory
     def hashFiles(self, dir_inner_path, ignore_pattern=None, optional_pattern=None):
--- a/src/Content/ContentManager.py
+++ b/src/Content/ContentManager.py
@@ -600,10 +600,10 @@ class ContentManager(object):
         elif len(relative_path) > 255:
             return False
         else:
-            return re.match(r"^[a-z\[\]\(\) A-Z0-9~_@=\.\+-/]+$", relative_path)
+            return re.match(r"^[^\x00-\x1F\x22\x2A\x3A\x3C\x3E\x3F\x5C\x7C]+$", relative_path)
 
     def sanitizePath(self, inner_path):
-        return re.sub("[^a-z\[\]\(\) A-Z0-9_@=\.\+-/]", "", inner_path)
+        return re.sub("[\x00-\x1F\x22\x2A\x3A\x3C\x3E\x3F\x5C\x7C]", "", inner_path)
 
     # Hash files in directory
     def hashFiles(self, dir_inner_path, ignore_pattern=None, optional_pattern=None):
--- a/src/Test/TestContent.py
+++ b/src/Test/TestContent.py
@@ -246,3 +246,8 @@ class TestContent:
             with site.storage.open("data/users/1C5sgvWaSgfaTpV5kjBCnCiKtENNMYo69q/content.json") as data:
                 site.content_manager.verifyFile("data/users/1C5sgvWaSgfaTpV5kjBCnCiKtENNMYo69q/content.json", data, ignore_same=False)
         assert "Potentially unsafe" in str(err.value)
+
+
+    @pytest.mark.parametrize("filename", ["test.txt", "test/!@#$%^&().txt", "ÜøßÂŒƂÆÇ.txt"])
+    def testPathValidation(self, site, filename):
+        assert site.content_manager.isValidRelativePath(filename)
--- a/src/Test/TestContent.py
+++ b/src/Test/TestContent.py
@@ -248,6 +248,10 @@ class TestContent:
         assert "Potentially unsafe" in str(err.value)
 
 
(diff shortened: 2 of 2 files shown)


HelloZeroNet:
Please use the suggested non-hex version of the pattern: `^[^\x00-\x1F"*/:<>?\\|]+$` and also adding a test os syncing files with utf8 encoded filenames would be great as it's not 100% that it will work

joshiemoore:
I replaced the hex version with the non-hex version (modified for syntax and functionality), could you clarify what you mean by a "test os"?

HelloZeroNet:
Sorry, I meant "test for syncing files with utf8 encoded filename"

joshiemoore:
I'm not sure if this is the sort of thing you're looking for, I'm not that experienced in writing tests for projects like this.

imachug:
I think using several `assert`s instead of `parametrize` is the way we write multitests. Please use the following test instead:

```python
    def testPathValidation(self, site):
        assert site.content_manager.isValidRelativePath("test.txt")
        assert site.content_manager.isValidRelativePath("test/!@#$%^&*().txt")
        assert site.content_manager.isValidRelativePath("ÜøßÂŒƂÆÇ.txt")
        assert site.content_manager.isValidRelativePath("тест.текст")
        assert site.content_manager.isValidRelativePath("𝐮𝐧𝐢𝐜𝐨𝐝𝐞𝑖𝑠𝒂𝒘𝒆𝒔𝒐𝒎𝒆")
```

I've added a few more testcases, BTW. @HelloZeroNet I'm not sure if the last one should be allowed or not.

purplesyringa (approved):

AyrA:
@imachug I would say all characters should be allowed. It's not the task of zeronet to decide if a file name is valid, this is the decision of the operating system. The only reasonable check for file names is to make sure they don't try to run directory traversal attacks. Everything else can be solved by aliasing the sha256 value of the file name as real name.

imachug:
> Everything else can be solved by aliasing the sha256 value of the file name as real name.

Sorry, but I say strong **no**. ZeroNet is more of a transport protocol than a web browser. It should not decide how files are stored in the file system: for example, using sha256 instead of file names will break Git repositories hostings on ZeroNet.

AyrA:
> for example, using sha256 instead of file names will break Git repositories hostings on ZeroNet.

Git repository hosting is already completely broken because you don't allow stuff to start with a dot. Which means `.gitignore` and the entire `.git` folder are out. You also deny files to end in `-old` or `-new`. ZeroNet has many arbitrarily made up file name rules that don't matter. `content.json` is also completely blacklisted regardless if it's the actual page file or a file made up by the user in a subdirectory. In other words, if I create a git repository with a .gitignore and a content.json file in them I will publish an empty site because all important things were excluded.

AyrA:
A side effect of aliasing is also that you get free file deduplication with it.

imachug:
@AyrA What I was talking about is a *bare* git repository, like how GitHub hosts content. In other words, the content of your `.git` directory, not the code itself.

BTW, `.*` works for me.

AyrA:
@imachug It's normal for you that <code>.*</code> works. Invalid files always work on the machine that hosts the page.

The offending code is here:

https://github.com/HelloZeroNet/ZeroNet/blob/5a746769d01db58d54ff121e784fbf87374a89f3/src/Content/ContentManager.py#L622

AyrA:
I want to add here, that `..` in the middle of a file name is also considered invalid according to ZeroNet:

https://github.com/HelloZeroNet/ZeroNet/blob/5a746769d01db58d54ff121e784fbf87374a89f3/src/Content/ContentManager.py#L598

This looks like a beginners level check for a directory traversal attack.

Here's how to properly check for this:

1. Combine the base (trusted) path and the relative (user supplied) path using built-in filesystem library function (this gets a rooted but not necessarily resolved path string)
2. Use same filesystem library to convert the path into an absolute path (this resolves `..` and other constructs)
3. Check if the result still starts with the base path.

imachug:
We *do* know what path traversal is. But we also know how ZeroNet works. Using `/../` in `content.json` (that's what that ContentManager method does) is *incorrect*, because that would mean that `a.txt` and `dir/../a.txt` are the same file, and you can create two conflicting entries for a single file, which is kinda bad. I agree that the check is not awesome, however -- I'll make a PR right now.

As for the `.*` files, I believe there wasn't such a check before, @HelloZeroNet please check whether it's absolutely required (updating `ignore` in content.json sounds like a better solution).

imachug:
See #2160.

joshiemoore:
@HelloZeroNet I was wondering if there is any further action required on this PR, as it's been a few months since it was last updated.

HelloZeroNet:
I just adding new integration tests to make sure it work properly

PR #2160, opened 2019-08-19
Allow files with `..` as a name substring

imachug:


imachug:
@HelloZeroNet Any ideas why py3.8 can't be downloaded from time to time?

HelloZeroNet:
I have restarted the build and it works now, so it's something with travis. If it happens again we can try to switch no python nightly

imachug:
Sounds like a travis issue to me too. Switching to another CI (e.g. GitLab CI/CD) might work better (see my mail).

PR #2166, opened 2019-08-19
Allow blob: protocol

### LINKED ISSUE #2165: Add blob: scheme to Content-Security-Policy

  * ZeroNet version: 0.7.0 (4188)
  * Operating system: Linux
  * Web browser: Firefox 68.0
  * Tor status: enabled
  * Opened port: yes
  * Special configuration: none

ZeroNet includes the following `Content-Security-Policy` HTTP header:

`Content-Security-Policy: default-src 'none'; script-src 'nonce-xxxxxx'; img-src 'self'; style-src 'self' 'unsafe-inline'; connect-src *; frame-src 'self'`

`'self'` value allows `blob:` scheme to be used on Chrome but not on Firefox.
I'd like to ask adding `blob:` scheme where `'self'` is currently used.

#### Steps to reproduce:

  1. Open http://127.0.0.1:43110/1MVyVRFfQ8nSR3dMxisz1pZBwrd8jJRezy/ (TiddlyWiki)
  2. Press drop-down menu on the tiddler → export tiddler → CSV

#### Observed Results:

The tiddler export functionality, which uses `blob:` scheme, succeeds on Chrome but not on Firefox.

#### Expected Results:

`blob:` scheme could be used on both Chrome and Firefox.

imachug:
Fixes #2165

### DIFF
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -259,7 +259,7 @@ class UiRequest(object):
         if noscript:
             headers["Content-Security-Policy"] = "default-src 'none'; sandbox allow-top-navigation allow-forms; img-src 'self'; font-src 'self'; media-src 'self'; style-src 'self' 'unsafe-inline';"
         elif script_nonce and self.isScriptNonceSupported():
-            headers["Content-Security-Policy"] = "default-src 'none'; script-src 'nonce-{0}'; img-src 'self'; style-src 'self' 'unsafe-inline'; connect-src *; frame-src 'self'".format(script_nonce)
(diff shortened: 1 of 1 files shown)


ValdikSS:
I believe `blob:` should not be in quotes.

`headers["Content-Security-Policy"] = "default-src 'none'; script-src 'nonce-{0}'; img-src 'self' blob:; style-src 'self' 'unsafe-inline' blob:; connect-src *; frame-src 'self' blob:".format(script_nonce)`

That way it works in Firefox and Chrome.
This modification should be reviewed by a web security specialist since it may introduce unwanted functionality or a security vulnerability.

imachug:
Fixed quotes.

As for me, I don't see any possible security issues here: it's `blob` after all.

Issue #2208, opened 2019-09-27
Nginx reverse proxy：Invalid origin: https://domain.com


TheresaQWQ:
### Step 1: Please describe your environment

  * ZeroNet version: latest
  * Operating system: Ubuntu 18.04.3 LTS
  * Web browser: Chrome
  * Tor status: not available
  * Opened port: yes
  * Special configuration: ____

https is work, but websocket console display Invalid origin: https://domain.com

TheresaQWQ:
![Chrome console](https://user-images.githubusercontent.com/22370654/65755102-0686c780-e145-11e9-8476-3e8a4b91b063.png)

purplesyringa:
Have you added `https://domain.com` or `domain.com` to `ui_host` configuration key?

purplesyringa:
@TheresaQWQ Did that help?

ghost:
I would use Apache if I'm you. @imachug I'm sure that your suggestion not worked. Actually you need two server to proxy from clearnet to ZeroNet. One is where zeronet is running and you need to unset the headers which added by ZeroNet. This server should accept incoming connections only from the front-end server on port 80 and forward it to 43110. 

On the front-end you should redirect port 80 to 443 and proxy that to the back-end server xx.xx.xx.xx:80 here you can set headers! 

I think this issue related to that he proxy request to xx.xx.xx.xx:43110/raw/example.bit

In any case the back-end server should only download the site from your local machine where you update and sign everything over tor! Do not use the back-end to publish/update your sites!

purplesyringa:
Why would you need two servers? Can't you just make the ZeroNet instance listen on port 80 publically?

ghost:
@imachug @HelloZeroNet You need two server because one removes the headers added by ZeroNet and the other adding new ones. Not to mention that is safer! 

I currently proxy on the front-end example.org:443 to "backendip-xx.xx.xx.xx:80/raw/BitcoinAddress". -FIREWALL- The back-end listens on port 80 and forwards it to 43110. Works perfectly. No sandbox whatsoever... Caching however... you know...

ghost:
[...]

Issue #2169, opened 2019-08-20
Proposal: Reverse iframe


imachug:
## Abstract

Currently ZeroNet uses a wrapper for sidebar and notifications UI, and embeds site content with an iframe. This proposal fixes some issues caused by iframing by embedding ZeroNet UI into site content.

## Rationale

- #2165, #2057 - All features have to be marked unsafe by default, so supporting new ones is troublesome;
- #2154 - Most people don't understand what NOSANDBOX means;
- #2086 - Some interactive content in iframes might get blocked;
- #2058 - Pull-to-refresh is working weird with iframes;
- #2003 - "Opening as child window" is sometimes impossible to bypass;
- #1736 - Opening link in a new tab is more difficult than required;
- #1707, #565 - Embedding a site into a site is unsafe with an iframe;
- #1667 - Browser header style can only be set by root frame;
- #1403 - Scrolling is impossible, iframe is most likely the reason;
- #1399, #1054, #29 - localStorage, IndexedDB, etc. are not available in sandboxed iframes;
- #1237 - Fullscreen API doesn't work;
- #1236, #1262 - Some Vue plugins don't work;
- #1010 - Creating Web Workers requires data: URI usage;
- #469 - History management requires ZeroFrame usage.

## Overview

Current architecture:

```
┌─────────────────────────────────────────────────────┐
│ WRAPPER (src/Ui/template/wrapper.html)              │
│ Stores secrets (e.g. wrapper key)                   │
│ ┌─────────────────────────────────────────────────┐ │
│ │ WS UPLINK                                       │ │
│ │ UiWebsocket API                                 │ │
│ └─────────────────────────────────────────────────┘ │
│ ┌─────────────────────────────────────────────────┐ │
│ │ NOTIFICATIONS (<div>)                           │ │
│ │ ╔═════════════════════════════════════════════╗ │ │
│ │ ║ SITE NOTIFICATIONS (<div>)                  ║ │ │
│ │ ║ Unsafe content (possible XSS)               ║ │ │
│ │ ╚═════════════════════════════════════════════╝ │ │
│ │ ┌─────────────────────────────────────────────┐ │ │
│ │ │ WRAPPER NOTIFICATIONS (<div>)               │ │ │
│ │ │ Safe content (no leaking or spoofing)       │ │ │
│ │ └─────────────────────────────────────────────┘ │ │
│ └─────────────────────────────────────────────────┘ │
[...]

Issue #2087, opened 2019-07-09
The domain naming scheme proposal


geekless:
Some related comments and discussions:
* https://github.com/HelloZeroNet/ZeroNet/issues/83#issuecomment-94332474
* https://github.com/HelloZeroNet/ZeroNet/issues/83#issuecomment-489451550
* https://github.com/HelloZeroNet/ZeroNet/issues/1696
* https://github.com/HelloZeroNet/ZeroNet/issues/104
* https://github.com/HelloZeroNet/ZeroNet/issues/2049

The idea:
* ZeroNet site should be accessible with both `http://127.0.0.1:43110/...` and `http://.../` via a proxy.
* When accessing zites with a proxy, the pseudo-TLD `.zeronet` is used:
  `http://<address>.zeronet/`
* When accessing zites with `http://127.0.0.1:43110/`, the `.zeronet` part can be ommitted and is ignored by the core if present.
  These point to the same site:
  * `http://127.0.0.1:43110/<address>` - the classic way.
  * `http://127.0.0.1:43110/<address>.zeronet` - for similar look&feel to the proxied addresses.
* The second-level domains are handled by plugins.

* * *

The proposed second-level domain schemes:

* **Plain hashes**: `<hash>.zeronet`
  * Proxied: http://1HeLLo4uzjaLetFx6NH3PMwFP3qbRbTf3D.zeronet
  * Non-proxied: http://127.0.0.1:43110/1HeLLo4uzjaLetFx6NH3PMwFP3qbRbTf3D

* **Namecoin domains**: `.bit.zeronet`
  * Proxied: http://Talk.ZeroNetwork.bit.zeronet
  * Non-proxied: http://127.0.0.1:43110/Talk.ZeroNetwork.bit

* **Cert-based**, as proposed in https://github.com/HelloZeroNet/ZeroNet/issues/1696: `.user.zeronet`
  * Proxied: http://geekless.zeroid.user.zeronet, http://geekless.kxoid.user.zeronet
  * Non-proxied: http://127.0.0.1:43110/geekless.zeroid.user, http://127.0.0.1:43110/geekless.kxoid.user

* **Opennic-based** (https://github.com/HelloZeroNet/ZeroNet/issues/104): `.opennic.zeronet`
  * Proxied: http://example.libre.opennic.zeronet
  * Non-proxied: http://127.0.0.1:43110/example.libre.opennic

[...]

PR #2214, opened 2019-10-01
Add support for DNS

### LINKED ISSUE #104: Add DNS Support
Add DNS Support instead of only supporting Namecoin.

I suggest to use TXT records similar to the following example:

````
subdomain        IN      TXT    "zero=1Gfey7wVXXg1rxk751TBTxLJwhddDNfcdp"
````

### PR DESCRIPTION
**Pull request by filips123**
This adds support for classic DNS. It closes #104 and it is part of #2049.

---

The plugin uses [DNSPython](https://github.com/rthalley/dnspython) library for getting DNS responses and [DNSLink](https://github.com/filips123/DNSLinkPy) for resolving DNSLink records.

It uses a few DNS-over-HTTPS resolvers with [OpenNIC](https://www.opennic.org/) TLD support, but it can also be configured to use other resolvers or system configuration. Results are also cached in files so domains will still work if the computer is offline.

There are already a few (my) domains that use DNS plugin which you can use for testing:

* `py.zeroframe.oss`
* `js.zeroframe.oss`

---

Note that because now plugin has additional dependencies, I had to change the configuration for Travis CI, GitLab CI, Docker and Vagrant to also install dependencies from (enabled) plugins. @HelloZeroNet ZeroBundle also needs to be updated to do this.

Also, DNSPython currently doesn't support DNS-over-HTTPS so I created PR rthalley/dnspython#393 with support for it. Until it is not merged and released, we have to use my temporary branch. Then you would need to update it to use a newly created version.

I also think that there this domain plugins aren't currently working if default `Zeroname` is disabled. I already reported this in #2128, when it was partly fixed (before that it was completely impossible to create domain plugin), but it appears that [there are still problems](https://github.com/HelloZeroNet/ZeroNet/issues/2128#issuecomment-523638267) if `Zeroname` is disabled. I wasn't able to test this with this plugin, but it needs to be investigated.

There are currently also some problems with CI. Tests works, but there is problem with linter (`flake8`). Because DNSPython [needs to be loaded](https://github.com/HelloZeroNet/ZeroNet/issues/104#issuecomment-534674685) after Tor monkey-patching is completed, I needed to load them from a class method. I [implemented](https://github.com/filips123/ZeroNet/blob/dns/plugins/DNS/DNSResolver.py#L32-L44) that with importing modules inside a function and adding them to `globals()`. This works normally as modules are now availible at global context. However, `flake8` doesn't like that and throws `F821 undefined name` errors. Is there any way to tell `flake8` that `dns` and `dnslink` are actually existing names?

---

@HelloZeroNet @imachug @ValdikSS Can you test and rewiev this?

### CODE REVIEW

File: plugins/DNS/requirements.txt [hunk 1]
```diff
@@ -0,0 +1,2 @@
+dnspython @ git+https://github.com/filips123/dnspython@dns-over-https
```
purplesyringa: Please hard-code commit hash for security.

purplesyringa (changes requested):

plugins/DNS/DNSResolver.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if config.tor == 'always':
+            response = type('lamdbaobject', (object,), {})()
+            response.flags = dns.flags.TC
+
+            query = lambda *x, **y: response
+            dns.query.udp = query
+
+        globals()['dns'] = locals()['dns']
+        globals()['dnslink'] = locals()['dnslink']
```
purplesyringa: The `globals`/`locals` magic is not required here. Please consider using the following instead:

```python
def loadModule(self):
    global dns, dnslink
    import dns.resolver
    import dnslink

    if config.tor == 'always':
        ...
```

plugins/DNS/SiteManagerPlugin.py
```diff
@@ -0,0 +1,34 @@
...
+    @property
+    def dns_resolver(self):
+        if not self._dns_resolver:
+            nameservers = config.dns_nameservers
+            configure = config.dns_configure
+
+            self._dns_resolver = DNSResolver(
+                site_manager=self,
+                nameservers=nameservers,
+                configure=configure,
```
purplesyringa: Dangling comma.

plugins/DNS/DNSResolver.py
```diff
@@ -0,0 +1,103 @@
...
+
+            self.loaded = True
+
+    def loadModule(self):
+        import dns.resolver
+        import dnslink
+
+        if config.tor == 'always':
+            response = type('lamdbaobject', (object,), {})()
+            response.flags = dns.flags.TC
```
purplesyringa: Personally, I don't like the use of type dynamic construction here. I think it might defeat some optimizations, though they probably won't matter because this code is only executed once. Still, `type()` call looks like magic here because we can just use a class (*and* we don't even define the attribute via `type` call). My proposal:

```python
class Response:
    flags = dns.flags.TC
query = lambda *x, **y: Response()
dns.query.udp = query
```

plugins/DNS/DNSResolver.py
```diff
@@ -0,0 +1,103 @@
...
+            query = lambda *x, **y: response
+            dns.query.udp = query
+
+        globals()['dns'] = locals()['dns']
+        globals()['dnslink'] = locals()['dnslink']
+
+    def loadCache(self, path=os.path.join(config.data_dir, 'dns_cache.json')):
+        if os.path.isfile(path):
+            try: self.cache = json.load(open(path))
+            except json.decoder.JSONDecodeError: pass
```
purplesyringa: I think separating this into 4 lines would follow the code conventions we have better.

plugins/DNS/requirements.txt
filips123: @imachug Done!

plugins/DNS/DNSResolver.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def saveCache(self, path=os.path.join(config.data_dir, 'dns_cache.json')):
+        json.dump(self.cache, open(path, 'w'), indent=2)
```
purplesyringa: *Please* always close opened files, this breaks in PyPy.

purplesyringa (changes requested):

plugins/DNS/SiteManagerPlugin.py
filips123: @imachug Is this really needed to be fixed? Python allows trailing commas in lists and function calls.

Does ZeroNet follow any coding standard? Because according to [PEP 8](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas), trailing comma here should be correct.

plugins/DNS/DNSResolver.py
filips123: Fixed!

plugins/DNS/DNSResolver.py
filips123: Fixed!

plugins/DNS/DNSResolver.py
filips123: Fixed!

plugins/DNS/DNSResolver.py
filips123: Fixed!

plugins/DNS/SiteManagerPlugin.py
purplesyringa: ZeroNet doesn't really follow PEP 8. I think that following the existing code is better than following a standard. You can take a look at many other places in ZeroNet code where the dangling comma is not added. So even if it's not the correct way, it's at least the *same* way.

filips123: > I think that following the existing code is better than following a standard.

Yes, But I didn't check all ZeroNet code standards so I just used some things from PEP 8.

> You can take a look at many other places in ZeroNet code where the dangling comma is not added.

Should dangling comma only not be added in function calls, or also not in multi-line lists?

> So even if it's not the correct way, it's at least the same way.

I agree. I will remove this comma, but in the future, it would be good to reformat ZeroNet code to some known Python coding standard (PEP 8).

purplesyringa:
Congrats, the tests are passing now.

plugins/DNS/SiteManagerPlugin.py
purplesyringa: Actually, flake8 *is* using PEP 8 but some rules are ignored, so we're nearby.

purplesyringa: > Should dangling comma only not be added in function calls, or also not in multi-line lists?

I'm not sure about this one, let me check in a sec.

filips123: @imachug What should be used for multi-line lists? Should there be dangling comma or not?

purplesyringa: We don't use dangling commas at all. Just a few examples to support this:

```python
casubjects = [
    "/C=US/O=Amazon/OU=Server CA 1B/CN=Amazon",
    "/C=US/O=Let's Encrypt/CN=Let's Encrypt Authority X3",
    "/C=US/O=DigiCert Inc/OU=www.digicert.com/CN=DigiCert SHA2 High Assurance Server CA",
    "/C=GB/ST=Greater Manchester/L=Salford/O=COMODO CA Limited/CN=COMODO RSA Domain Validation Secure Server CA"
]
```

(and a few more from CryptConnection). Also:

```python
return self.render(
    "src/Ui/template/wrapper.html",
    server_url=server_url,
    inner_path=inner_path,
    file_url=re.escape(file_url),
    file_inner_path=re.escape(file_inner_path),
    address=site.address,
    title=html.escape(title),
    body_style=body_style,
    meta_tags=meta_tags,
    query_string=re.escape(inner_query_string),
    wrapper_key=site.settings["wrapper_key"],
    ajax_key=site.settings["ajax_key"],
    wrapper_nonce=wrapper_nonce,
    postmessage_nonce_security=postmessage_nonce_security,
    permissions=json.dumps(site.settings["permissions"]),
    show_loadingscreen=json.dumps(show_loadingscreen),
    sandbox_permissions=sandbox_permissions,
    rev=config.rev,
    lang=config.language,
    homepage=homepage,
    themeclass=themeclass,
    script_nonce=script_nonce
)
```

purplesyringa:
@filips123 Should this PR be merged if it doesn't work without Zeroname plugin?

filips123:
@imachug It should be. Zeroname is still the default domain system and almost no one would disable it. But I think that this issue should be fixed quickly.

But I actually only tested this few months ago with ENS plugin (which uses same structure as this). I will also try to test this again now.

**Update:** I tested this again and I can confirm that it works only when Zeroname is enabled. When Zeroname is disabled, the site will try to load (it will open wrapper and page title will be set), but the actual page will load forever. See debug log [here](https://pastebin.com/QFyUiExs) - the error is (almost) the same as in #2128. When Zeroname is enabled, it will work normally.

plugins/DNS/SiteManagerPlugin.py
filips123: Also fixed!

filips123:
@HelloZeroNet Do you know what could be a reason for that error?

HelloZeroNet:
I think this should be a separate plugin, because depending on the traditional dns system somehow defeat the decentralized and uncensorable goals of the project.

I will check why is it not working without zeroname

filips123:
@HelloZeroNet I don't think so. I would agree with you if my plugins would be the only way for domains in ZeroNet. But since users can also use Zeroname (or any other domain plugin), it would actually make it more decentralized as users won't be forced to use one particular domain system.

ghost:
Exactly! Multiple centralized solutions provides more decentralization than one centralized solution (and I would even say multiple solutions is better than one *period*). Of course, the best would be multiple decentralized solutions, but we have more work to get to that point.

filips123:
> Of course, the best would be multiple decentralized solutions, but we have more work to get to that point.

And it is actually not so easy to implement a really decentralized solution. Almost every third-party domain system for integration with ZeroNet (or any other project) would require either:

* Complete protocol implementation with locally stored data (which in case of blockchain-based systems means to have the whole blockchain stored locally).
* Data stored in a custom location (like Zeroname site).
* Data from a third-party provider (like Infura for Ethereum).

filips123:
@HelloZeroNet @shortcutme Did you find why this plugin doesn't work without Zeroname?

HelloZeroNet:
Yes, I pushed an update (Rev4238) that should make it work without the Zeroname plugin.

It happened because of the missing actionSiteMedia patching without the Zeroname plugin.
I have moved the domain resolving to the parsePath function and also added a short cache (10sec) for isDomain / resolveDomain function, because the parsePath function could be called multiple times for each http request.
https://github.com/HelloZeroNet/ZeroNet/commit/43a574225885c2d839e116b5a07ac0fb214048f9
https://github.com/HelloZeroNet/ZeroNet/commit/924a61309a351fa3e56309fcc75d6cb4f5e34c02

filips123:
@HelloZeroNet Should I change anything in this plugin or it will simply work?

From your commits, I saw that you removed `actionSiteMedia`. Does this mean that the whole `UiRequestPlugin` is not needed anymore?

ghost:
Hi @filips123 @HelloZeroNet 
Look https://github.com/HelloZeroNet/ZeroNet/issues/2208 I don't see how _dnslink is better than a reverse proxy to /raw/. Problematic is the forced headers with ZeroNet. At least when proxy and _dnslink detected to "raw" ZeroNet should remove the headers frame etc... Unnecessary. Anyway content.json is exposed to the world which may be a problem if ZeroNet signs the same file two times! If that's the case the private key can be recovered. Right now I stick to reverse proxy. As mentioned on the link above..

ghost:
@filips123 With reverse proxy I can use any TLD I want to host on ZeroNet. I don't really think this _dnslink will allow me to use for example: 

zeronet.example.org. IN CNAME zeronet.gateway.example.org
_dnslink.zeronet.example.org. IN TXT **dnslink=/raw/bitcoinaddress** 

That would be the goal no? The headers and sandbox in any case where /raw/ requested should be removed!

d14na:
@filips123 i've been thinking about how to bridge ENS to Zeronet for a while now. With the new ENS system and 3,4,5-letter domains being auctioned now, I think the system will gain traction within the crypto community. What's the status of #2164? Is this working? 

If so, then imo the next step is to bridge back to the clearnet for users NOT running the client. Cloudflare offers this for IPFS websites. Currently in the process of moving almost all of our front-end code over to IPFS. Honestly, it has its issues (mainly caching seems sketchy, but perhaps more seeding is needed), but overall its been working flawlessly for months. 

You can read about it here, if you're unfamiliar:
https://www.cloudflare.com/distributed-web-gateway/

The idea is, our proxy at 0net.io would serve (as it does now) the files to Clearnet users, however, they would be able to use `https://example.eth` instead of `https://0net.io/example.bit`. Obviously, this would NOT be limited to just `.eth`, I'm thinking `.xyz` too (they have a very progressive admin team).

While still making improvements to the proxy's underlying infrastructure, its realatively stable at this point https://status.0net.io. Let me know if you'd be interested in adding that support to your plugin. And, well, how you think we'd go about implementing that into the current proxy's code.

Regardless, good luck with this effort. Don't hesitate to reach for help testing, debugging, whatever. cheers!

ghost:
@d14na This has nothing to do with dnslink on Zeronet far as I read your comment. Lastly forget 0net.io because that is a big trash. If you want to proxy clearnet to ZeroNet you will need your own servers for front and back-end!!!!!

purplesyringa:
A private key can only be recovered if you sign something twice using the same random number, which is next to impossible

filips123:
> Look #2208 I don't see how _dnslink is better than a reverse proxy to /raw/.

DNSLink is actually solves a bit different problem than reverse proxy. Main goal of DNS plugin is to also allow users to use normal DNS instead of Namecoin. However, this can also be expended to reverse proxy later. See other existing issues in this repository about DNS and proxies (I don't have currently time to link them all).

> i've been thinking about how to bridge ENS to Zeronet for a while now. With the new ENS system and 3,4,5-letter domains being auctioned now, I think the system will gain traction within the crypto community. What's the status of #2164? Is this working?

It should be working. Also please test it for any issues.

However, there are still few small known issues with it. See PR comments for details.

Main problem is that there is no easy way to create actual ENS domains that link to ZeroNet sites. This will be fixed when my PR to ENS manager will be merged and released. Until then you have to manually create ENS domain in programmatic way using Web3 library.

Other problem is that Web3.py has not yet merged my changes for ContentHash records which are needed for ZeroNet. Until then, you have to install Web3.py from my temporary branch.

> The idea is, our proxy at 0net.io would serve (as it does now) the files to Clearnet users, however, they would be able to use https://example.eth instead of https://0net.io/example.bit. Obviously, this would NOT be limited to just .eth, I'm thinking .xyz too (they have a very progressive admin team).

This is also already known idea 😁 See existing issues which are related to domains and proxies.

This could be achieved  with CNAME/A/AAAA record as well as DNSLink record so domain system would then forward request to ZeroNet instance. This would require my DNS plugin as well as few other ZeroNet changes that are not made yet (to allow accessing ZeroNet via custom domain).

One other feature is to also point CNAME/A/AAAA record to local ZeroNet instance so users will be able to access it using their own ZeroNet instance.

filips123:
@HelloZeroNet @shortcutme Can you check my previous questions:

> From your commits, I saw that you removed `actionSiteMedia`. Does this mean that the whole `UiRequestPlugin` is not needed anymore?

Also https://github.com/HelloZeroNet/ZeroNet/pull/2214#issuecomment-538234536, https://github.com/HelloZeroNet/ZeroNet/pull/2214#issuecomment-538526212 and https://github.com/HelloZeroNet/ZeroNet/pull/2214#issuecomment-538528158. So, when this is resolved, I think that this should be merged even if it is using centralized DNS.

d14na:
> Lastly forget 0net.io because that is a big trash. 

could you please explain? constructive criticism/feedback will help improve its usability. Or is this about NOT encouraging the use of proxies in general? imo, proxies have always help with on-boarding new (curious) users, same applies to TOR proxies.

> If you want to proxy clearnet to ZeroNet you will need your own servers for front and back-end!!!!!

I don't know what you mean by **front and back-end**. Assuming 0net.io if the "front-end", then what "back-end" server are you saying is required?

d14na:
> It should be working. Also please test it for any issues.

Okay great, I will. I think it would also help me better understand how the plugins work. I have 2 that I'm working/experimenting on, based on @rllola tutorial, but now they're broken on Py3.

> Main problem is that there is no easy way to create actual ENS domains that link to ZeroNet sites.

To be clear, are you talking about "registering" the ENS domains or "linking" them to zites? I've done many registrations, and I would believe "registration" to way beyond the scope of your plugin.

> Until then you have to manually create ENS domain in programmatic way using Web3 library.

Why wouldn't you just use the dapp at https://app.ens.domains? I'm not clear what you're trying to accomplish.

> Until then, you have to install Web3.py from my temporary branch.

that's fine.

> This is also already known idea 😁 See existing issues which are related to domains and proxies.

ok

> This would require my DNS plugin as well as few other ZeroNet changes that are not made yet (to allow accessing ZeroNet via custom domain).

yeah, this is where I see the challenge; but I'll start by working with your plugin and see if things don't make better sense by this point.

> One other feature is to also point CNAME/A/AAAA record to local ZeroNet instance so users will be able to access it using their own ZeroNet instance.

that sounds good

ghost:
> > Lastly forget 0net.io because that is a big trash.
> 
> could you please explain? constructive criticism/feedback will help improve its usability. Or is this about NOT encouraging the use of proxies in general? imo, proxies have always help with on-boarding new (curious) users, same applies to TOR proxies.
> 
> > If you want to proxy clearnet to ZeroNet you will need your own servers for front and back-end!!!!!
> 
> I don't know what you mean by **front and back-end**. Assuming 0net.io if the "front-end", then what "back-end" server are you saying is required?

You just keep using your bridge... I still say 0net.io is a trash! You don't understand **front and back-end**? What you don't understand? You keep talking about your trash 0net.io but you don't understand what the hell is that...

purplesyringa:
@BugsAllOverMe There is no such thing as frontend and backend servers in decentralized networks. If you think we misunderstood you, please give us an example.

ghost:
@imachug Certainly somehow need to forward traffic from clearnet to ZeroNet and the only way you can do it is to have a bunch of back-end server (cluster) with ZeroNet running on them than over firewall from the front-end (where you host example.org) you can proxy all requests to ZeroNet. That is in fact the goal of ZeroNet. If you assume people will download ZeroNet just to access websites hosted on it than you are wrong! Nobody will download it. Specially not when there is a few people on GitHub who controls the entire development of ZeroNet. Pretty centralized if you ask me. There is only two way to access ZeroNet sites. One is to download it (but than stupid headers and frame will pretty much lock developers into a sandbox) or you run many servers as I noted above. The later will allow you to strip ZeroNet from frame and unnecessary wrong headers and add your own.

purplesyringa:
> Certainly somehow need to forward traffic from clearnet to ZeroNet and the only way you can do it is to have a bunch of back-end server (cluster) with ZeroNet running on them than over firewall from the front-end (where you host example.org) you can proxy all requests to ZeroNet.

Got it. In your case, ZeroNet is backend and nginx is frontend, right? Actually, this works well if you host ZeroNet *as well as* other sites, but if you only want to host a ZeroNet proxy, you don't need nginx. Actually, you probably don't want nginx to modify ZeroNet traffic though.

> If you assume people will download ZeroNet just to access websites hosted on it than you are wrong!

Well, uh, they do. And that's not GitHub where they find it.

> but than stupid headers and frame will pretty much lock developers into a sandbox

They are **not** stupid. That's for sandboxing and making users safe. Escaping the sandbox might result in *big* problems. Please don't try to remove additional security layers if you don't replace them somehow.

filips123:
> To be clear, are you talking about "registering" the ENS domains or "linking" them to zites?

I'm talking about linking domains to ZeroNet. Plugin uses content hash record to store hire address, but currently there is no easy way to create it for ZeroNet.

> Why wouldn't you just use the dapp at https://app.ens.domains? I'm not clear what you're trying to accomplish.

My point is that https://app.ens.domains currently does not support linking domains to zites. I already created PR ensdomains/ui#23 to add support for this, but it has not been yet merged.

hacktivist:
@imachug Actually I run Apache. The sandbox don't need when you proxy requests. The sandbox only usable when someone actually has ZeroNet installed on the local machine. If you don't escape the frame while you proxy clearnet to ZeroNet than your website becomes unreadable by search engines. Most headers simply don't need when you proxy request to https://127.0.0.1:43110/raw/bitcoinaddress. I use Content-Security-Policy: upgrade-insecure-requests;base-uri 'self';default-src 'self'; and not ZeroNet's default.

hacktivist:
@filips123 You can proxy any TLD to ZeroNet!

filips123:
@HelloZeroNet @shortcutme Based on 5ca3401eb992dd52fa2a784de03d64b886246640, I assume that `UiRequestPlugin` needs to be removed. I also removed it here so this PR is now completely ready.

HelloZeroNet:
Yes you are right, UiRequestPlugin can be removed and I did based on your recommendation

d14na:
> I'm talking about linking domains to ZeroNet. Plugin uses content hash record to store hire address, but currently there is no easy way to create it for ZeroNet.

ok, i just wasn't clear

> I already created PR ensdomains/ui#23

if i understand this PR correctly: 
1. your updating `content-hash` to v2.4.4 (i'm sure for a very good reason).
2. it seems you're adding `zeronet://` as a "supportedCodec". 

I have no issue with that, but why not ALSO support for `zero://`? Are you only allowed one? it seemed to me that the community was in greater support of the latter.

filips123:
@D14na

> your updating content-hash to v2.4.4

I updated `content-hash` because was needed to be updated to support `zeronet` codec.

ENS uses `content-hash` (EIP 1577) to differ between multiple content hash types (IPFS, Swarm, Tor, ZeroNet...). And `content-hash` was updated to support `zeronet` codec.

> I have no issue with that, but why not ALSO support for `zero://`?

Because `zero://` and `zeronet:/` are different protocols/schemes. `zero://` is used for internal ZeroNet communication between trackers. `zeronet://` is used as an identifier to let know ENS system that this is ZeroNet domain. See https://github.com/HelloZeroNet/ZeroNet/issues/2049#issuecomment-513975113 for details.

It doesn't matter which is "more popular". They are used for different things. The community would have to update things in any way if they want to use DNS.

For the same reason, I also use `dnslink=/zeronet/ADDRESS` instead of `dnslink=/zero/ADDRESS`.

purplesyringa:
@filips123 Whilst `zeronet` was recommended by @HelloZeroNet, there was later some argument on what protocol to use, namely it was brought up in #2169 and then moved to #2087. We were thinking about several ways:

- Using `zhttp`/`zhttps` protocol;
- Using `zero`/`zeronet` protocol;
- Using `.zero`/`.zeronet` TLD.

Personally, I think that the first solution makes most sense from the size of the protocol itself -- ZeroNet does not actually support HTTP or websockets or whatever; it just *simulates* some features. But then we had some problems with implementing that so the issue is now stale.

Thus, I guess we'll have to admit that the ENS protocol name has no connection to how end users communicate with ZeroNet: think of it as a codec or a type or something like that.

filips123:
@imachug Yes, deciding which way should be used for accessing ZeroNet is another long story... But this is not related to ENS or DNS codecs that are used in my plugins.

> Thus, I guess we'll have to admit that the ENS protocol name has no connection to how end users communicate with ZeroNet: think of it as a codec or a type or something like that.

Also yes, here `zeronet://` actually is codec/type for ENS/DNSLink protocol name.

Also, because `zero://` (as a communication/transport protocol) may also be added to `multicodec` later in the future (in case if ZeroNet gets support for more protocols), it is better to separate `zero://` (as a communication/transport protocol) and `zeronet://` as a client/service name in ENS. This is actually what @HelloZeroNet said.

d14na:
> zeronet:// is used as an identifier to let know ENS system that this is ZeroNet domain.

oh, so this is for "internal" / "system" -use"?

> Yes, deciding which way should be used for accessing ZeroNet is another long story..

yeah, that's what I thought it was for, but i got it now! think i'm mostly caught up, just need to make time to go through the code now.

filips123:
> oh, so this is for "internal" / "system" -use"?

Yes, it is only to tell DNS/ENS that record should be handled by ZeroNet.

d14na:
@filips123 do you have separate repo for your plugin? meaning, can it be installed with the plugin manager? what i've seen appears "built-in" to core.

filips123:
@d14na No, because it was meant to be built-in to core. If you want to use it, download my ZeroNet repository and copy DNS plugin to your installation.

d14na:
> download my ZeroNet repository and copy DNS plugin to your installation

`git clone https://github.com/filips123/ZeroNet.git ZeroNet-filips123`, but I don't see the plugin. what's the path? or do I need to change branches?

__EDIT: ok, i changed the branch to `DNS`__

d14na:
@filips123 just so I'm clear, what's the reason for having this "built-in"? I see 2 requirements: `dnspython` (which is your patched) and `dnslink`. Is there a way to "auto-install" these requirements with the the Plugin Manager? if not, how problematic is the process for the user? would it be possible to run the adt'l dependencies in a container? (eg venv?)

i'm sure I'll have more questions, but I'll see if I can get it working and hopefully offer some useful feedback first

filips123:
@D14na

> just so I'm clear, what's the reason for having this "built-in"?

Domain plugin that needs to be installed separately is kinda useless. If your site uses it, you will still have to provide classic address-based URL for users that don't have it. And the main point of domain systems is that you don't have to provide addresses.

> Is there a way to "auto-install" these requirements with the the Plugin Manager? if not, how problematic is the process for the user?

Currently no.

In CI, requirements from the plugin are also installed. In local installations, they need to be installed with the manual calling of plugin's `requirements.txt`. I might also add this to README.

You can use this to install all plugins' dependencies:

```
for PLUGIN in $(ls plugins/[^disabled-]*/requirements.txt); do pip install --upgrade -r ${PLUGIN}; done
```

In ZeroBundle, that requirements would need to be installed directly. This is currently not used, but it needs to be added after this plugin is released.

> would it be possible to run the adt'l dependencies in a container? (eg venv?)

What do you mean with this?

Installing plugin's dependencies is same as installing normal dependencies from root `requirements.txt`. You just need to install additional `requirements.txt` files from plugins.

d14na:
> Domain plugin that needs to be installed separately is kinda useless.

ahh, of course, i wasn't thinking. regardless, a separate repo makes it easier for others to test, and that's probably the best path to core adoption.

> What do you mean with this?

it doesn't matter, I'm just working with a lot with docker now, so everything is a container to me

filips123:
> ahh, of course, i wasn't thinking. regardless, a separate repo makes it easier for others to test, and that's probably the best path to core adoption.

Still, it is meant to be included in core as soon as possible. That's why it doesn't have a separate repository.

> it doesn't matter, I'm just working with a lot with docker now, so everything is a container to me

You can use provided ZeroNet's `Dockerfile`. I updated it to also install dependencies from plugins.

hacktivist:
Hello! 

I going to have my own version of ZeroNet. I already made significant changes in the headers it sends, MIME types and enabled caching. I see here a conversation which indicates that some people pushing changes which are insecure solely for they own profit. @filips123 You attacked here a person who talked about how to use ZeroNet as back-end to a reverse proxy and yet here you are [disabled sandbox in local](https://github.com/HelloZeroNet/ZeroNet/pull/2238) and pushed a DNS plugin to be included in ZeroNet core. I think need to review very carefully all code you are pushing or propose. I personally don't trust you.

ValdikSS:
I haven't tried to run the plugin yet, only checked the code.

1. The plugin uses third-party DoH resolvers by default. I don't think this is a good idea. In my opinion, it should use system resolver by default unless configured otherwise or Tor always mode enabled.
2. `isDomain` regex doesn't allow international domain names. Or are they converted to punycode?

filips123:
@ValdikSS

> The plugin uses third-party DoH resolvers by default. I don't think this is a good idea. In my opinion, it should use system resolver by default unless configured otherwise or Tor always mode enabled.

Problem is that default system resolver will most likely use Google or ISP nameservers. And Google is not a good solution for any privacy/decentralized project and ISPs are also inspecting traffic in some countries.

Another problem is that normal DNS traffic is unencrypted so everyone can inspect/modify it. So I'm using DoH which is encrypted.

Another reasons why I'm using this specific nameservers is that they support OpenNIC TLDs. OpenNIC is alternative domain root and has free domains (but you can still use normal domains). Free alternative to ZeroNet addresses is also something that users would want. Default system resolvers most likely won't support it.

But you still can enable default system resolver or use custom nameservers if you want.

> `isDomain` regex doesn't allow international domain names. Or are they converted to punycode?

I actually wasn't thinking about this. Can you check yourself if that domains work?

d14na:
> Still, it is meant to be included in core as soon as possible. 

i'd like to work a-bit on the code, without cloning the core repo. i can use a private repo, i just thought a repo from you on github would be better (and official, until the core merge)

_i don't really know how things work around here, but my advice is to start with a separate repo, for testing amongst the community, once there's enough support, then I'm sure it will naturally find its way into the core lib_

> You can use provided ZeroNet's `Dockerfile`. I updated it to also install dependencies from plugins.

nice. i've been using virtualbox for the longest, but now everything is docker, its sooo much lighter on the host. actually what i was looking for was a docker-like solution for python, which I just got working (__venv__); that way i can install your fork without polluting the os.

> There are already a few (my) domains that use DNS plugin which you can use for testing:
> 
> * `py.zeroframe.oss`
> * `js.zeroframe.oss`

__So the plugin is working, here's the output:__

`py.zeroframe.oss` resolves to `http://127.0.0.1:43110/1GitLiXB6t5r8vuU2zC6a8GYj9ME6HMQ4t/repo/?178UMrEt8vxf5Z6LPxkrcWFPXPfD9XNNH7`

```
[05:22:46] DNSPlugin resolver: py.zeroframe.oss -> 178UMrEt8vxf5Z6LPxkrcWFPXPfD9XNNH7
[05:22:48] Site:1GitLi..MQ4t zeroframepy.git/packed-refs fileGet error: FileNotFoundError: [Errno 2] No such file or directory: './data/178UMrEt8vxf5Z6LPxkrcWFPXPfD9XNNH7/zeroframepy.git/packed-refs' in UiWebsocket.py line 687 > FilePack/FilePackPlugin.py line 195 > SiteStorage.py line 223
```

---

`js.zeroframe.oss` resolves to `http://127.0.0.1:43110/1GitLiXB6t5r8vuU2zC6a8GYj9ME6HMQ4t/repo/?1HJS6quLaqNLYxHVM9nL2rfmAsRvumqC1H`

```
[05:25:06] DNSPlugin resolver: js.zeroframe.oss -> 1HJS6quLaqNLYxHVM9nL2rfmAsRvumqC1H
[05:25:09] Site:1GitLi..MQ4t zeroframejs.git/packed-refs fileGet error: FileNotFoundError: [Errno 2] No such file or directory: './data/1HJS6quLaqNLYxHVM9nL2rfmAsRvumqC1H/zeroframejs.git/packed-refs' in UiWebsocket.py line 687 > FilePack/FilePackPlugin.py line 195 > SiteStorage.py line 223
```

__what's next?__

filips123:
@d14na I can provide you ZIP file with the plugin.

d14na:
> I can provide you ZIP file with the plugin.

its already working! i guess i was just looking for more zites to test out.

---

__so 2 things:__

1. how do I register names like yours? where's the registrar? if its NOT on zeronet, then how do we get it on?

2. my endgame is `.eth` support (as mentioned #2049), so I'm highly motivated to contribute towards any development leading to that end. if you could point me in the __"direction"__ of what it would take to start integrating `.eth`, that would be much appreciated.

filips123:
> how do I register names like yours? where's the registrar? if its NOT on zeronet, then how do we get it on?

You can register normal DNS domains in your DNS manager. That `.oss` domains from my example are OpenNIC and can be registered [here](https://www.opennic.org/).

The exact process depends on your DNS manager, but in summary you need to create `_dnslink.your.domain` with TXT record `dnslink=/zeronet/YOUR-ZERONET-ADDRESS`. Then wait some time so DNS records are updated and your site should be accessed from `http://127.0.0.1:43110/your.domain`.

For ENS domains, you will be able to register them with ENS Manager website. You will need to create content hash record with value `zeronet://YOUR-ADDRESS`. Note that this won't work until ensdomains/ui#23 is merged.

> my endgame is .eth support (as mentioned #2049)

I also created plugin for ENS support in #2164. It should work, but please check that PR for some details and some issues about it.

d14na:
> That `.oss` domains from my example are OpenNIC and can be registered [here](https://www.opennic.org/).

ok. don't know much about opennic or really dns in general, so i've got some reading to do.

and what about registering `.zeronet` with opennic? has anyone done that? is there reason to do that? also, i see `.o` is registered, what about registering `.0`? is that legal as per dns spec? does it matter, if its on opennic? regardless, the TLD registration states:

> You MUST have operated a Tier 2 DNS server continuously for at least 3 months

i feel like i'd want to do this asap, just so that in 3 mos it could be available as an option. what do you think?

---

> with TXT record `dnslink=/zeronet/YOUR-ZERONET-ADDRESS`

is this what YOUR plugin handles? or is this done at opennic?

> create content hash record with value `zeronet://YOUR-ADDRESS`

have a ton of names, but have never actually done this. something to try out this weekend

> Note that this won't work until [ensdomains/ui#23](https://github.com/ensdomains/ui/pull/23) is merged.

just from examining this [commit](https://github.com/ensdomains/ui/pull/23/commits/9d79a20af509fcda2c61e83f703daa27d26db3d0), i'm wondering why line # 52 already has `zeronet` in the regex, was zeronet previously supported in some way?

and then this [commit](https://github.com/ensdomains/ui/pull/23/commits/3fd9a1d6c60692f0ad60ad2e70c24d5df136ba4c) is adding 3 new dependencies. what are they for? imo, the less changes, the faster the merge, just something to consider

> I also created plugin for ENS support in #2164.

why can't you import your ens support into THIS plugin? do you think its better to have both? i believe you sent me a .zip of ENS (can't remember how i got it), so I'll start reviewing that code shortly. i need to spend a bit more time with all this dns code (to get comfortable), then i'll be ready to hit the slack / gitter / whatever channels to help get that PR merged asap.

thank you for making all of this. let me know what help i can offer.

filips123:
@d14na

> and what about registering `.zeronet` with opennic? has anyone done that? is there reason to do that? also, i see `.o` is registered, what about registering `.0`? is that legal as per dns spec? does it matter, if its on opennic?

There isn't any need to do this.

Unline normal domains, `.zeronet` TLD (which may be used to access ZeroNet in the future) is only special-usage pseudo-TLD. It's only use-case is to clearly separate requests that should be made to ZeroNet from clearnet requests/websites. But the actual domain resolving is based on a first subdomain (`example.bit.zeronet` would resolve `example.bit` using Namecoin).

However, it would be good to register it to IANA as a special-use domain to prevent other companies from registering it themselves and making a conflict with ZeroNet.

> is this what YOUR plugin handles? or is this done at opennic?

You need to do this at OpenNIC. For `.oss` domains, you can use [this](http://be.libre/) (requires you that you have OpenNIC nameservers used) and for `.o` domains you can do this [from GitHub](https://github.com/moderntld/.o/wiki/Claiming-Your-Domain). For other domains, you need to do this in your DNS manager.

The plugin then checks if the domain has that TXT record and resolves it to address.

> have a ton of names, but have never actually done this. something to try out this weekend

*(This is for Ethereum's ENS and is not actually related to this plugin)*

Unfortunately, you can't use ENS Manager website until ensdomains/ui#23 is merged. You can try to build that website yourself (from a branch of that PR) or use Web3.py (also from my temporary branch) to programmatically set content hash record.

> just from examining this commit, i'm wondering why line # 52 already has zeronet in the regex, was zeronet previously supported in some way?

This is just merge commit and this is how GitHub displays them. No, support for ZeroNet was not available before. [This](https://github.com/ensdomains/ui/pull/23/files) is complete diff.

> and then this commit is adding 3 new dependencies. what are they for? imo, the less changes, the faster the merge, just something to consider

Also merge commit. Those dependencies were added to the original ENS Manager repository.

> why can't you import your ens support into THIS plugin? do you think its better to have both?

I think that each domain system should have a separate plugin. This would make them more transparent, easier for development/debugging and also allow users to manage each one separately.

This specific plugin is made specifically for classic/normal DNS.

> i believe you sent me a .zip of ENS (can't remember how i got it)

I didn't send you ZIP file as I forgot to do this (there was actually some old ZIP of it but it is few months old)... But you have source code [here on GitHub](https://github.com/filips123/ZeroNet/tree/ethereum-name-service/plugins/ENS).

If you want to directly get ZIP files, they are available here:

* [DNS plugin](https://github.com/HelloZeroNet/ZeroNet/files/3773847/DNS.zip)
* [ENS plugin](https://github.com/HelloZeroNet/ZeroNet/files/3773848/ENS.zip)

d14na:
> But the actual domain resolving is based on a first subdomain (`example.bit.zeronet` would resolve `example.bit` using Namecoin).

got it!

> The plugin then checks if the domain has that TXT record and resolves it to address.

got it!

> Unfortunately, you can't use ENS Manager website until [ensdomains/ui#23](https://github.com/ensdomains/ui/pull/23) is merged. You can try to build that website yourself (from a branch of that PR) or use Web3.py (also from my temporary branch) to programmatically set content hash record.

so you're saying this CAN be achieved WITHOUT necessarily relying on the merger? i know that is of course the goal, but if there is a way to test this now with actual ENS domains, i'd try to make that happen asap.

not afraid of docs, but if you could provide an example of the __CORRECT__ CLI CMD _"to programmatically set content hash record"_ that would likely speed up my trial-and-error process.

> [This](https://github.com/ensdomains/ui/pull/23/files) is complete diff.

oh i see, i was on the wrong tab. that looks really clean, no issues there

> I think that each domain system should have a separate plugin.

ok

> This would make them more transparent, easier for development/debugging

absolutely!

> allow users to manage each one separately

imo, that's NOT a benefit to the user

> But you have source code [here on GitHub](https://github.com/filips123/ZeroNet/tree/ethereum-name-service/plugins/ENS).

great! learning a great deal about Zeronet internals just from these plugins. 

thanks again!

filips123:
## How to register ZeroNet DNS domain

### OpenNIC .libre/.oss domain

1. Set up your system DNS resolver to use [OpenNIC](https://www.opennic.org/) DNS servers. This is required only for the process of registration.

2. Open [domain registrator](http://be.libre/).

3. Register a new domain if you haven't it already.

4. Create TXT record with content `dnslink=/zeronet/YOUR-SITE-ADDRESS`.

5. Save the record and wait until DNS servers are updated.

6. Your website should be accessible from `http://127.0.0.1:43110/your-domain.oss`.

### Other OpenNIC domains

1. Set up your system DNS resolver to use [OpenNIC](https://www.opennic.org/) DNS servers. This is required only for the process of registration.

2. Follow the registration process of your desired registrators and add the same TXT record. All registrators are listed [here](https://www.opennic.org/).

3. Your website should be accessible from `http://127.0.0.1:43110/your-domain.tld`.

### Other domains

1. Use your registrator/DNS manager website to create TXT record.

2. Your website should be accessible from `http://127.0.0.1:43110/your-domain.tld`.

filips123:
@d14na I added description for registering DNS domain here and steps for registering ENS domain in ENS plugin PR.

filips123:
@HelloZeroNet @shortcutme Can you merge this?

As discussed in https://github.com/HelloZeroNet/ZeroNet/pull/2214#issuecomment-538234536, https://github.com/HelloZeroNet/ZeroNet/pull/2214#issuecomment-538526212 and https://github.com/HelloZeroNet/ZeroNet/pull/2214#issuecomment-538528158, this won't defeat the decentralized goals of ZeroNet as this won't be the only supported domain system. Users would still be able to use Namecoin or any other domain system for their sites.

Also, the domain plugin that needs to be installed separately is kinda useless. If your site uses it, you will still have to provide classic address-based URL for users that don't have it. And the main point of domain systems is that you don't have to provide addresses.

This domain plugin would also enable accessing ZeroNet sites as normal websites. This means that if site owner would point A record of domain to IP of public or local ZeroNet instance, the site will be accessible just like normal website. Although this is not currently supported, it is something that I and @imachug are currently working on.

And in the future, we could also find other DoH resolvers with support for OpenNIC and include them here. This would make resolving more decentralized as there would be multiple resolvers. And maybe we could then also chose random resolver from a list for each user.

**Update:** ~~There is some small things that needs to be added before this can be merged. I will do it later today.~~ That thing were dependencies. This is now fixed.

ValdikSS:
@filips123, I'll test the plugin today to check how good does it work.

filips123:
@HelloZeroNet @imachug @krixano There is one last thing that needs to be done before this PR can be merged. Problems are additional plugin dependencies.

---

The plugin includes additional dependencies only for it. But when updating ZeroNet from UI, that dependencies will not be installed so the plugin will not load.

One of the solutions for this is to manually bundle plugin dependencies along with plugin code here in the repository. But this is not the best solution as it would pollute repository with third-party dependencies and make things harder to maintain.

Another solution is to install dependencies on its first run. This won't require to include dependencies to repository, but still install them so the plugin will be usable. Here is the problem that this probably doesn't work with ZeroBundle as dependencies in ZeroBundle are stored in the compiled file. But we can still install dependencies to `src/lib` directory which is not compiled with ZeroBundle. But there is still a problem here because we would need to access and use `pip` programmatically (or make some lightweight PyPI downloader).

The final solution is to install dependencies directly into ZeroBundle. But this means that all users would have to manually reinstall ZeroBundle in order to update ZeroNet.

---

I would go for the second solution (install dependencies to `src/lib` with some lightweight PyPI downloader on first plugin run). But I would also install dependencies to ZeroBundle for all new users because it is better to have dependencies pre-installed.

HelloZeroNet:
I don't want to depend on pip, so I recommend to create plugins that does not have any external dependencies or include them with the plugin.

ghost:
You can require that dependencies be put into the plugins manifest or content.json or whatever the file is - the plugin settings file. Then when a person goes to install the plugin, the dependencies are listed and the user is prompted to install them from within the client, and the client installs them.

d14na:
> You can require that dependencies be put into the plugins manifest or content.json or whatever the file is - the plugin settings file. Then when a person goes to install the plugin, the dependencies are listed and the user is prompted to install them from within the client, and the client installs them.

does this currently exist? specifically...

> the user is prompted to install them from within the client

how is the (dependencies) installation triggered by the client? and if it cannot be, i believe that is what @filips123 is alluding to being added

krixano:
> and if it cannot be, i believe that is what @filips123 is alluding to being added

Yes, it **is** what filips is saying. I already Know that!

But I expanded on his idea by proposing dependencies be put into the plugin's settings file and adding a prompt to install dependencies, which filips did *not* suggest.

About installing dependencies from the clients, I would think it's possible by just simply starting a new process.

purplesyringa:
What if we put the dependencies to src/lib/ for now and switch to a decetralized PyPI afterwards?

filips123:
@HelloZeroNet I bundled all plugin's dependencies (DNSPython and DNSLink) into ZIP file along with plugin's source code. That ZIP file is then added to Python's path (Python can load modules from ZIP) in case if dependencies are not already installed.

I also included support for DNS-reverse-proxy (implemented by @imachug) which enables you to create a site which is hosted on ZeroNet and access it like normal website. To do this, you need to create DNSLink TXT record that points to site address and CNAME/A/AAAA that points to ZeroNet instance. In case of local ZeroNet instance, the site can be accessed at `example.com:43110`. In case of public ZeroNet proxy, site can be accessed directly at `example.com` (because most proxies host ZeroNet on port 80). Note that some sites (like ZeroHello, probably also other official sites with dynamic links to other sites) are not yet ready for this as they make incorrect links to other sites. This needs to be fixed in their end.

filips123:
@HelloZeroNet @shortcutme Can you check and merge this?

Dependencies are now bundles with source code (and all licenses are compatible) so there shouldn't be any problem with using this plugin.

This would allow accessing ZeroNet sites from classic DNS domains which is useful in case if site author already has domain (or want to use free OpenNIC domain) and don't want to register another Namecoin domain.

And it would also allow accessing ZeroNet as reverse-proxy. So if the domain has CNAME/A/AAAA record that points to ZeroNet instance and TXT record which contains the address of the site, the site will be directly accessible from that domain.

In case of local instances, this will have to include port `43110`. Solution for this would be to use `127.43.11.0:80` as default host of ZeroNet UI. This will also enable viewing ZeroNet sites with local instances from on port 80 while still keeping ZeroNet separated from other services which may run on port 80 (as it would use `127.43.11.0` which is another loopback IP). IPs in `127.*.*.*` should resolve to loopback in most cases, but I think that in macOS, the only loopback address is `127.0.0.1`. This means that for it, we would need to manually assign new IP to loopback interface, which probably requires root permissions.

But this is actually not so important for this PR so it would be great if you can merge it.

And even if classic DNS is centralized, adding it as additional, but enabled by default, domain system won't defeat decentralized goals of ZeroNet, as users will actually have more choice between domain systems. And with ENS support, it would make it even more decentralized.

HelloZeroNet:
Ok, I'm not against it, I will do some tests later on it.

filips123:
My PR for DNS-over-HTTPS support has been merged, so we don't need to depend on my fork anymore. However, we still need to use Git-based version, as changes still aren't released in new version.

filips123:
@imachug I tested DNS reverse proxy with my public ZeroNet proxy and problem is that sometimes cookies are not sent to WebSocket connection which means that connection fails with Multiuser plugin. Do you think it would be possible to send that cookie to WebSocket connection even if it is in a different origin than the website?

Also, zero button links to `zero/1HeLLo4uzjaLetFx6NH3PMwFP3qbRbTf3D` but this can probably be fixed easily with some condition.. There might also be some other problems with other features that try to connect to wrong URL.

Also, can you check if there can be any possible security problems with accessing WebSocket from a different origin and such?

purplesyringa:
> Do you think it would be possible to send that cookie to WebSocket connection even if it is in a different origin than the website?

AFAIK there's no way to add a cookie to another site. The only solution I've managed to come up with is sending the cookie via the websocket, say, in the first message. @HelloZeroNet Do you think that might work?

purplesyringa:
> Also, zero button links to zero/1HeLLo4uzjaLetFx6NH3PMwFP3qbRbTf3D but this can probably be fixed easily with some condition..

Where do you want it to redirect to? Probably the root of the site?

purplesyringa:
> Also, can you check if there can be any possible security problems with accessing WebSocket from a different origin and such?

I think that's safe. We're using wrapper_key as a security measure so there's no problem as long as the key verification is in action.

filips123:
> AFAIK there's no way to add a cookie to another site.

In non-web-based WebSocket clients (in Node.js and Python) you can set cookies. However, I'm not sure if browsers also allow this.

So yes, if sending cookies is not possible, it should be sent either in URL or in first message.

> Where do you want it to redirect to? Probably the root of the site?

Yes.

purplesyringa:
> In non-web-based WebSocket clients (in Node.js and Python) you can set cookies. However, I'm not sure if browsers also allow this.

You can set cookies in browsers too but only if they are same-origin.

filips123:
There is one other thing. Is there any reason why WebSocket requests are made to different origin that is specified in `server_url` instead of accessing from same origin? I think it may be with some issues with old Chrome proxy extension with WebSockets but I'm not sure?

purplesyringa:
I think there was a reason but I can't remember now. @HelloZeroNet?

HelloZeroNet:
Yeah PAC-based proxy configs/extensions does not work with websocket. So when you access http://zeronet.bit you still have to connect http://127.0.0.1:43110/Websocket

purplesyringa:
What is the state of this PR?

filips123:
Just accessing DNS domains works.

However, DNS reverse proxy with Multiuser plugin sometimes doesn't because of cookies and CORS.

filips123:
I have idea how this cross-origin cookie/multiuser handling could be implemented:

1. Wrapper/client normally sets cookie and tries to use it on WebSocket connection, like it does now.

2. When WebSocket server (Multiuser plugin) receives request, it checks for cookie. If it exists, it uses it to get account/address, also like it does now.

3. However, if cookie isn't set (probably because of CORS), instead of failing with error, it immediately sends some message/request/action/command to wrapper/client for user address.

4. Wrapper/client listens for that message, and if it receives it, it sends user address as response. Then it can start sending commands normally.

5. Once server receives response, it uses it as account. If it receives any other commands before response (and response was requested), it fails with error that user isn't authenticated (similarly to what currently happens).

Advantage of this is that if cookies are sent in first place, there isn't any need for any additional message, which means that it shouldn't affect speed. Additionally, it stays compatible with all current clients.

purplesyringa:
> Wrapper/client listens for that message, and if it receives it, it sends user address as response. Then it can start sending commands normally.

A small fix: user masterseed, not address.

filips123:
Master seed or master address? Cookie currently contains master address, so that response should probably also use it. And given that it will probably be used in somewhere in browser's storage (probably cookie just like now, but client just reads it from cookie and sends it to server if needed), it probably won't be good to store sensitive information such as master seed.

purplesyringa:
Daaaaaamn, and that's what we're calling "secure"... /cc @HelloZeroNet

PR #2215, opened 2019-10-02
Remove permission api call

rllola:
This PR is a re-write of https://github.com/HelloZeroNet/ZeroNet/pull/1494

- [x] It adds a `wrapperPermissionRemove` call. Calling `wrapperPermissionRemove` will prompt a notification message to ask user for confirmation.

- [x] If more than one permission is asked at the same time an error message will be send.

- [x] If the permission is not already granted it will prompt an informative message.

- [x] Also added a small the same behavior for `wrapperPermissionAdd`; it will send an error if more than one permission being passed as an argument and prompt a notification if permission already granted.

NOTE: The `wrapperPermissionAdd` verification for already granted permission was failing because we are checking if `Array is in Array` because in the documentation we passed the permission as an array (https://zeronet.io/docs/site_development/zeroframe_api_reference/#wrapperpermissionadd). In order to avoid this I added an if statement to check we passed an array.

NOTE 2: A small site to test http://127.0.0.1:43111/1MMF8RiHAqU2Ac6guTX5KLBeEsat846ien/


src/Ui/media/Wrapper.coffee
```diff
@@ -280,14 +282,37 @@ class Wrapper
...
 	actionPermissionAdd: (message) ->
 		permission = message.params
+		if Array.isArray(permission) && permission.length > 1
+			@sendInner {"cmd": "response", "to": message.id, "result": {"error": "Can only grant one permission at the time"}}
+			return false
+		if Array.isArray(permission) && permission.length == 1
+			permission = permission[0]
 		$.when(@event_site_info).done =>
 			if permission in @site_info.settings.permissions
+				@notifications.add "notification-#{message.id}", "info", "Permission already granted.", 4000
```
purplesyringa, line 292: Please remove this. This condition was added for the use case when you don't want to check whether you have the permission already.

purplesyringa (changes requested):

src/Ui/media/Wrapper.coffee
```diff
@@ -280,14 +282,37 @@ class Wrapper
...
+			@sendInner {"cmd": "response", "to": message.id, "result": {"error": "Can only remove one permission at the time"}}
+			return false
+		if Array.isArray(permission) && permission.length == 1
+			permission = permission[0]
+		$.when(@event_site_info).done =>
+			if permission not in @site_info.settings.permissions
+				@notifications.add "notification-#{message.id}", "info", "Permission already removed.", 4000
+				return false
+			@ws.cmd "permissionDetails", permission, (permission_details) =>
+				@displayConfirm "This site want to remove permission:" + " <b>#{@toHtmlSafe(permission)}</b>" + "<br><small style='color: #4F4F4F'>#{permission_details}</small>", "Remove", =>
```
purplesyringa: *wants

purplesyringa (changes requested):

purplesyringa:
I'm afraid the current implementation can give a false sense of security. A zite can draw a box similar to the real permission remove box, and the user would think that they removed the permission so they're now safe. This looks like a huge problem.

src/Ui/media/Wrapper.coffee
rllola: I think giving some form of feedback on what is happening is a good practice. If not a notification, we can send a message back.

Also note that the condition was always false because permission was an array. Even when the permission was already granted it would keep asking you.

purplesyringa: You could pass a single argument instead of an array of arguments but whatever. I think that you should just reply with an "already added" message. (but *not* an error)

HelloZeroNet:
I don't think it's a huge problem (actually I'm not sure if we need confirmation on permission remove at all), but to improve the situation we could give different look to the confirm dialogs generated by the wrapper.

purplesyringa:
I think there's no point in the site removing its own permission. It makes more sense to make a single configuration page for all sites.

src/Ui/media/Wrapper.coffee
rllola: > You could pass a single argument instead of an array of arguments but whatever.

Yeah, we need to update the documentation though because in the example it is an array (https://zeronet.io/docs/site_development/zeroframe_api_reference/#wrapperpermissionadd). I am not sure how to handle this without breaking compatibility in case people have done the same as in the documentation.

> I think that you should just reply with an "already added" message. (but not an error)

I can remove the notification and replace by a response to the site. However I do have an argument in favor of keeping it. The user is used to be asked if he wants to grant or not a permission. In the case the permission would have already been granted (but the user might not be aware of it) it would not prompt the question and the site might neither inform the user of it. The lack of information and the incoherent experience could be confusing for the user.

Does it make sense to you ?

purplesyringa: I'm afraid it still doesn't. See, many sites need Merger:whatever permission. They don't check whether they already have it, they just send wrapperPermissionAdd on page load. Currently, it would only bother the user once, but if you add notification it'll show up when the site is loaded and bother the user a lot more.

rllola: You made a good point. I agree with you regarding the Merger Site permission. I will proceed to modify it. Thank you for the feedback.

Issue #2100, opened 2019-07-15
IANA protocol register


6543:
we should register the zero:// handler

https://www.iana.org/form/protocol-assignment

filips123:
See also #2087.

It would be good to standardize and registered almost all parts of ZeroNet. This includes standardization of ZeroNet protocol (as new RFC) and `zero://` protocol, `.zeronet` TLD and port `43110` (at IANA).

Problem is that some of that standardizations require to register a company and pay a large amount of money. For example, to register official TLD you need to pay $185000.

6543:
so do it piece by piece?
if TLD is to expencive ... start with RFC and port 43110?

PS: spell check loves me :D so RFC isn't the best task for me

Merith-TK:
I would be willing to donate power to mine menero to do this

6543:
@Merith-TK this is an good idear i created #2101

filips123:
It might be possible to register `.zeronet` domain as "Special-Use Domain". I don't know if there is also a requirement to pay for that but it is based on RFCs.

For example, TLD `.onion` is registered for special use with [RFC 7686](https://tools.ietf.org/rfc/rfc7686.txt).

Same should be done (if possible) for `.zeronet`. It should not be hard to write that RFC. For example `.onion`'s RFC contains only 7 pages and actual RFC text is only on 4 pages.

filips123:
Here are also my thoughts about different ways about accessing ZeroNet and implementing them (see #2231 and #2087). I commented this here because there are currently quite some ZeroNet issues about DNS/protocols/schemes... and I think that this one is currently the most suitable for this type of comment.

---

## Ways of accessing ZeroNet

### zero:// scheme

* Simpler to implement than `z*://` schemes.
* A bit harder to implement than `.zeronet` TLD.
* Is almost schematically valid.
* But it in conflict with support for protocols other than HTTP and ZeroNet communication protocol.
* Does not allow using public ZeroNet instances.

### z*:// schemes

* Could be `zhttp://` for HTTP, `zhttps://` for HTTPS, `zftp://` for FTP...
* Harder to implement as each custom scheme would need to be handled separatly in the browser and every program you want to use it (local programs, CURL, API clients...).
* More schematically valid as it correctly separates different access protocols.
* Still does not allow using public ZeroNet instances.

### .zeronet TLD

* Simpler to implement (see the second part of this comment).
* Can be used in any program without too mich changes (also see second part of comment).
[...]

PR #2231, opened 2019-10-21
Road to better sandbox. Lowercase announce

purplesyringa:
This is the first PR of the "Road to better sandbox" series. This one announces lower-case site addresses as well as full site addresses to trackers.

The next steps should be:

- Refer to sites with lowercase addresses only;
- Support `http://address.zeronet` or `zhttp://address` or `zero://address` schemes;
- Switch to reverse iframe.

### DIFF
--- a/plugins/AnnounceBitTorrent/AnnounceBitTorrentPlugin.py
+++ b/plugins/AnnounceBitTorrent/AnnounceBitTorrentPlugin.py
@@ -59,7 +59,8 @@ class SiteAnnouncerPlugin(object):
         tracker.connect()
         if not tracker.poll_once():
             raise AnnounceError("Could not connect")
-        tracker.announce(info_hash=self.site.address_sha1, num_want=num_want, left=431102370)
+
+        tracker.announce(info_hash=self.site.address_lower_sha1, num_want=num_want, left=431102370)
         back = tracker.poll_once()
         if not back:
             raise AnnounceError("No response after %.0fs" % (time.time() - s))
@@ -67,6 +68,19 @@ class SiteAnnouncerPlugin(object):
             peers = back["response"]["peers"]
         else:
             raise AnnounceError("Invalid response: %r" % back)
+
+        if len(peers) < num_want:
+            # Also announce full site address. Using num_want instead of
+            # num_want - len(peers) is okay here because we expect many peers to
+            # switch to lowercase addresses
+            tracker.announce(info_hash=self.site.address_sha1, num_want=num_want, left=431102370)
+            back = tracker.poll_once()
+            if not back:
+                raise AnnounceError("No response after %.0fs" % (time.time() - s))
+            elif type(back) is dict and "response" in back:
+                peers += back["response"]["peers"]
+            else:
+                raise AnnounceError("Invalid response: %r" % back)
 
         return peers
 
@@ -105,46 +119,50 @@ class SiteAnnouncerPlugin(object):
             port = self.fileserver_port
         else:
             port = 1
-        params = {
-            'info_hash': self.site.address_sha1,
-            'peer_id': self.peer_id, 'port': port,
-            'uploaded': 0, 'downloaded': 0, 'left': 431102370, 'compact': 1, 'numwant': num_want,
-            'event': 'started'
-        }
 
-        url = protocol + "://" + tracker_address + "?" + urllib.parse.urlencode(params)
+        for hash in self.site.address_lower_sha1, self.site.address_sha1:
+            params = {
+                'info_hash': hash,
+                'peer_id': self.peer_id, 'port': port,
+                'uploaded': 0, 'downloaded': 0, 'left': 431102370, 'compact': 1, 'numwant': num_want,
+                'event': 'started'
+            }
+            url = protocol + "://" + tracker_address + "?" + urllib.parse.urlencode(params)
 
-        s = time.time()
-        response = None
-        # Load url
-        if config.tor == "always" or config.trackers_proxy != "disable":
-            timeout = 60
-        else:
-            timeout = 30
-
-        with gevent.Timeout(timeout, False):  # Make sure of timeout
-            req = self.httpRequest(url)
-            response = req.read()
-            req.close()
-            req = None
-
-        if not response:
-            raise AnnounceError("No response after %.0fs" % (time.time() - s))
-
-        # Decode peers
-        try:
-            peer_data = bencode.decode(response)["peers"]
-            if type(peer_data) is not bytes:
-                peer_data = peer_data.encode()
+            s = time.time()
             response = None
-            peer_count = int(len(peer_data) / 6)
-            peers = []
-            for peer_offset in range(peer_count):
-                off = 6 * peer_offset
-                peer = peer_data[off:off + 6]
-                addr, port = struct.unpack('!LH', peer)
-                peers.append({"addr": socket.inet_ntoa(struct.pack('!L', addr)), "port": port})
-        except Exception as err:
-            raise AnnounceError("Invalid response: %r (%s)" % (response, Debug.formatException(err)))
+            # Load url
+            if config.tor == "always" or config.trackers_proxy != "disable":
+                timeout = 60
+            else:
+                timeout = 30
+
+            with gevent.Timeout(timeout, False):  # Make sure of timeout
+                req = self.httpRequest(url)
+                response = req.read()
+                req.close()
+                req = None
+
+            if not response:
+                raise AnnounceError("No response after %.0fs" % (time.time() - s))
+
+            # Decode peers
+            try:
+                peer_data = bencode.decode(response)["peers"]
+                if type(peer_data) is not bytes:
+                    peer_data = peer_data.encode()
+                response = None
+                peer_count = int(len(peer_data) / 6)
+                peers = []
+                for peer_offset in range(peer_count):
+                    off = 6 * peer_offset
+                    peer = peer_data[off:off + 6]
+                    addr, port = struct.unpack('!LH', peer)
+                    peers.append({"addr": socket.inet_ntoa(struct.pack('!L', addr)), "port": port})
+            except Exception as err:
+                raise AnnounceError("Invalid response: %r (%s)" % (response, Debug.formatException(err)))
+
+            if len(peers) >= num_want:
+                break
 
         return peers
--- a/plugins/AnnounceLocal/AnnounceLocalPlugin.py
+++ b/plugins/AnnounceLocal/AnnounceLocalPlugin.py
@@ -80,13 +80,15 @@ class LocalAnnouncer(BroadcastServer.BroadcastServer):
         back = []
         sites = list(self.server.sites.values())
 
-        # Split adresses to group of 100 to avoid UDP size limit
-        site_groups = [sites[i:i + 100] for i in range(0, len(sites), 100)]
+        # Split adresses to group of 50 to avoid UDP size limit (two entries per
+        # site because of case)
+        site_groups = [sites[i:i + 50] for i in range(0, len(sites), 50)]
         for site_group in site_groups:
             res = {}
             res["sites_changed"] = self.server.site_manager.sites_changed
-            res["sites"] = [site.address_hash for site in site_group]
+            res["sites"] = sum([[site.address_hash, site.address_lower_hash] for site in site_group], [])
             back.append({"cmd": "siteListResponse", "params": res})
+
         return back
 
     def actionSiteListResponse(self, sender, params):
@@ -95,7 +97,7 @@ class LocalAnnouncer(BroadcastServer.BroadcastServer):
         num_found = 0
         added_sites = []
         for site in self.server.sites.values():
-            if site.address_hash in peer_sites:
+            if site.address_hash in peer_sites or site.address_lower_hash in peer_sites:
                 added = site.addPeer(sender["ip"], sender["port"], source="local")
                 num_found += 1
                 if added:
--- a/plugins/AnnounceLocal/Test/TestAnnounce.py
+++ b/plugins/AnnounceLocal/Test/TestAnnounce.py
@@ -73,7 +73,7 @@ class TestAnnounce:
 
         response_cmds = [response[1]["cmd"] for response in responses]
         assert response_cmds == ["discoverResponse", "siteListResponse"]
-        assert len(responses[-1][1]["params"]["sites"]) == 1
+        assert len(responses[-1][1]["params"]["sites"]) == 2  # two entries per site
 
         # It should only request siteList if sites_changed value is different from last response
         with Spy.Spy(announcer_remote, "handleMessage") as responses:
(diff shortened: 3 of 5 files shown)


filips123:
@imachug Can you also add support for `http://address.zeronet`? As noted in existing DNS related issues, Amazon applied for `.zero` TLD so there could be a conflict if they start using it. And if `.zeronet` becomes an official TLD for ZeroNet, we might also be able to register it to IANA as "Special-Use Domain". See https://github.com/HelloZeroNet/ZeroNet/issues/2100#issuecomment-516552136 for details.

Also, because handling lowercase addresses and supporting custom URL schemes and TLDs is a bit different thing than the actual implementation of new sandbox, but might be better to create this in separate PRs. Handling lowercase addresses and supporting custom URL schemes and TLDs should be the first part, and when it is finished, you should make another PR for actual sandbox implementation. **Update:** It looks like you already planned that 😃:

> This is the first PR of the "Road to better sandbox" series. This one announces lower-case site addresses as well as full site addresses to trackers.

purplesyringa:
Right, that's why it's a series of PRs instead of a single PR. But I still need this PR to get merged and people to get using it because we should drop full addresses soon.

filips123:
@imachug Yes, I missed that part of PR description...

Can you also add support for `.zeronet` TLD? Also, although I would like to get this working as fast as possible, we would really need to decide and standardize the way of accessing ZeroNet (custom protocol, custom TLD). If we do this before actual implementation will be used, it won't be hard to do this. But if we do this after users start using any of (unofficial) ways, it would be hard to tell them to start using another way.

I will also create another comment in one of DNS related issues about my thoughts about diffreant ways for accessing ZeroNet soon.

purplesyringa:
By the way, I've been recently exploring possible solutions for that multi-domain thing.

## Extensions

Looks great at first but:

- You have to pay for Google Developer account (for Chrome);
- You have to pay for Microsoft Developer account (for Edge);
- You have to create an MacOS app;
- You have to create your own browser instead of mobile Chrome.

## PAC file

Unsupported by Chrome. Supported globally by Windows, but that's a global setting (huh). Unsupported on Linux & Chrome without dnsmasq.

## Local proxy / local DNS

Looks like the best solution for now but possible insecure if *all* traffic passes through ZeroNet.

purplesyringa:
@filips123 Done.

filips123:
@imachug I created a comment about different ways of accessing ZeroNet and ways of implementing that in https://github.com/HelloZeroNet/ZeroNet/issues/2100#issuecomment-544664709.

krzysztof113:
For the schemes support:
Let's forget the extensions. They are all problems and limitations. They can be useful only for convenience improvements.
New version of firefox on android doesn't support them. Sideloading has it's problems, for example when using native host in Chrome you need to copy generated id into created files. Also Google can unpublish any moment they see fit.

filips123:
> New version of firefox on android doesn't support them

Because it is a preview version. It will get extensions support before it replaces the current version.

---

Also, which way would you suggest? Extensions and PAC file are the only simple way to do this.

purplesyringa:
@HelloZeroNet Any thoughts on this?

HelloZeroNet:
The problem with this is you won't be able to request the site from the peers, because on the protocol level the address is case-sensitive.

So:
 - You visit site 1hello that is actually 1HeLLo
 - You get peers from the tracker for 1hello
 - You try to connect to peers and execute `getFile "1hello", "content.json"`, but it will fail, because the peer has the site 1HeLLo

It's possible to change the protocol to make it insensitive, but it will result in lots of failed requests, because of clients that has not updated yet.

Another problem is the multiple tracker request, that is not really ideal.

Idea for different solutions:
 - On the first announce the tracker could indicate that this site has peers on different casing, the client could store it and use that case format from now to announce and for the file request and maybe even for the directory name. (Cons: It's not compatible with torrent trackers until the first successful request to a zero-protocol tracker)
 - Encoding it in the url: 1HeLLo -> 1h-el-l-o.anydomain.com

But the first step would be implement and use bech32 for the new sites.

purplesyringa:
> It's possible to change the protocol to make it insensitive, but it will result in lots of failed requests, because of clients that has not updated yet.

I've *already* made the protocol case-insensitive.

purplesyringa:
> Another problem is the multiple tracker request, that is not really ideal.

That will only apply for a few months. We'll remove case-sensitive address announcing when enough peers update.

purplesyringa:
> But the first step would be implement and use bech32 for the new sites.

Nope! That just adds unnecessary complication and breaks even more compatibility! And it doesn't help the problem at all because we can't just stop supporting old sites.

imachug:
@HelloZeroNet

purplesyringa:
@HelloZeroNet

HelloZeroNet:
The problem:
```
λ py -3 zeronet.py --silent peerCmd 127.0.0.1 15441 listModified "{'site': '1HeLLo4uzjaLetFx6NH3PMwFP3qbRbTf3D', 'since': 0}"
{
  "modified_files": {
    "content.json": 1571256561
  },
  "cmd": "response",
  "to": 1
}
```

```
λ py -3 zeronet.py --silent peerCmd 127.0.0.1 15441 listModified "{'site': '1hello4uzjaLetFx6NH3PMwFP3qbRbTf3D', 'since': 0}"
null
```

So even if you get the peers you won't be able to execute command on them unless you know the correct casing for the site address. So for compatibility it would be better to get the correct casing from the trackers.

imachug:
Did you run the command on my branch or `py3`?

purplesyringa:
@HelloZeroNet It's okay for now. This PR is from a PR series. It doesn't modify the way peers access sites or communicate. The only change is to announce lower-case addresses which will be required for next PRs.

purplesyringa:
@HelloZeroNet I have some great news, webextensions work really well. I'd like this PR to be merged ASAP to make sure I can share results as soon as possible.

d14na:
> But the first step would be implement and use bech32 for the new sites.

what's wrong with this approach? at first glance, this PR looks pretty hack'ish

> And it doesn't help the problem at all because we can't just stop supporting old sites.

why would you have to stop supporting old sites?

---

anyway, in no way an expert in bech32, but i do understand WHY is came to be, seems logical to support (and/or transition to) this address format. just my 2c

purplesyringa:
> what's wrong with this approach? at first glance, this PR looks pretty hack'ish

The next step would be to switch from `127.0.0.1:43110/address` to `address.zeronet`. Hostnames can't contain uppercase letters so we have to make addresses case-insensitive. One way to do this is to use bech32. However, this will break all existing vanity addresses and also make the format unfamilar to users. An easier way is to just make existing addresses lowercase.

purplesyringa:
There are lots of links to existing sites on ZeroNet. They would stop working if you switched to bech32.

purplesyringa:
@HelloZeroNet I really need this PR to be merged ASAP. The third PR is almost ready, and I have an extension for Firefox already.

purplesyringa:
Good news: I got the extension working in Chrome!

filips123:
@imachug Does it also work in Firefox and Edge? Also, did you do anything with DNS reverse-proxying? Because at least part which handles existing domains does not need this PR (but my DNS plugin PR).

purplesyringa:
It does work in Firefox. I haven't checked Edge though, I'm working on Tor Browser support now.

I'll check your DNS stuff right now.

purplesyringa:
@d14na I got the extension working on Chrome and I think all features I've thought of are ready. Could you please upload [webext.zip](https://github.com/HelloZeroNet/ZeroNet/files/3783710/webext.zip) to Chrome Web Store?

HelloZeroNet:
> @HelloZeroNet I really need this PR to be merged ASAP. The third PR is almost ready, and I have an extension for Firefox already.

I still think we should go for different solution, because if we merge it like this, then you won't be able to download the site from peers that has not updated to latest version and it's also creates twice as many request to trackers.

purplesyringa:
> then you won't be able to download the site from peers that has not updated to latest version

Hm? See, this PR only *announces* sites, so the trackers get both full and lower-case addresses. It doesn't change the way sites are *downloaded*. Users won't see any changes at all.

> and it's also creates twice as many request to trackers.

That's right. However, most trackers on ZeroNet are zero://-based so it should still be fast; also, this code is only going to run for a few weeks until we get most users updated.

ghost:
> @imachug Can you also add support for `http://address.zeronet`? As noted in existing DNS related issues, Amazon applied for `.zero` TLD so there could be a conflict if they start using it. And if `.zeronet` becomes an official TLD for ZeroNet, we might also be able to register it to IANA as "Special-Use Domain". See [#2100 (comment)](https://github.com/HelloZeroNet/ZeroNet/issues/2100#issuecomment-516552136) for details.
> 
> Also, because handling lowercase addresses and supporting custom URL schemes and TLDs is a bit different thing than the actual implementation of new sandbox, but might be better to create this in separate PRs. Handling lowercase addresses and supporting custom URL schemes and TLDs should be the first part, and when it is finished, you should make another PR for actual sandbox implementation. **Update:** It looks like you already planned that smiley:
> 
> > This is the first PR of the "Road to better sandbox" series. This one announces lower-case site addresses as well as full site addresses to trackers.

Slow down! What are you talking about? You think IANA just will add .zeronet as "Special-Use Domain"? :rofl: 
Doesn't need to! You can use any TLD for Zeronet! Amazon even applied for .amazon and has problems with countries where is Amazon. .zero? Ahh man, they should have .NULL! 

**I hate USA and Israel and they should be bombed to the ground! Fucking fascists! That's what they are! War criminal maniac terrorists!**

purplesyringa:
Hm, there are special-use domains. AFAIK we can register them by issuing an RFC. We could make a proof of concept (i.e. this PR series) and when we get enough people used to it, we'd make an RFC.

filips123:
> You think IANA just will add .zeronet as "Special-Use Domain"? 

They already added `.onion` domain, so I see no problem with adding `.zeronet`. You "just" need to create RFC specifying its usage.

> Doesn't need to!

If Amazon starts using and selling `.zero` TLD, there will be a conflict with ZeroNet which would use the same TLD. This means that normal `.zero` websites will not be usable for ZeroNet users and would generally create mess.

> Hm, there are special-use domains. AFAIK we can register them by issuing an RFC.

True.

ghost:
> Hm, there are special-use domains. AFAIK we can register them by issuing an RFC. We could make a proof of concept (i.e. this PR series) and when we get enough people used to it, we'd make an RFC.

Yes I know but they did not added .onion for at least 15 years... Think about. 
Internet Assigned Numbers Authority is a shitty American corporation. I have Namecoin and I fully support Namecoin. .bit is better than any trash administered by IANA.

filips123:
> Yes I know but they did not added .onion for at least 15 years... Think about.

Actually 11 years. It was the first introduced in 2004 and officialized in 2015.

Of course they won't register domain now as nobody uses it. But when ZeroNet and `.zeronet` TLD becomes more popular, we can create RFC to officialize it. Until then, it is pseudo TLD.

> I hate USA and Israel and they should be bombed to the ground! Fucking fascists! That's what they are! War criminal maniac terrorists!

Yes, but what you are currently doing is not a solution...

ghost:
@filips123 The solution I think would be to have only sandbox when you access zeronet on 43110 but it should not have sandbox when someone proxy normal TLD's to it or running a gateway. The sandbox can be removed in this cases and use X-Frame-Options: deny etc. Also Referrer-Policy. 

Only in this way ZeroNet can spread faster than Tor spreads currently! Even better ZeroNet actually helps Tor!

HelloZeroNet:
> Hm? See, this PR only announces sites, so the trackers get both full and lower-case addresses. It doesn't change the way sites are downloaded. Users won't see any changes at all.

I think it will lead to more dead sites, because after we do this you won't be able to connect to peers that running older version of the client.

purplesyringa:
Hm, it might make sense to make zero:// trackers support case-insensitive announce. Do you think it'll work?

purplesyringa:
@HelloZeroNet

HelloZeroNet:
Hm after thinking about it a bit more it won't be easy, because we not sending the site address to the tracker, but the hash of the site address.

So to do that we need to modify the protocol to send the address instead.

HelloZeroNet:
Right now I think the cleanest solution would be to switch to bech32 for new sites and encode old site address casing to domain `1h-el-l-o.anydomain.com` this way you will be able to connect to old clients and does not need any modification to the protocol.

purplesyringa:
What about the following (dumb) idea: we'd take a full address, transform it to lowercase but remember which characters were uppercase; we'd then encode that mask with base58 and append it to the address. This will support old vanity addresses. What do you think?

HelloZeroNet:
Well base58 is also case-sensitive, but using only lower-case characters could work, but I think marking uppercase characters with `-` is easier to do.

purplesyringa:
> Well base58 is also case-sensitive

🤦‍♂ A typo, sorry.

> but I think marking uppercase characters with - is easier to do

It *is* easier to do but it'll break vanity addresses. As a usecase: I often type `http://127.0.0.1:43110/1hell` and wait for my browser to remember the suffix (`o...`) and the correct case from the history. However, if we add `-` I'll also have to remember where to put dashes.

purplesyringa:
@HelloZeroNet Do you think that using my solution is acceptable?

HelloZeroNet:
Yes, it sounds good to me.

d14na:
> I got the extension working on Chrome and I think all features I've thought of are ready. Could you please upload webext.zip to Chrome Web Store?

@imachug google web store is blocking new account confirmation until $5 is paid; for "fraud" reasons. they're "refunding" my cc txs (don't know why), and they have NO CONTACT (other than no-reply@...); so i've reached out to the android team for help, i'll keep you posted if/when i'm able to submit to the web store.

__edit: if there's an EDGE solution ready; I'm happy to try my luck with MS__

filips123:
@d14na Any update about Chrome Web Store?

purplesyringa:
Ok, we're now getting licensing issues fixed (sslcrypto & bencode-open should fix almost all backend compatibility issues) so I'm returning to this PR series. I've just made lowercase addresses recoverable. The following algorithm is used:

## Making lowercase

- If the address starts with `0`, return it.
- Otherwise, make a binary integer with `i`th bit set iff `i`th letter/digit is uppercase.
- Encode that binary integer in little-endian format with 36-character alphabet (abc...xyz01..89) in 7 octets.
- Append the encoded integer to the original address.
- Replace the first address character with `0`.

## Recovering

- If the address starts with `1`, return it.
- Otherwise, parse the last 7 octets as a binary integer in little-endian format with 36-character alphabet (abc...xyz01..89).
- Capitalize `i`th character iff `i`th bit of the binary integer is set.
- Replace the first address character with `1`.

Example: 1HeLLo4uzjaLetFx6NH3PMwFP3qbRbTf3D -> 0hello4uzjaletfx6nh3pmwfp3qbrbtf3dlqf5lug.

filips123:
How would you handle this is address will be converted to lowercase before it will be sent to ZeroNet?

For example, if we choose `http://aDDREss.zeronet` scheme, `aDDREss` will be (probably?) converted to lowercase before it will send to ZeroNet server.

purplesyringa:
You'd use 0hello....zeronet to access ZeroHello. The backend would convert it to full address. I'm kinda not sure whether announcing is required at all in this case though... We could just reconvert the addresses on announce.

purplesyringa:
Yeah these changes are not required. I'm gonna work on the next PR.

PR #2242, opened 2019-10-24
No restart after update

rllola:
Check if it wants to restart after update before trying to restart ZeroNet.

It would leave the option to just "update and shutdown". It is also give a bit more flexibility for people using ZeroNet inside an application.

### DIFF
--- a/src/Ui/UiWebsocket.py
+++ b/src/Ui/UiWebsocket.py
@@ -1124,6 +1124,7 @@ class UiWebsocket(object):
 
             import main
             main.update_after_shutdown = True
+            main.restart_after_shutdown = True
             SiteManager.site_manager.save()
             main.file_server.stop()
             main.ui_server.stop()
--- a/zeronet.py
+++ b/zeronet.py
@@ -38,8 +38,9 @@ def main():
             import update
             print("Updating...")
             update.update()
-            print("Restarting...")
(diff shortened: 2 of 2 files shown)


HelloZeroNet:
Please also add it to main.py or it will display exception on shutdown

https://github.com/HelloZeroNet/ZeroNet/blob/py3/src/main.py

hacktivist:
Don't bother @rllola ... Just keep developing ZeroNet without sending PR to @HelloZeroNet . Worthless to help these Hungarian maniac fascists in Budapest. I seen and tested your ZeroNetBrowser and while it crashed at least you do try to do something positive and contribute for free to ZeroNet. The sad part of this is that while ZeroNet has now about 10 BTC collected as donations they will never use a cent of those donations to run Public Gateways or make videos educate people about ZeroNet. Nothing positive... 
I personally going to work on my own version of ZeroNet in a private repo.  Contribute nothing to this repo.

rllola:
I am not sure what should I add in the main.py file?

If it is the `restart_after_shutdown` variable, it is already set in the main.py file.

https://github.com/HelloZeroNet/ZeroNet/blob/py3/src/main.py#L15

HelloZeroNet:
Ah you are right, thanks

PR #2248, opened 2019-10-26
Update UiRequest.py

ghost:
Fonts are NOT applications! Added PGP related MIME types.

### DIFF
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -192,14 +192,37 @@ class UiRequest(object):
         file_name = file_name.lower()
         ext = file_name.rsplit(".", 1)[-1]
 
-        if ext == "css":  # Force correct css content type
+        # A non-complete list of MIME types
+        if ext == "txt":
+            content_type = "text/plain"
+        elif ext == "html":
+            content_type = "text/html"
+        elif ext == "css":
             content_type = "text/css"
-        elif ext == "js":  # Force correct javascript content type
+        elif ext == "js":
             content_type = "text/javascript"
-        elif ext == "json":  # Correct json header
+        elif ext == "woff2":
+            content_type = "font/woff2"
+        elif ext == "woff":
+            content_type = "font/woff"
+        elif ext == "ttf":
+            content_type = "font/ttf"
+        elif ext == "otf":
+            content_type = "font/otf"
+        elif ext == "png":
+            content_type = "image/png"
+        elif ext == "webp":
+            content_type = "image/webp"
+        elif ext == "webm":
+            content_type = "video/webm"
+        elif ext == "json":
             content_type = "application/json"
-        elif ext in ("ttf", "woff", "otf", "woff2", "eot"):
-            content_type = "application/font"
+        elif ext == "asc":
+            content_type = "application/pgp-keys"
+        elif ext == "gpg":
+            content_type = "application/pgp-encrypted"
+        elif ext == "sig":
+            content_type = "application/pgp-signature"
         else:
             content_type = mimetypes.guess_type(file_name)[0]
 
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -199,8 +199,6 @@ class UiRequest(object):
             content_type = "text/html"
         elif ext == "css":
             content_type = "text/css"
-        elif ext == "js":
-            content_type = "text/javascript"
         elif ext == "woff2":
             content_type = "font/woff2"
         elif ext == "woff":
@@ -209,14 +207,8 @@ class UiRequest(object):
             content_type = "font/ttf"
         elif ext == "otf":
             content_type = "font/otf"
-        elif ext == "png":
-            content_type = "image/png"
         elif ext == "webp":
             content_type = "image/webp"
-        elif ext == "webm":
-            content_type = "video/webm"
-        elif ext == "json":
-            content_type = "application/json"
         elif ext == "asc":
             content_type = "application/pgp-keys"
         elif ext == "gpg":
@@ -308,10 +300,16 @@ class UiRequest(object):
             headers["Access-Control-Allow-Headers"] = "Origin, X-Requested-With, Content-Type, Accept, Cookie, Range"
             headers["Access-Control-Allow-Credentials"] = "true"
 
-        if content_type == "text/html":
-            content_type = "text/html; charset=utf-8"
         if content_type == "text/plain":
             content_type = "text/plain; charset=utf-8"
+        if content_type == "text/html":
+            content_type = "text/html; charset=utf-8"
+        if content_type == "text/css":
+            content_type = "text/css; charset=utf-8"
+        if content_type == "application/json":
+            content_type = "application/json; charset=utf-8"
+        if content_type == "application/javascript":
+            content_type = "application/javascript; charset=utf-8"
 
         # Download instead of display file types that can be dangerous
         if re.findall("/svg|/xml|/x-shockwave-flash|/pdf", content_type):
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -305,7 +305,7 @@ class UiRequest(object):
             
         # UTF-8 character encoding
         if content_type in ("text/plain", "text/html", "text/css", "application/javascript", "application/json", "application/manifest+json"):
-            content_type = "%s;" % content_type + " " + "charset=utf-8"
+            content_type = content_type + "; charset=utf-8"
             
         # Caching
         if status in (200, 206) and content_type.startswith("text"):  # Cache MIME type text/* for 1 week
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -287,7 +287,7 @@ class UiRequest(object):
             headers["Access-Control-Allow-Origin"] = "*"
             
         if noscript:
-            headers["Content-Security-Policy"] = "default-src 'none'; sandbox allow-top-navigation allow-forms; img-src 'self'; font-src '127.0.0.1:43110'; media-src 'self'; style-src 'self' 'unsafe-inline';"
+            headers["Content-Security-Policy"] = "default-src 'none'; sandbox allow-top-navigation allow-forms; img-src 'self'; font-src *; media-src 'self'; style-src 'self' 'unsafe-inline';"
         elif script_nonce and self.isScriptNonceSupported():
             headers["Content-Security-Policy"] = "default-src 'none'; script-src 'nonce-{0}'; img-src 'self' blob:; style-src 'self' blob: 'unsafe-inline'; connect-src *; frame-src 'self' blob:".format(script_nonce)
 
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -308,21 +308,15 @@ class UiRequest(object):
             content_type = "%s;" % content_type + " " + "charset=utf-8"
             
         # Caching
-        cache_text = content_type.startswith("text")
-        cache_image = content_type.startswith("image")
-        cache_font = content_type.startswith("font")
-        cache_video = content_type.startswith("video")
-        nocache_application = content_type.startswith("application")
-
-        if status in (200, 206) and cache_txt_css_html:  # Cache MIME type text/* for 1 week
+        if status in (200, 206) and content_type.startswith("text"):  # Cache MIME type text/* for 1 week
             headers["Cache-Control"] = "public, max-age=604800"
-        if status in (200, 206) and cache_image:  # Cache MIME type image/* for 1 month
+        if status in (200, 206) and content_type.startswith("image"):  # Cache MIME type image/* for 1 month
             headers["Cache-Control"] = "public, max-age=2629746"
-        if status in (200, 206) and cache_font:  # Cache MIME type font/* for 6 months
+        if status in (200, 206) and content_type.startswith("font"):  # Cache MIME type font/* for 6 months
             headers["Cache-Control"] = "public, max-age=15778476"
-        if status in (200, 206) and cache_video:  # Cache MIME type video/* for 1 year
+        if status in (200, 206) and content_type.startswith("video"):  # Cache MIME type video/* for 1 year
             headers["Cache-Control"] = "public, max-age=31556952"
-        if status in (200, 206) and nocache_application:  # Not caching MIME type application/*
+        if status in (200, 206) and content_type.startswith("application"):  # Not caching MIME type application/*
             headers["Cache-Control"] = "no-cache, no-store, private, must-revalidate, max-age=0"
         headers["Content-Type"] = content_type
         headers.update(extra_headers)
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -207,6 +207,12 @@ class UiRequest(object):
             content_type = "application/pgp-encrypted"
         elif ext == "sig":
             content_type = "application/pgp-signature"
+        elif ext == "js":
+            content_type = "application/javascript"
+        elif ext == "json":
+            content_type = "application/json"
+        elif ext == "webmanifest":
+            content_type = "application/manifest+json"
         elif ext in ("woff", "woff2", "ttf", "otf", "sfnt", "collection"):
             content_type = "font/%s" % ext
         else:
@@ -300,10 +306,12 @@ class UiRequest(object):
             content_type = "text/html; charset=utf-8"
         if content_type == "text/css":
             content_type = "text/css; charset=utf-8"
-        if content_type == "application/json":
-            content_type = "application/json; charset=utf-8"
         if content_type == "application/javascript":
             content_type = "application/javascript; charset=utf-8"
+        if content_type == "application/json":
+            content_type = "application/json; charset=utf-8"
+        if content_type == "application/manifest+json":
+            content_type = "application/manifest+json; charset=utf-8"
 
         # Download instead of display file types that can be dangerous
         if re.findall("/svg|/xml|/x-shockwave-flash|/pdf", content_type):
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -285,7 +285,7 @@ class UiRequest(object):
         headers["Referrer-Policy"] = "no-referrer"
         if content_type != "text/html" and self.env.get("HTTP_REFERER") and self.isSameOrigin(self.getReferer(), self.getRequestUrl()):
             headers["Access-Control-Allow-Origin"] = "*"  # Allow load font files from css
-        if content_type == "text/javascript" and not self.env.get("HTTP_REFERER"):
+        if content_type == "application/javascript" and not self.env.get("HTTP_REFERER"):
             headers["Access-Control-Allow-Origin"] = "*"  # Allow loading JavaScript modules in Chrome
 
         if noscript:
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -199,14 +199,6 @@ class UiRequest(object):
             content_type = "text/html"
         elif ext == "css":
             content_type = "text/css"
-        elif ext == "woff2":
-            content_type = "font/woff2"
-        elif ext == "woff":
-            content_type = "font/woff"
-        elif ext == "ttf":
-            content_type = "font/ttf"
-        elif ext == "otf":
-            content_type = "font/otf"
         elif ext == "webp":
             content_type = "image/webp"
         elif ext == "asc":
@@ -215,6 +207,8 @@ class UiRequest(object):
             content_type = "application/pgp-encrypted"
         elif ext == "sig":
             content_type = "application/pgp-signature"
+        elif ext in ("woff", "woff2", "ttf", "otf", "sfnt", "collection"):
+            content_type = "font/%s" % ext
         else:
             content_type = mimetypes.guess_type(file_name)[0]
 
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -301,23 +301,41 @@ class UiRequest(object):
             headers["Access-Control-Allow-Headers"] = "Origin, X-Requested-With, Content-Type, Accept, Cookie, Range"
             headers["Access-Control-Allow-Credentials"] = "true"
 
-        if content_type in ("text/plain", "text/html", "text/css", "application/javascript", "application/json", "application/manifest+json"):
-            content_type = "%s;" % content_type + " " + "charset=utf-8"
-
         # Download instead of display file types that can be dangerous
         if re.findall("/svg|/xml|/x-shockwave-flash|/pdf", content_type):
             headers["Content-Disposition"] = "attachment"
             
-            # UTF-8 character encoding
-        cacheable_type = (
-            content_type == "text/css" or content_type.startswith("image") or content_type.startswith("video") or
-            self.env["REQUEST_METHOD"] == "OPTIONS" or content_type == "application/javascript"
+        # UTF-8 character encoding
+        if content_type in ("text/plain", "text/html", "text/css", "application/javascript", "application/json", "application/manifest+json"):
+            content_type = "%s;" % content_type + " " + "charset=utf-8"
+            
+        # Caching
+        cache_txt_css_html = (
+            content_type.startswith("text")
+        )
+        cache_image = (
+            content_type.startswith("image")
+        )
+        cache_font = (
+            content_type.startswith("font")
+        )
+        cache_video = (
+            content_type.startswith("video")
+        )
+        nocache_application = (
+            content_type.startswith("application")
         )
 
-        if status in (200, 206) and cacheable_type:  # Cache Css, Js, Image files for 10min
-            headers["Cache-Control"] = "public, max-age=600"  # Cache 10 min
-        else:
-            headers["Cache-Control"] = "no-cache, no-store, private, must-revalidate, max-age=0"  # No caching at all
+        if status in (200, 206) and cache_txt_css_html:  # Cache MIME type text/* for 1 week
+            headers["Cache-Control"] = "public, max-age=604800"
+        if status in (200, 206) and cache_image:  # Cache MIME type image/* for 1 month
+            headers["Cache-Control"] = "public, max-age=2629746"
+        if status in (200, 206) and cache_font:  # Cache MIME type font/* for 6 months
+            headers["Cache-Control"] = "public, max-age=15778476"
+        if status in (200, 206) and cache_video:  # Cache MIME type video/* for 1 year
+            headers["Cache-Control"] = "public, max-age=31556952"
+        if status in (200, 206) and nocache_application:  # Not caching MIME type application/*
+            headers["Cache-Control"] = "no-cache, no-store, private, must-revalidate, max-age=0"
         headers["Content-Type"] = content_type
         headers.update(extra_headers)
         return self.start_response(status_texts[status], list(headers.items()))
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -308,21 +308,11 @@ class UiRequest(object):
             content_type = "%s;" % content_type + " " + "charset=utf-8"
             
         # Caching
-        cache_txt_css_html = (
-            content_type.startswith("text")
-        )
-        cache_image = (
-            content_type.startswith("image")
-        )
-        cache_font = (
-            content_type.startswith("font")
-        )
-        cache_video = (
-            content_type.startswith("video")
-        )
-        nocache_application = (
-            content_type.startswith("application")
-        )
+        cache_text = content_type.startswith("text")
+        cache_image = content_type.startswith("image")
+        cache_font = content_type.startswith("font")
+        cache_video = content_type.startswith("video")
+        nocache_application = content_type.startswith("application")
 
         if status in (200, 206) and cache_txt_css_html:  # Cache MIME type text/* for 1 week
             headers["Cache-Control"] = "public, max-age=604800"
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -282,6 +282,7 @@ class UiRequest(object):
         headers["Connection"] = "Keep-Alive"
         headers["Keep-Alive"] = "max=25, timeout=30"
         headers["X-Frame-Options"] = "SAMEORIGIN"
+        headers["Referrer-Policy"] = "no-referrer"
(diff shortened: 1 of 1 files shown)


purplesyringa:
Why are the `.png` (and similar) changes required? `mimetypes` module sets correct mimetypes for (most) images already.

ghost:
Feel free to add or remove MIME types. I think it's better to have common extensions stated than rely on the guessing of the `mimetype` module. This is my preference, that is why I included the .png, .webm and others.

purplesyringa:
It's not *guessing*. The `mimetypes` module detects mimetype based on the extension, and `.png` is hardcoded. You can take a look at its source code if you're not sure.

ghost:
`else: content_type = mimetypes.guess_type(file_name)[0]` **is guessing** by checking the extension and then looking up in the module the supposedly correct MIME type. I think that ZeroNet should not rely on the module since some types are incorrect or not included at all like the PGP MIME types which I stated.

purplesyringa:
> since some types are incorrect

Which ones?

I see no reason in adding unneeded code when it works correctly. To be honest, I think that it's mimetypes module that should be patched, not ZeroNet code, but that's unlikely.

purplesyringa:
I'd leave `.js`, `.webp`, fonts and gpg stuff only. I'm also not sure whether `.js` is required (`application/javascript` is a correct mimetype as well but old browsers might not support it). Everything else is correct.

HelloZeroNet:
> Type name:
> application

According to https://www.iana.org/assignments/media-types/application/font-woff

purplesyringa:
@HelloZeroNet [Take a look at this RFC](https://tools.ietf.org/html/rfc8081).

purplesyringa:
@HelloZeroNet @filips123 @krixano Do you see any problems in using `application/javascript` instead of `text/javascript`? I'd leave `text/javascript` but we could remove it (and thus simplify code) if `application/javascript` works as well.

purplesyringa:
`.txt`, `.html` and `.css` are handled by mimetypes module as well. BTW, I'd recommend not to use `.html` extension to detect mimetype because ZeroNet uses `text/html` mimetype to detect when to add sandbox. Mentioning `.html` might make it look like if `mimetypes` module could be removed, but there are other (strange) extensions like `.htm` and `.xht`, so removing might have serious consequences for ZeroNet security.

filips123:
@imachug Regarding JS mime type, I already searched about this around the web and it seems that  nobody actually know which one is correct.

Also, if you are already adding new mime types, you should also add `.webmanifest` ==> `application/manifest+json` mime type as this is official mime type for Web App Manifest for PWA. It was already added to `mimetypes` module (by me), but the change is only present in Python 3.8+.

HelloZeroNet:
```
$ curl --head https://github.githubassets.com/static/fonts/inter/Inter-Medium.woff | grep content-type
Content-Type: application/font-woff
```

But if we want to change it, then
```
        elif ext in ("woff", "woff2", "ttf"):
            content_type = "font/%s" % ext
```
would be more clean/compact solution

ghost:
Yeah @imachug that's correct! https://www.iana.org/assignments/media-types/media-types.xhtml#font and [[RFC8081](https://tools.ietf.org/html/rfc8081)] states **`font/`** instead of `application/` MIME for fonts.

ghost:
@filips123 JavaScript should be `application/javascript` and not text/javascript according to [RFC4329](https://tools.ietf.org/html/rfc4329#page-9).

ghost:
This pr also belongs in the mimetypes repo - don't blame ZeroNet devs for things that aren't of their doing. In the meantime, a workaround is good, because this fonts issue has been a problem for a while I think.

HelloZeroNet:
> fonts issue has been a problem for a while I think.

As far I know there were no issue reported that is connected to invalid mime types.

All issues I have seen was because if someone enables the UiPassword plugin, then the cookie won't be passed to the css font requests and the client will display the login form instead of the resource.

ghost:
Idk, I'm pretty sure there was a comment in one issue that mentioned this problem with the mime type. Maybe I can find it, idk.

ghost:
@HelloZeroNet It's right here: #1940 
Tangdou closed it before it was even fixed. I made a comment on it recently.

ghost:
Hold on a sec... Previously, I wasn't getting correct results when running this in python when that issue (#1940 ) originally occured:
> print(__import__('mimetypes').guess_type('MaterialIcons-Regular.woff'))

But now I am getting "font/woff" in python. I suppose the mimetypes module was fixed for fonts but the zeronet workaround wasn't removed or something? idk...

Aso, @HelloZeroNet , some people were still getting font problems in Tor browser very recently... don't know if this pr fixes it though.

HelloZeroNet:
> @HelloZeroNet It's right here: #1940
> Tangdou closed it before it was even fixed. I made a comment on it recently.

As he stated in last comment: "When enable the ui_password plugin , the problem will reproduce ."

So it was because of the sandbox.

filips123:
> But now I am getting "font/woff" in python. I suppose the mimetypes module was fixed for fonts but the zeronet workaround wasn't removed or something?

Which Python version do you have? Because `mimetypes` is in standard library, it is updated with Python. So maybe this was changed in recent Python versions.

purplesyringa:
@filips123 @krixano `mimetypes` library has a small built-in list of common extensions, but other mimetypes are loaded from registry (in Windows) and configuration files (in Linux).

ghost:
@filips123 
> Why you didn't add Web App Manifest. If I would add it, this would be another PR and if you already have PR that updated mime types this would be unnecessary complicating.

See: https://github.com/w3c/manifest/issues/689#issuecomment-408558218

filips123:
@imachug There are around 100 built-in extensions but yes other are loaded from system.

purplesyringa:
@hacktivist It looks like the comment you're referring too agrees with @filips123, doesn't it?

filips123:
@hacktivist Can you explain that comment a bit more? Because I see it as point **for** using `.webmanifest` extension and correct content type.

ghost:
> As he stated in last comment: "When enable the ui_password plugin , the problem will reproduce ."
> 
> So it was because of the sandbox.

First of all, that last comment could be interpreted in two different ways... that the problem occurs with ui_password enabled or that it occurs with ui_password enabled **also**.

Secondly, don't take things out of context and actually look at the issue and what I quoted for my last comment... previously in the issue I tested the mimetypes library and it didn't give the correct results for fonts **in the python REPL** which indicated a potential mimetypes library or OS config problem. Hence why **you** suggested this:

> So it depends on the operating system configuration if its application/font-woff or None. In case of None we return "application/octet-stream" as content-type which works for browsers I tested.
> 
> To make it more predictable we could write our own content-type resolver that does not depends on operating system configuration files.

If operating systems are inconsistent, then that's a problem that obviously needs to be worked around. Also notice that you said you'll return "application/octet-sream" if None was given back...

ghost:
@filips123 
Quote from https://github.com/w3c/manifest/issues/689#issuecomment-408558218
> If a developer doesn't have access to server configurations then `.json` would be the safest way to deploy the manifest.

Also please look at [Web App Manifest #warning](https://w3c.github.io/manifest/#issue-container-generatedID)

purplesyringa:
Right, but we *do* have access to server configurations, don't we?

> The official file extension for the manifest is .webmanifest. Some web servers recognize this extension and transfer the file using the standardized media type for a manifest (application/manifest+json). Developers can also choose a different extension (e.g. .json) or none at all (e.g. /api/GetManifest), but are strongly encouraged to transfer the manifest using the application/manifest+json media type.

purplesyringa:
Sorry, my browser lagged so I didn't see you were referring to the warning. Anyways, mimetypes are usually rather stable so I doubt they'll change it.

ghost:
@imachug He can use `<link rel="manifest" href="manifest.webmanifest">` or `<link rel="manifest" href="manifest.json">` and the `mimetypes` module will do its job.

ghost:
Ok, well your previous comments suggested you thought it shouldn't be added because of webmanifest's instability/non-standardization...

purplesyringa:
@hacktivist Why use hacks when a small fix is enough?

ghost:
Should I add back `application/json`? I removed because `mimetypes` module states correctly. If we add `application/manifest+json` than I think `application/json` also needs to be added.

purplesyringa:
> If we add application/manifest+json than I think application/json also needs to be added.

Why though? I think that adding application/manifest+json is enough.

purplesyringa:
Do you mean that application/manifest+json is already reported correctly by mimetypes module? This only works on *some* devices because it's not built-in.

filips123:
Mimetypes module only resolves .webmanifest in Python 3.8 or if this is in system configuration. But in most cases, it is not and most users don't have Python 3.8 yet.

And the only change needed for this is to check for `.webmanifest` and return mime type `application/manifest+json`. Changing `.json` is not needed.

ghost:
I adding both. Do we need also `application/javascript`? If html, css etc stated than we should add this one too. What about encoding? `.webmanifest` also should be utf8 right?

purplesyringa:
> I adding both.

But why? Why would you duplicate mimetypes's work?

> Do we need also application/javascript? If html, css etc stated than we should add this one too.

We should *remove* html and css.

> What about encoding? .webmanifest also should be utf8 right?

I think so.

ghost:
I adding `application/javascript` for `.js`, `application/json` for `.json` and `application/manifest+json` for `.webmanifest`. I will not remove `text/simple`,` text/html` and `text/css` because from line 303 these MIME types are explicitly stated with `utf8` encoding.

purplesyringa:
< I adding application/javascript for .js, application/json for .json

But why? .js and .json are already handled by mimetypes.

> I will not remove text/simple

Does that make sense? I mean, these two parts are unrelated...

> text/html and text/css because from line 303 these MIME types are explicitly stated with utf8 encoding.

Ok, feel free to ignore these because the diff would be smaller.

HelloZeroNet:
> @krixano Also notice that you said you'll return "application/octet-sream" if None was given back...

Loading fonts with application/octet-sream content-type should work.

Eg. Cloudflare/Firefox serve it this way:

```
$ curl --head https://www.mozilla.org/media/fonts/Metropolis-Bold.6a80125e795a.woff2
Content-Type: binary/octet-stream
```

So I still think the problem was because of missing cookie header and not because of invalid content type, but including more content types in UiRequest.py should not hurt.

HelloZeroNet:
I prefer serving js/css/json as `text/*`, because this way if you open in a browser you can view the content instead of downloading it.

Google.com also serves it this way, so I think it can be considered a valid option.

Update: Looks like in latest chrome js/json also displayed and not downloaded when serving it as application/* (maybe it got changed or I was remember incorrectly)

ghost:
> > @krixano Also notice that you said you'll return "application/octet-sream" if None was given back...
> 
> Loading fonts with application/octet-sream content-type should work.
> 
> Eg. Cloudflare/Firefox serve it this way:
> 
> ```
> $ curl --head https://www.mozilla.org/media/fonts/Metropolis-Bold.6a80125e795a.woff2
> Content-Type: binary/octet-stream
> ```
> 
> So I still think the problem was because of missing cookie header and not because of invalid content type, but including more content types in UiRequest.py should not hurt.

It should be served as `font/` because if it's not than ZeroNet will use the `mimetypes` module which will serve it as `application/` which is wrong.

Do not even think about to follow or listen what CloudFlare does or say because than I will be forced to pull out all of my hair one-by-one.

ghost:
@imachug `application/manifest+json` is also handled by the `mimetypes` module https://github.com/python/cpython/blob/3.8/Lib/mimetypes.py at least in Python3 which should be the target of ZeroNet because January 1st 2020 will be Python2's EOL (End of Life).

filips123:
@Hacktivist Only since Python 3.8. But most users don't use it yet.

purplesyringa:
@hacktivist 

> 3.8

> at least in Python3

src/Ui/UiRequest.py
```diff
@@ -300,18 +300,8 @@ def sendHeader(self, status=200, content_type="text/html", noscript=False, allow
...
-            content_type = "text/html; charset=utf-8"
-        if content_type == "text/css":
-            content_type = "text/css; charset=utf-8"
-        if content_type == "application/javascript":
-            content_type = "application/javascript; charset=utf-8"
-        if content_type == "application/json":
-            content_type = "application/json; charset=utf-8"
-        if content_type == "application/manifest+json":
-            content_type = "application/manifest+json; charset=utf-8"
+        if content_type in ("text/plain", "text/html", "text/css", "application/js", "application/json", "application/manifest+json"):
```
purplesyringa: application/js -> application/javascript

ghost: :O Thanks!

ghost:
> Loading fonts with application/octet-sream content-type should work.
> 
> Eg. Cloudflare/Firefox serve it this way:
> 
> $ curl --head https://www.mozilla.org/media/fonts/Metropolis-Bold.6a80125e795a.woff2
Content-Type: binary/octet-stream
> 
> So I still think the problem was because of missing cookie header and not because of invalid content type, but including more content types in UiRequest.py should not hurt.

@HelloZeroNet I don't think you get what I was saying at all. You were trying to say that that other issue wasn't about this at all, but you literally propose a "fix" for this exact problem in that other issue (and you never actually stated that it was fixed) - so it **was** about this.

> So I still think the problem was because of missing cookie header and not because of invalid content type

Because apparently you can't have two problems at once right? I was having that content-type font problem also, and I wasn't using ui_password (and I probably have photo-evidence of this somewhere on ZeroMe)... so that goes to show.
I mean, logically, the problem was *very clearly* **both** ui_password and a wrong content-type on fonts considering both YOU and I tested it and it wasn't giving back the correct content-type, but let's just forget about that why don't we!

Anyways @HelloZeroNet , icon fonts don't currently work in Tor browser - fix your crap!

New Issue: #2249

ghost:
The MIME type `font/*` should be used on **fonts** because, I do serving ZeroNet content with a reverse proxy and some resource is cached on the client, more specifically the 1,5 MB `woff2`.

ghost:
@imachug Now the fonts working? :)

src/Ui/UiRequest.py
```diff
@@ -285,24 +299,41 @@ def sendHeader(self, status=200, content_type="text/html", noscript=False, allow
...
         if re.findall("/svg|/xml|/x-shockwave-flash|/pdf", content_type):
             headers["Content-Disposition"] = "attachment"
-
-        cacheable_type = (
-            content_type == "text/css" or content_type.startswith("image") or content_type.startswith("video") or
-            self.env["REQUEST_METHOD"] == "OPTIONS" or content_type == "application/javascript"
+            
+        # UTF-8 character encoding
+        if content_type in ("text/plain", "text/html", "text/css", "application/javascript", "application/json", "application/manifest+json"):
+            content_type = "%s;" % content_type + " " + "charset=utf-8"
```
purplesyringa: What magic is this??

purplesyringa (changes requested):

src/Ui/UiRequest.py
```diff
@@ -285,24 +299,41 @@ def sendHeader(self, status=200, content_type="text/html", noscript=False, allow
...
+        )
+        nocache_application = (
+            content_type.startswith("application")
         )
 
-        if status in (200, 206) and cacheable_type:  # Cache Css, Js, Image files for 10min
-            headers["Cache-Control"] = "public, max-age=600"  # Cache 10 min
-        else:
-            headers["Cache-Control"] = "no-cache, no-store, private, must-revalidate, max-age=0"  # No caching at all
+        if status in (200, 206) and cache_txt_css_html:  # Cache MIME type text/* for 1 week
```
purplesyringa: AFAIK we don't do cache invalidation so this might be a problem. Static site update might take a long while...

purplesyringa (changes requested):

src/Ui/UiRequest.py
ghost: This magic works. If you now start arguing about UTF8 character encoding... What you don't like?

purplesyringa: Formatting and concatenation in the same expression, lol.

src/Ui/UiRequest.py
ghost: Cache is `public` as you see. In the other hand after `siteSign` there should be `sitePublish` which will notify all peers about the changes.

purplesyringa: How does this help? I mean, the browser would still use cached html, wouldn't it?

ghost: Yes, but not for 10 minutes. It would be crazy to assume that a static sites changes in every 10 minutes. 
Does the browser cache `content.json`? No, is not because we are not caching any MIME type **application/***, specially not JavaScript.

src/Ui/UiRequest.py
ghost: Anything is possible in Python. :rofl:

src/Ui/UiRequest.py
purplesyringa: It's crazy to assume 10 minutes but it's crazy to assume a whole week too. I'd recommend 2 hours or so.

purplesyringa: How is content.json related? It's not accessed by browser at all.

purplesyringa: Ok, it looks like you thought of site updating as of publishing. I actually meant that users would receive updates but their browsers wouldn't show the changes.

ghost: Note:
On line 299 in `cacheable_type` JavaScript was `application/javascript`:
`            self.env["REQUEST_METHOD"] == "OPTIONS" or content_type == "application/javascript"`

On line 272 which was also [removed](https://github.com/HelloZeroNet/ZeroNet/pull/2248/commits/091f30fbd7163a28736e970f9a432b67d4dab4c7), JavaScript used as `text/javascript`:
`        if content_type == "text/javascript" and not self.env.get("HTTP_REFERER"):`

ghost: > Ok, it looks like you thought of site updating as of publishing. I actually meant that users would receive updates but their browsers wouldn't show the changes.

Depend, I seen that many people rocking with Tor Browser. :man_technologist: 
Caching is handy when you have a gateway or ZeroNet is behind a reverse proxy.

purplesyringa: What about making a `--cache_for` configuration option then? @HelloZeroNet

hacktivist: I can add `stale-while-revalidate` to MIME type **text/*** if that helps.

purplesyringa: Ok, I guess that makes sense. I'd still add `--cache_for` though, but I guess it should be another PR.

ghost: @imachug I don't think that need `--cache_for`. 
I do have 200+ static websites and I can tell you for sure, that I'm not updating them every 10 minutes. Not even in one day. Caching is good how you see in these changes.

src/Ui/UiRequest.py
filips123: Just use `content_type = content_type + " " + "charset=utf-8;"`. This current version is really weird...

src/Ui/UiRequest.py
purplesyringa: In case of ZeroNet, static websites are often actually made with jekyll or similar tools (look at @caryoscelus'es blog). *These* kinds of sites are updated rather often.

src/Ui/UiRequest.py
hacktivist: **text/css; charset=utf-8** :heavy_check_mark: 
~~text/html charset=utf-8;~~

ghost: **text/html; charset=utf-8** :heavy_check_mark: 
~~text/html charset=utf-8;~~

src/Ui/UiRequest.py
ghost: Nothing is stopping him in lowering the default `max-age` value if needed.

src/Ui/UiRequest.py
filips123: I misplaced `;`. I mean `content_type = content_type + "; charset=utf-8"` or just `content_type += "; charset=utf-8"`.

src/Ui/UiRequest.py
purplesyringa: Who is "him"? @caryoscelus?

ghost: Yes. You come up with one specific example and I replied accordingly. 

According to my checks by refreshing the site in the browser you will eventually get a fresh copy of the static resources. 
When you use _back_ and _forward_ button than obviously you will see the cached version. 

See in ZeroNet `frame` and using `raw`.

purplesyringa: > Yes. You come up with one specific example and I replied accordingly.

caryoscelus (or whoever runs a site) can't just change caching settings on *others'* computers. How would you even do that?

> According to my checks by refreshing the site in the browser you will eventually get a fresh copy of the static resources.

This "eventually" often includes accidentally pressing Shift / Ctrl or doing something in web console. That's not something that an average user should think about.

> See in ZeroNet frame and using raw.

Now what?.. Are you asking me to use raw mode? That's often not an option -- static sites *can* include JS, e.g. for animations.

ghost: Let's say you have a chat in ZeroNet and you send `text/html` to the cache for one week while using JavaScript to print messages. Not to mention you can use JavaScript to refresh the site time-to-time without needing to mess around the default caching of the resources. Correct me if I'm wrong.

purplesyringa: Let's talk about a specific example -- a static blog. Pretend that I had this blog on clearnet (e.g. on GitHub pages) but then I moved it to ZeroNet. So, my blog doesn't work with ZeroNet API, it's just a collection of html files and a few JS files that are used for animations and other interactive content. Now, how would you reload site content when a new post is published?

ghost: No MIME type starts with `application/*` is cached! This includes JavaScript, JSON and others like `application/octet-stream`. 
You can use query string on static resources like ZeroNet does, example: **all.css?rev=666**.

purplesyringa: We're talking about *HTML pages*, not scripts!

ghost: Let me be very clear on that ZeroNet is not in development in order to provide hosting for Jekyll! If you want a blog than there are a tons of options. You can **clone ZeroTalk** for example or just go and run it on WordPress. 
Most people don't even use this frameworks and it will be fatal to assume that HTML pages must work like a dynamic blog! I say it again. The proposed caching is good.

purplesyringa: What the heck? ZeroNet aims at easy single-click zite creation. It should be easy to both create sites from scratch and copy clearnet sites. You can't just say "we won't do that"!

hacktivist: I can also tell you this in Hungarian if you need. I born in Hungary, but left the country when Orban come into power nearly 10 years ago. My sister is a "border hunter" in the Croatian border, hunting real living people. I strongly believe that ZeroNet can really hurt those in power and I will not going to argue with you about how long the HTML should be cached. 

I, for that matter will always use this caching because as I said I do have 200+ static websites. It is already a pain to sign the new content and I don't want to see my server overloaded with request for static resources like HTML, CSS, video and other which are currently fetched every 10 minutes.

purplesyringa: Ok, that's why I asked to add a `--cache_for` option. It could be set to 2 hours by default and increased by those who have lots of sites.

ghost:
:disappointed: I sent the wrong link in https://github.com/HelloZeroNet/ZeroNet/pull/2248/commits/99bea958e79ebcfff5669df6c724193d70ed3140 that should be: https://github.com/HelloZeroNet/ZeroNet/pull/2248#discussion_r339349062

src/Ui/UiRequest.py
ghost: > Most people don't even use this frameworks and it will be fatal to assume that HTML pages must work like a dynamic blog! I say it again. The proposed caching is good.

Yeah, you're an idiot.

He was talking about static HTML pages. If you change an HTML page it won't get the latest version until the cache is expired.

You know... have the actual **text inside the html file**!

hacktivist: Choose your words you retard!

hacktivist: I will send another PR but the caching stays how it is. Nobody and I say it again NOBODY in this word changes HTML pages every 10 minutes or so. You can use JavaScript or whatever you want if you need to update the text. Show me an example HTML site where the text changes every 10 minutes otherwise be silent.

ghost: I'm not the one here who thinks that everything *has* to use javascript to load content.

And you know what ZeroNet **is** in development to do... provide websites to users! LOL Jekyll is a generator for a website, right? So...

> Let me be very clear on that ZeroNet is not in development in order to provide hosting for Jekyll! 

Actually, ZeroNet is **precisely** for this.

> Most people don't even use this frameworks

Where are you gettting this information? Source please.

> The proposed caching is good.

So... your idiotic opinion is reason enough to **actively prevent** people from easily changing the caching?

hacktivist: I will send another PR but the caching stays how it is. Nobody and I say it again NOBODY in this world changes HTML pages every 10 minutes or so. You can use JavaScript or whatever you want if you need to update the text. Show me an example HTML site where the text changes every 10 minutes otherwise be silent.

hacktivist: Go fckyourself with you Jekyll.

ghost: If you think that static zites aren't updated more than every week, then yeah... you're an idiot. If you think that zites have control over caching of stuff, then you're just assuming ZeroNet has this features without ever seeing it anywhere.

And btw, ZeroTalk is not a Blog... what you meant was *ZeroBlog*.

hacktivist: You know what I will not send a PR because you are a disgusting idiot who probably never in his life wrote a website from 0 with HTML. You just talk shit about programs that generate HTML sites but you fail to come up with an example where HTML is changes every 10 minutes. 

ZeroNet should check the content.json and if the hash is different than refresh the site. That would be the solution but instead of writing the code up you are here to spit shit.

ghost: > ZeroNet should check the content.json and if the hash is different than refresh the site. 

You clearly have no idea what you're talking about. LOL

ghost: > sites but you fail to come up with an example where HTML is changes every 10 minutes.

You know how I can tell you're a **very fishy** person... because you keep spewing that everyone's saying "10 minutes" when nobody has said that. Everyone here has only said changes happen shorter than **a week**.

I'll be making my own PR.

Btw, I **did** have a static zite, lol. And it changed frequently. Again with the fishiness... you clearly spew crap when you don't actually know what the hell you're talking about, lol

hacktivist: @krixano Seriously, you opened a fucking issue about font loading and I did fixed it. You are an idiot and you know it because you was unable to fix such a simple thing as font loading in Tor Browser.

hacktivist: You make your own PR but copy nothing from mine!

ghost: Right... Because apparently everyone who opens an issue also can't program? Because apparently you think that there's **no other** reason for people to not fix the issue themselves. Because apparently people don't have work or school or anything else going on in their life. Again a **very fishy person**.

Btw, there's a different between fixing something and half-assing it (and deliberately preventing things from working because you're an asshole).

Btw, you should probably stop with the Strawman (and Tu Quoque)

hacktivist: I think you should be kicked out of this repo because you appreciate nothing but talking shit. ZeroNet should really distance itself from people like you. You ruining it not developing. What developing, you don't know even how to fix font loading or where to look. You are a disaster and you look like one according to your photo, website etc..

ghost: Right... says the person who also said crap about ivanq who is the second highest for most commits and has fixed more vulnerabilities than you've even found in ZeroNet, LOL

hacktivist: I'm sorry? I come to this repo 2 days ago! 
Sorry that there is no a million PR already. How can be someone so retarded like you?

ghost: > You make your own PR but copy nothing from mine!

Does this mean you're claiming full copyright for your commit?

@HelloZeroNet I'd be weary of the potential legal consequences of this in the future. This is precisely why open-source repos have a file that requires all committers to give up rights to their commits.

hacktivist: I'm Hungarian too just that you know.

HelloZeroNet:
I have pushed modifications in Rev4250 that has updated list of bundled content types based on this PR and also solves the font loading problems.
Adding long cache values is not a great idea as the content can change any time.

ghost:
Also posted on https://github.com/antifa/ZeroNet/issues/1

Here's a 3 way diff. Center is the original before hacktivist/antifa's PR. Left is nofish's implementation. Right is hacktivist/antifa's implementation from the PR (https://github.com/HelloZeroNet/ZeroNet/pull/2248)

Screenshot from 2019-11-05 00-28-20
Screenshot from 2019-11-05 00-27-28
Screenshot from 2019-11-05 00-27-47
Screenshot from 2019-11-05 00-28-03

Here's a video of me scrolling through the whole file:
https://www.youtube.com/watch?v=YaxcDekczYo

Here's another diff I did - a two way diff comparing ZeroNet current code with antifa/hacktivist's PR:
https://www.diffchecker.com/NO4OAFuM
(Note: because this is two way, it doesn't show which parts were in there originally before the PR or nofish's implementation).

ghost:
I'm posting here to update people on what's going on with the user @hacktivist / @antifa. After antifa was blocked, he created a new account called @LiberateZeroNet. He has deleted his account again and created a new account called @fukcyouintheass. This account has already been deleted, and 6 hours ago he has created *another* new account called AnonymousChronicler. I believe that he has an additional account that he has created an hour ago called bela-miklos. Both of these accounts have one repository that happens to be a copy of ZeroNet source code (but not shown as a fork, and under a different name).

Every single time I find out about an additional username this person created, I report it to GitHub Support where I have an issue open with them about the user hacktivist all the way back from when he still used that name - I created this report with GitHub Support on October 30th.

Screenshot from 2019-11-05 19-02-08
Screenshot from 2019-11-05 19-02-25
Screenshot from 2019-11-05 19-02-40
Screenshot from 2019-11-05 19-02-55
Screenshot from 2019-11-05 19-03-13

ghost:
Another update - Looks like another account by this person has been created: PICSANRUGLAK
Joined an hour ago.

PR #2255, opened 2019-10-28
BUHH

ghost:
:angry:

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,27 @@
+ZeroNet proudly uses target _blank on links while escaping sandbox! 
+How can be someone so stupid that breaking apart my Pull Request, 
+copy what he likes and than making changes without that fixing the security hole what currently ZeroNet has. 
+You don't even read Pull Request. What a shame! 
+
+
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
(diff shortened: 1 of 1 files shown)


hacktivist:
You just don't wanted to merge my PR because you and some kids are so fucking ignorant that it was better to break my pull request but you are also a retard because the security hole is front of your eyes. In the big rush you forgot copy that one from my PR.

HelloZeroNet:
Please open a new issue with the details if you found any security problem.

ghost:
I'm gonna take a wild guess and say that he's talking about line 281 in his pr @HelloZeroNet

ghost:
Screenshot from 2019-10-28 11-43-09

HelloZeroNet:
The no-referrer header does not fixes the linked issue.
To avoid tabnabbing we would need to add `noopener` rel to the links, but as far as I know there is no way to do it by http headers.

ghost:
@HelloZeroNet https://cheatsheetseries.owasp.org/cheatsheets/HTML5_Security_Cheat_Sheet.html#tabnabbing
Screenshot from 2019-10-28 11-56-58

Also, we have the `wrapperOpenWindow` zeroframe cmd that we need to do this for.

HelloZeroNet:
The wrapperOpenWindow already work like this:
https://github.com/HelloZeroNet/ZeroNet/blob/py3/src/Ui/media/Wrapper.coffee#L235

ghost:
Ok, I didn't know that. Good!

ghost:
> The wrapperOpenWindow already work like this:
> https://github.com/HelloZeroNet/ZeroNet/blob/py3/src/Ui/media/Wrapper.coffee#L235

Sending the header is better!

PR #2259, opened 2019-10-29
emptiness

ghost:
no worry man

### DIFF
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -275,14 +275,14 @@ class UiRequest(object):
         headers["Version"] = "HTTP/1.1"
         headers["Connection"] = "Keep-Alive"
         headers["Keep-Alive"] = "max=25, timeout=30"
-        headers["X-Frame-Options"] = "sameorigin"
+        headers["Referrer-Policy"] = "no-referrer"
         if content_type != "text/html" and self.env.get("HTTP_REFERER") and self.isSameOrigin(self.getReferer(), self.getRequestUrl()):
             headers["Access-Control-Allow-Origin"] = "*"  # Allow load font files from css
 
         if noscript:
-            headers["Content-Security-Policy"] = "frame-ancestors 'self'; default-src 'none'; sandbox allow-top-navigation allow-forms; img-src *; font-src * data:; media-src *; style-src * 'unsafe-inline';"
+            headers["Content-Security-Policy"] = "frame-ancestors 'self'; sandbox allow-top-navigation allow-forms;"
         elif script_nonce and self.isScriptNonceSupported():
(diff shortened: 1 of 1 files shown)


purplesyringa:
Here's what I get when I add an iframe:

> Refused to display 'http://127.0.0.1:43110/1HeLLo4uzjaLetFx6NH3PMwFP3qbRbTf3D' in a frame because an ancestor violates the following Content Security Policy directive: "frame-ancestors 'self'".

Was that intended or do you know how to fix that?

purplesyringa:
Ok, something to investigate then: can we use `none` instead? It looks like SAMEORIGIN (instead of DENY) has been in the code from the beginning, but I think it makes more sense to use DENY. The change was introduced by @MuxZeroNet but he left the project. @HelloZeroNet Do you know why 
(potentionally unsafe) SAMEORIGIN was used?

hacktivist:
:+1: `none`

HelloZeroNet:
Deny work work, because you have to load the site into the sandboxed iframe

purplesyringa:
@hacktivist Could you update the PR then?

HelloZeroNet:
Btw the this PR will remove the limitations in case script nonce is not supported by the browser or script_nonce is not supplied to the function.

So X-Frame-Options header is still necessary

And the format is also broken: you need to use ; instead of , and please put it to the end of the rule

hacktivist:
So `X-Frame-Options: deny` and `Content-Security-Policy: frame-ancestors 'none';` ?
Can't have `X-Frame-Options: sameorigin` while `frame-ancestors 'none'`!

hacktivist:
If `X-Frame-Options: deny` than ZeroNet will not work. 
If `Content-Security-Policy: frame-ancestors` is `'none'` than `X-Frame-Options:` **must** be `deny`! 

Keeping `X-Frame-Options: sameorigin` and setting `Content-Security-Policy: frame-ancestors` to `'self'` **without** disposition `enforce`.

@HelloZeroNet: https://github.com/HelloZeroNet/ZeroNet/pull/2259#issuecomment-547594614
> Deny ~~work~~ won't work, because you have to load the site into the sandboxed iframe.

hacktivist:
@imachug @HelloZeroNet 

I just see now in: https://www.w3.org/TR/CSP2/#frame-ancestors-and-frame-options

> The frame-ancestors directive obsoletes the X-Frame-Options header. If a resource has both policies, the frame-ancestors policy SHOULD be enforced and the X-Frame-Options policy SHOULD be ignored.

This changed in https://www.w3.org/TR/CSP3/#frame-ancestors-and-frame-options

HelloZeroNet:
no, thanks

PR #2263, opened 2019-10-30
Don't share ignored files

purplesyringa:
This allows sensitive data to be stored in ignored files. (caution: don't use it on proxies though)

### DIFF
--- a/src/File/FileRequest.py
+++ b/src/File/FileRequest.py
@@ -13,6 +13,7 @@ from Config import config
 from util import RateLimit
 from util import Msgpack
 from util import helper
+from util import SafeRe
 from Plugin import PluginManager
 from contextlib import closing
 
@@ -214,6 +215,23 @@ class FileRequest(object):
             self.connection.badAction(5)
             return False
         try:
+            # Don't send ignored files
+            if not params["inner_path"].endswith("/content.json"):
+                dirs = params["inner_path"].split("/")  # Parent dirs of content.json
+                inner_path_parts = []  # Filename relative to content.json
+                content = None
+                while dirs:
+                    inner_path_parts.insert(0, dirs.pop())
+                    content_inner_path = "%s/content.json" % "/".join(dirs)
+                    content = site.content_manager.contents.get(content_inner_path.strip("/"))
+                    if content:
+                        break
+                relative_inner_path = "/".join(inner_path_parts)
+                if content and content.get("ignore") and SafeRe.match(content["ignore"], relative_inner_path):
+                    self.response({"error": "Ignored file"})
+                    self.connection.badAction(5)
+                    return False
+
             file_path = site.storage.getPath(params["inner_path"])
             if streaming:
                 file_obj = site.storage.open(params["inner_path"])
--- a/src/Test/TestPeer.py
+++ b/src/Test/TestPeer.py
@@ -49,6 +49,10 @@ class TestPeer:
         buff = peer_file_server.getFile(site_temp.address, "content.json")
         assert b"sign" in buff.getvalue()
 
+        # Testing ignored file
+        buff = peer_file_server.getFile(site_temp.address, "js/ignored.json")
+        assert not buff
+
         connection.close()
         client.stop()
 


HelloZeroNet:
It could be useful, but it could have significant performance cost as you have to read/parse content.json on every file request.

If we introduce this probably we need to add the same for http requests and API commands as well.

purplesyringa:
> as you have to read/parse content.json on every file request.

Not really. ContentManager caches content.json, doesn't it?

purplesyringa:
> If we introduce this probably we need to add the same for http requests and API commands as well.

Here's a usecase: I run a script in background in browser which reads privatekey from an ignored file. I want the script to access it by I don't want to allow accessing it from outside.

HelloZeroNet:
> Not really. ContentManager caches content.json, doesn't it?

It does have some caching, but only keep last 10 requested one, so if someone downloads a file with 1000s of user files it could be costly

btw: you need to add `site_temp.storage.write("js/ignored.json", b"test")` to the test or it will fail anyway as the file does not exists

HelloZeroNet:
> Here's a usecase: I run a script in background in browser which reads privatekey from an ignored file. I want the script to access it by I don't want to allow accessing it from outside.

I would recommend to give it a random file name and store it in the settings

purplesyringa:
> btw: you need to add site_temp.storage.write("js/ignored.json", b"test") to the test or it will fail anyway as the file does not exists

I added `js/ignored.js` file, didn't I?

purplesyringa:
> I would recommend to give it a random file name and store it in the settings

Isn't that a hack? This solution looks better to me...

HelloZeroNet:
> I added js/ignored.js file, didn't I?

Ah yeah, I did not noticed.

> Isn't that a hack? This solution looks better to me…

Using random files instead of authentication is pretty common. (Google and FB also serving private files this way)

Pros for using random files:
 - Does not have performance impact when serving files
 - Makes it harder to steal the private key if cross-origin file read get compromised
 - Secure in multi-user environment. (Public proxy)

ghost:
But can't users just use the fileList zeroframe command on a public proxy to list out all the files?

HelloZeroNet:
> But can't users just use the fileList zeroframe command on a public proxy to list out all the files?

You are right, it's possible to do that. So the only thing left is userSetSettings

ghost:
Ok, here's what I propose. A new thing called "ziteSetSettings" and "ziteGetSettings" that stores *zite* settings in a location inaccessible from the frontend. 

The difference between this and userSettings would be that userSetSettings are *user-specific* while ziteSetSettings would *not* be. A good use for this is a proxy owner adding a setting that affects a zite *for all users of the proxy*, and technically sensitive data can also be stored here.

HelloZeroNet:
What's the problem with userSetSettings/userGetSettings in this case?

Can you give example for ziteSetSettings? I'm also not sure how could you use that for sensitive data if it's accessible for all users.

ghost:
Right.... I'm dumb, basically, lol

Anyways, the idea is still cool imo - the ability to have site-settings that apply to all users of a proxy, but is changeable by the proxy-owner.

----

Edit: What we want this for is for KxoId creation - the ability to host a "Trusted Peer" on a proxy without proxy users getting the private key used to sign ids. Basically, we need the ability for the *plugin* (BackgroundProcessing in this case) to store a zite setting somewhere outside of any zite folders and this setting should be only accessible to the plugin/zeronet core (and *not* zeroframe). But it also has to be zite-specific too.

I believe ivanq should probably just add this to his plugin.

purplesyringa:
Closing this PR then.

HelloZeroNet:
> Anyways, the idea is still cool imo - the ability to have site-settings that apply to all users of a proxy, but is changeable by the proxy-owner.

I'm not sure about the use-case, but it can be done by writing the setting to an ignored json file.

There is also a siteSetSettingsValue, but currently it's limited to one key.

ghost:
Well... the main idea would be that the proxy-owner has control over it, so an ignored json file would require they go into the zite directory to change that file.

Issue #2323, opened 2019-11-22
Site content.json load error: ProgrammingError: Recursive use of cursors not allowed.


slrslr:
### Step 1: Please describe your environment

  * ZeroNet version: rev4290
  * Operating system: W10
  * Web browser: FF beta
  * Tor status: available

### Step 2: Describe the problem:

I am updating zeronet every a few days. In last maybe 2-4 days i found Zeronet to crash at least 2 times. I mean process is not running. Today i checked the error.log which was empty and last line in debug.log was:

> [2019-11-22 ***] ERROR    Site:1ADQAH..g6aA Json data/users/mypublickeyhere/content.json load error: ProgrammingError: Recursive use of cursors not allowed. in SiteStorage.py line 342 > ContentFilter/ContentFilterPlugin.py line 180 > SiteStorage.py line 101 > Db.py line 364 > DbCursor.py line 192 > DbCursor.py line 103

(ZN tried to publish new comment on that site. But on re-start publishing same or similar comment in same place on that site seem to work)
In previous sessions, debug log files does not contain any errors at the end of files.

HelloZeroNet:
Thanks for reporting, I was able to reproduce similar problem and it should be fixed by: #2323

slrslr:
@HelloZeroNet 
today i was running rev4303 on Windows10 and i seen similar error like one described in this issue above. Here is the error i got under rev4303:

> [2019-11-26 \*\*\*] ERROR    Site:18Pfr2..pSzh WebSocket handleRequest error: ProgrammingError: Recursive use of cursors not allowed. in UiWebsocket.py line 79 > UiWebsocket.py line 232 > OptionalManager/UiWebsocketPlugin.py line 163 
>  {"cmd":"optionalFileList","params":{"filter":"","address":"18Pfr2oswXvD352BbJvo59gZ3GbdbipSzh","limit":1000},"wrapper_nonce":"\*\*\*","id":7}

i remember that on that site i clicked Seed file button, clicked some video in "Up next" sidebar of the site that appears when playing a video.

one hour later, zeronet process silently crashed (disappeared) and above mentioned log line was the last ERROR line in the file, rest was just INFO. error.log was empty..

HelloZeroNet:
Thanks, I was able to reproduce the error this time:
Looks like it happens the cursor and the returned result processor is in separate threads.
Working on the fix...

HelloZeroNet:
[...]

PR #2324, opened 2019-11-22
potential fix for #2323

ghost:


HelloZeroNet:
Thanks, it would be great if we could reproduce the problem and make sure gevent.lock.RLock works for different threads.

and probably it would be better to use
```
try:
    from gevent.lock import RLock
except:
    from gevent.coros import RLock
```

as gevent.coros was only used up until gevent 1.2.0 (2016-12-23)

ghost:
> Thanks, it would be great if we could reproduce the problem and make sure gevent.lock.RLock works for different threads.
> 
> and probably it would be better to use
> 
> ```
> try:
>     from gevent.lock import RLock
> except:
>     from gevent.coros import RLock
> ```
> 
> as gevent.coros was only used up until gevent 1.2.0 (2016-12-23)

No problem, changed the import statements. 
As for reproducing the problem, I'm not sure... The issue only gives a vague description of what the problem is.

HelloZeroNet:
I was able to reproduce a similar problem:
```
sqlite3.OperationalError: cannot start a transaction within a transaction
2019-11-23T11:40:59Z <Greenlet at 0x350c8c8: threadedExecute(['INSERT INTO test (key, value) VALUES (?, ?)', (0)> failed with OperationalError
```

Unfortunately looks like RLock does not work between threads:

```

  File "C:\Python3\lib\site-packages\gevent\lock.py", line 223, in acquire
    rc = self._block.acquire(blocking)
  File "src\gevent\_semaphore.py", line 100, in gevent.__semaphore.Semaphore.acquire
  File "src\gevent\_semaphore.py", line 128, in gevent.__semaphore.Semaphore.acquire
  File "src\gevent\_abstract_linkable.py", line 192, in gevent.__abstract_linkable.AbstractLinkable._wait
  File "src\gevent\_abstract_linkable.py", line 165, in gevent.__abstract_linkable.AbstractLinkable._wait_core
  File "src\gevent\_abstract_linkable.py", line 169, in gevent.__abstract_linkable.AbstractLinkable._wait_core
  File "src\gevent\_greenlet_primitives.py", line 60, in gevent.__greenlet_primitives.SwitchOutGreenletWithLoop.switch
  File "src\gevent\_greenlet_primitives.py", line 60, in gevent.__greenlet_primitives.SwitchOutGreenletWithLoop.switch
  File "src\gevent\_greenlet_primitives.py", line 64, in gevent.__greenlet_primitives.SwitchOutGreenletWithLoop.switch
  File "src\gevent\__greenlet_primitives.pxd", line 35, in gevent.__greenlet_primitives._greenlet_switch
greenlet.error: cannot switch to a different thread
```

But it work this way:

```py
    global is_executing
    while is_executing:
        time.sleep(0.01)
    is_executing = True

    try:
        res = db.execute(*query)
    finally:
        is_executing = False
    return res
```

it's not the most elegant solution and some unnecessary sleep could happen, but it has minimal overhead under normal conditions and expect that this this will happen rarely.

my full test script:
```py
import time
import gevent
import gevent.monkey
import gevent.threadpool
import sqlite3
import shutil
gevent.monkey.patch_all(thread=False, threading=False)

def watcher():
    print("Watcher started")
    while 1:
        s = time.time()
        time.sleep(0.1)
        if time.time() - s > 0.15:
            print("- Block:", time.time() - s)
gevent.spawn(watcher)

db = sqlite3.connect("db.db", isolation_level="DEFERRED", check_same_thread=False)
db.row_factory = sqlite3.Row
try:
    db.execute("DROP TABLE test")
except Exception as err:
    pass
db.execute('PRAGMA journal_mode=WAL')
db.execute("CREATE TABLE test (key INTEGER, value TEXT)")

pool_db = gevent.threadpool.ThreadPool(10)

from gevent.lock import RLock
import gevent.thread

is_executing = False
num_blocked = 0
lock = RLock()
def execute(*query):
    global is_executing, num_blocked
    while is_executing:
        num_blocked += 1
        time.sleep(0.01)
    is_executing = True

    try:
        res = db.execute(*query)
    finally:
        is_executing = False
    return res

def threadedExecute(query):
    res = pool_db.spawn(execute, *query).get()
    return [query, res.fetchall()]

s = time.time()
threads = []
for i in range(1000):
    query = ["INSERT INTO test (key, value) VALUES (?, ?)", (i, "data %s" % i)]
    t = gevent.spawn(threadedExecute, query)
    threads.append(t)

gevent.joinall(threads)
print("Insert executed in %.3fs" % (time.time() - s))
print("Blocked:", num_blocked)

threads = []
for i in range(1000):
    query = ["SELECT * FROM test WHERE test.value LIKE 'data %s%%' OR test.value = ? LIMIT 5" % i, ["data %s" % i]]
    t = gevent.spawn(threadedExecute, query)
    threads.append(t)

gevent.joinall(threads)
print("Select executed in %.3fs" % (time.time() - s))
print("Blocked:", num_blocked)

for thread in threads:
    query, res = thread.value
    assert len(res), "No result for: %s" % query
    assert res[0]["value"].startswith(query[1][0]), "%s not starts with %s" % (res[0]["value"], query[1])

```

HelloZeroNet:
Update: `from gevent._threading import Lock` looks compatible with threading, so replacing RLock with that should work

ghost:
So replace reentrant locks with Locks and see if that works?

PR #2339, opened 2019-12-04
Update jp.json

viasnake:
Update japanese translation file jp.json.


HelloZeroNet:
Looks like lots of things from the previous translation has been removed. Can you please also include them?

viasnake:
I think that the deleted translation is a translation of Sidebars.
https://github.com/HelloZeroNet/ZeroNet/blob/py3/plugins/Sidebar/languages/jp.json

Do I need to include them in this translation file as well?

HelloZeroNet:
You are right, it's not necessary to be there, thanks.

PR #2340, opened 2019-12-05
Adding VTT MIME type and making sure that all MIME types starting with text/* are using UTF-8 character encoding.

SwissConfederation:
VTT stands for "Video Text Tracks Format" and it requires UTF-8 character encoding. 
About VTT: https://www.w3.org/TR/webvtt1/

This will fix the "subtitles not shown at all" issue on videos which use them on ZeroNet.

### DIFF
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -35,6 +35,7 @@ content_types = {
     "json": "application/json",
     "sig": "application/pgp-signature",
     "txt": "text/plain",
+    "vtt": "text/vtt",
     "webmanifest": "application/manifest+json",
     "webp": "image/webp"
 }
@@ -292,7 +293,7 @@ class UiRequest(object):
             headers["Access-Control-Allow-Headers"] = "Origin, X-Requested-With, Content-Type, Accept, Cookie, Range"
             headers["Access-Control-Allow-Credentials"] = "true"
 
-        if content_type in ("text/plain", "text/html", "text/css", "application/javascript", "application/json", "application/manifest+json"):
+        if content_type.startswith("text") or content_type in ("application/javascript", "application/json", "application/manifest+json"):
             content_type += "; charset=utf-8"
 
         # Download instead of display file types that can be dangerous


src/Ui/UiRequest.py
```diff
@@ -292,7 +293,7 @@ def sendHeader(self, status=200, content_type="text/html", noscript=False, allow
             headers["Access-Control-Allow-Headers"] = "Origin, X-Requested-With, Content-Type, Accept, Cookie, Range"
             headers["Access-Control-Allow-Credentials"] = "true"
 
-        if content_type in ("text/plain", "text/html", "text/css", "application/javascript", "application/json", "application/manifest+json"):
+        if content_type.startswith("text") or content_type in ("application/javascript", "application/json", "application/manifest+json"):
```
purplesyringa: I think text/ would be more correct.

SwissConfederation: You think wrong.

purplesyringa: Hm. Let me explain my choice. You want to handle `text/*`, i.e. MIME types within `text` section. That's exactly what `startswith("text/")` does.

On the other hand, `startswith("text")` matches MIME sections that start with `text`. Currently, the only section stating with `text` is `text` itself. Even if this works, it's counterintuitive and it might break in case more MIME sections are added in the future.

purplesyringa: This is the same as subdirectory matching: `file.startswith(dir + "/")` matches iff `file` is inside `dir` whilst `file.startswith(dir)` doesn't work for e.g. `file="1.txt", dir="1"`.

SwissConfederation: An example [content_type.startswith(‘text’)](https://www.iquda.co.uk/embedded-pdf-file-view-plone/) from www.iquda.co.uk. 

> startswith("text") matches MIME sections that start with text

That's correct and so all text should be encoded in UTF-8. 

There will be never a MIME type like textdirdur/imachug for video files! I don't get it why you talk about directories.... 

The `text` MIME type always used on text files and this will never change, so this pull request can't break anything now or in the future.

SwissConfederation: If you don't mind, I mark it as resolved.

purplesyringa: `text/*` will always be used for text files, that's right. However, it's not guaranteed that something like `text-like/*` isn't added in the future.

As for the directories, that was rather obvious, wasn't it? MIME sections are like directories, i.e. `text/plain` is like a file called `plain` in `text` directory.

SwissConfederation:
```
        # Caching
        if status in (200, 206) and content_type.startswith("text"):  # Cache MIME type text/* for 1 week
            headers["Cache-Control"] = "public, max-age=604800"
        if status in (200, 206) and content_type.startswith("image"):  # Cache MIME type image/* for 1 month
            headers["Cache-Control"] = "public, max-age=2629746"
        if status in (200, 206) and content_type.startswith("font"):  # Cache MIME type font/* for 6 months
            headers["Cache-Control"] = "public, max-age=15778476"
        if status in (200, 206) and content_type.startswith("video") or content_type.startswith("audio"):  # Cache MIME type video/* and audio/* for 1 year
            headers["Cache-Control"] = "public, max-age=31556952"
        if status in (200, 206) and content_type.startswith("application") or content_type.startswith("message"):  # Not caching MIME types application/* and message/*
            headers["Cache-Control"] = "no-cache, no-store, private, must-revalidate, max-age=0"
        headers["Content-Type"] = content_type
        headers.update(extra_headers)
        return self.start_response(status_texts[status], list(headers.items()))
```

SwissConfederation: I close this discussion, this is clearly resolved. I can even write `content_type.startswith("te")` and it will still match all text/* MIME types.

purplesyringa: > This caching pretty much covers the entire mimetypes module.

Right but there's a reason why caching wasn't added...

> So in the future I'm doomed?

Uh, maybe. I don't know whether that's going to happen but there's some probability.

> You talk about directories and how a piece of code look like, but here and now in this pull request we talking about content_type.startswith.

How are MIME types different from file paths? The only difference I can find is that directory names / sections are registered by some committee, not created by users.

purplesyringa: > I close this discussion, this is clearly resolved. I can even write content_type.startswith("te") and it will still match all text/* MIME types.

It will match text/* but it might match something else like test/* and such. That's clearly not what you're looking for.

imachug:
BTW, I'm not sure why Python 3.8 is failing. There's some bug with streaming I guess.

src/Ui/UiRequest.py
SwissConfederation: Right now, it is good without forward slash. When a new MIME arrives like `text-like/*` :rofl: as you said, than I will send another PR. Playing as Nostradamus-Imachug is not the best right now. I can also predict that in the future will nobody use ZeroNet... Right? So why developing it.. You see. I take care of UiRequest when needed.

SwissConfederation:
> BTW, I'm not sure why Python 3.8 is failing. There's some bug with streaming I guess.

Check the logs... :roll_eyes: I have nothing to do with that.

src/Ui/UiRequest.py
purplesyringa: Come on, `text/` will work for sure and `text` might break so you say "I'll fix that later" instead of defeating the problem right now?

imachug:
You know, comments aren't just for you. I should have pinged @HelloZeroNet though.

SwissConfederation:
He not own those machines at travis! He sees what we see. These are just some random failures. Is not the first time travis fails...

src/Ui/UiRequest.py
SwissConfederation: Let's see you travis... :rofl:

imachug:
These failures are **not** random. If tests are failing, there are some reasons for their failures and they might occur in the wild.

imachug:
Just to make sure... what license are you licensing this contribution under? *As if you'd contribute not to sue us*

SwissConfederation:
It should be GPLv3+ because of the **future**. You said [once](https://github.com/HelloZeroNet/ZeroNet/pull/2340#discussion_r354362478): 
> so you say "I'll fix that later" instead of defeating the problem right now?

We should have a GPLv3+ and not GPLv3-only because than again need voting if new version comes in a few years.. Defeat the problem now @imachug :+1:

imachug:
Hm. It's actually the same logic as with the slash: GPLv3-only works for sure but GPLv4/5/etc. might be not what we're looking for.

SwissConfederation:
Doesn't matter what you looking for now! :smiley: If licensed under GPLv3-only if later version comes out let's say in the next 20 years, than probably would be extremely difficult to relicense it. 

If GPLv3+ is the license than everyone who get a copy can upgrade his versions without looking for previous contributors on a non-existent GitHub... in the future of course.

imachug:
GPLv3-only will always work. GPLv3+ might contain "...and allows beating the author to death in case some parts of the software are not provided for free..." and such.

SwissConfederation:
**It is exactly the same license!** One allows to use newer versions of GPL the other is not. That's the only difference.

imachug:
I meant GPLv4 / GPLv5 / GPLvn instead of GPLv3+, sorry if that was difficult to understand.

SwissConfederation:
> I meant GPLv4 / GPLv5 / GPLvn instead of GPLv3+, sorry if that was difficult to understand.

I did understand you perfectly.

SwissConfederation:
Be aware that very possibly soon or later @HelloZeroNet also will drops out of the project. I don't think that everyone else who comes after should be restricted to one license. They should able to use newer versions if any! They can distribute they modifications under either the new license (maybe even better) or keep using GPLv3.

ExtremelyMassiveStar:
@imachug I add FLAC also to MIME types.

imachug:
Please create a separate PR if you want to. Consider using a single PR by feature.

ExtremelyMassiveStar:
> Please create a separate PR if you want to. Consider using a single PR by feature.

Well, GitHub automatically does this.

imachug:
Please separate the license change and mime type stuff to two PRs. You can revert license change on your py3 branch and ask GitHub to edit license on a new branch (it'll be sth like patch-1 I guess).

imachug:
You know, there's a reason for a-single-PR-per-feature model, it allows less cluttered discussions.

ExtremelyMassiveStar:
Please mind your own business. I at least sending pull requests. I don't see why need to separate them. I change the title if you will feel better. Not to mention you (@imachug) started to talk here about the license! So seems appropriate to send here the pr. All in.

ExtremelyMassiveStar:
> You know, there's a reason for a-single-PR-per-feature model, it allows less cluttered discussions.

Maybe this is obvious and don't need discussion. Discussing with whom? You and @filips123 ? You gonna be kidding me. All you do is criticizing me while I send pull requests.

ExtremelyMassiveStar:
@imachug did you test frame-ancestors? I think Content-Security-Policy can be stripped even more.

ExtremelyMassiveStar:
@imachug do you have time to check this for me? you can test it with your zites. I only tested with mine.

ExtremelyMassiveStar:
@HelloZeroNet your Test folder causes travis to fail nearly all the time.

ExtremelyMassiveStar:
Related to https://github.com/HelloZeroNet/ZeroNet/pull/2340/commits/0df747fdba8913f37a34126fefd87150d896f8d2
https://w3c.github.io/webappsec-csp/#frame-ancestors-navigation-response
`Given a request (request), a string navigation type ("form-submission" or "other"), a response (navigation response) a browsing context (target), a string check type ("source" or "response"), and a policy (policy) this algorithm returns "Blocked" if one or more of the ancestors of target violate the frame-ancestors directive delivered with the response, and "Allowed" otherwise.`

imachug:
> do you have time to check this for me? you can test it with your zites. I only tested with mine.

Will do either tomorrow or later roday.

PR #2356, opened 2019-12-10
Replace incompatible, terminated license with GPLv3+; make changes to README.md; update UiRequest.py

ghost:


filips123:
Just that you know, merging this PR without consent from all contributors would be illegal...

PR #2358, opened 2019-12-12
Switch to bencode_open

purplesyringa:
The official bencode parser implementation (bencode.py) is license under Open-Source BitTorrent license which is incompatible with GPL. This PR replaces bencode.py with my implementation (which is [available on PyPI as well](pypi.org/project/bencode_open)).

### DIFF
--- a/plugins/AnnounceBitTorrent/AnnounceBitTorrentPlugin.py
+++ b/plugins/AnnounceBitTorrent/AnnounceBitTorrentPlugin.py
@@ -3,7 +3,7 @@ import urllib.request
 import struct
 import socket
 
-import bencode
+import bencode_open
 from lib.subtl.subtl import UdpTrackerClient
 import socks
 import sockshandler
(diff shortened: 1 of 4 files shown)


hiod:
Nice work @imachug! https://github.com/HelloZeroNet/ZeroNet/issues/2273

purplesyringa:
@HelloZeroNet Please take a look at [this ZeroTalk thread](http://127.0.0.1:43110/Talk.ZeroNetwork.bit/?Topic:1578976152_17FotRRc6msHBzM2jrs6tQCmpLaXi1kHd5). Do you know what could cause that problem?

HelloZeroNet:
I just downloaded the latest snap and it does have the library in it, so it must be some import problem: Looks like the snap package using a custom launcher.py that does not adds src/lib to the sys.path
https://github.com/ZeroNerds/zeronet-snap/blob/master/src/scripts/launcher.py#L142
So I think `from lib import bencode_open` should fix the problem, but I will try to reproduce/test tomorrow.

HelloZeroNet:
I was able to reproduce and fix the problem in Rev4399: https://github.com/HelloZeroNet/ZeroNet/commit/a16d55c86394a514c4115c9f16d8145f3f4c1d98

For some reason for me the snap package drops permission error when tries to bind the ports, but it could be something with my configuration or snapd is not fully compatible with VM/Debian

PR #2371, opened 2019-12-20
Spanish and Hungarian translation update.

decentralizedauthority:


decentralizedauthority:
Travis fails nearly all the time and is kind of annoying.

PR #2385, opened 2020-01-05
Compression support & fixes

purplesyringa:
1. Fix `.tar.bz2` archive support.
2. Send `Content-Encoding` header for `.gz` and `.br` files. This allows compressing files and using them directly, like `<script src="main.js.gz">`.

### DIFF
--- a/plugins/FilePack/FilePackPlugin.py
+++ b/plugins/FilePack/FilePackPlugin.py
@@ -21,9 +21,6 @@ def openArchive(archive_path, file_obj=None):
         if archive_path.endswith("tar.gz"):
             import tarfile
             archive_cache[archive_path] = tarfile.open(archive_path, fileobj=file_obj, mode="r:gz")
-        elif archive_path.endswith("tar.bz2"):
-            import tarfile
-            archive_cache[archive_path] = tarfile.open(archive_path, fileobj=file_obj, mode="r:bz2")
         else:
             import zipfile
             archive_cache[archive_path] = zipfile.ZipFile(file_obj or archive_path)
@@ -48,7 +45,7 @@ class UiRequestPlugin(object):
             file_obj = None
             path_parts = self.parsePath(path)
             file_path = "%s/%s/%s" % (config.data_dir, path_parts["address"], path_parts["inner_path"])
-            match = re.match("^(.*\.(?:tar.gz|tar.bz2|zip))/(.*)", file_path)
+            match = re.match("^(.*\.(?:tar.gz|zip))/(.*)", file_path)
             archive_path, path_within = match.groups()
             if archive_path not in archive_cache:
                 site = self.server.site_manager.get(path_parts["address"])
@@ -102,7 +99,7 @@ class UiRequestPlugin(object):
 class SiteStoragePlugin(object):
     def isFile(self, inner_path):
         if ".zip/" in inner_path or ".tar.gz/" in inner_path:
-            match = re.match("^(.*\.(?:tar.gz|tar.bz2|zip))/(.*)", inner_path)
+            match = re.match("^(.*\.(?:tar.gz|zip))/(.*)", inner_path)
             archive_inner_path, path_within = match.groups()
             return super(SiteStoragePlugin, self).isFile(archive_inner_path)
         else:
@@ -130,7 +127,7 @@ class SiteStoragePlugin(object):
 
     def walk(self, inner_path, *args, **kwags):
         if ".zip" in inner_path or ".tar.gz" in inner_path:
-            match = re.match("^(.*\.(?:tar.gz|tar.bz2|zip))(.*)", inner_path)
+            match = re.match("^(.*\.(?:tar.gz|zip))(.*)", inner_path)
             archive_inner_path, path_within = match.groups()
             archive = self.openArchive(archive_inner_path)
             path_within = path_within.lstrip("/")
@@ -154,7 +151,7 @@ class SiteStoragePlugin(object):
 
     def list(self, inner_path, *args, **kwags):
         if ".zip" in inner_path or ".tar.gz" in inner_path:
-            match = re.match("^(.*\.(?:tar.gz|tar.bz2|zip))(.*)", inner_path)
+            match = re.match("^(.*\.(?:tar.gz|zip))(.*)", inner_path)
             archive_inner_path, path_within = match.groups()
             archive = self.openArchive(archive_inner_path)
             path_within = path_within.lstrip("/")
@@ -181,7 +178,7 @@ class SiteStoragePlugin(object):
 
     def read(self, inner_path, mode="rb", **kwargs):
         if ".zip/" in inner_path or ".tar.gz/" in inner_path:
-            match = re.match("^(.*\.(?:tar.gz|tar.bz2|zip))(.*)", inner_path)
+            match = re.match("^(.*\.(?:tar.gz|zip))(.*)", inner_path)
             archive_inner_path, path_within = match.groups()
             archive = self.openArchive(archive_inner_path)
             path_within = path_within.lstrip("/")
--- a/src/Ui/UiRequest.py
+++ b/src/Ui/UiRequest.py
@@ -707,7 +707,7 @@ class UiRequest(object):
         return block
 
     # Stream a file to client
-    def actionFile(self, file_path, block_size=64 * 1024, send_header=True, header_length=True, header_noscript=False, header_allow_ajax=False, file_size=None, file_obj=None, path_parts=None):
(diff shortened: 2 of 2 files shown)


Crypto-Anarchist:
This commit should be not merged! Fixing the "archive" support is fine but to mess with UiRequest.py just to serve your preprecompressed files are absurd. Why would we need such thing in the main repository? Can't you keep it to yourself?

purplesyringa:
Uh, I can't. I was looking for a way to optimize network usage and noticed that compressing my JS files would be a simple yet efficent solution. For example, one of my sites takes about 1.6 MiB incompessed and 400 KiB compessed.

Crypto-Anarchist:
I do understand you intention but it is wrong to assume that everyone will precompress they resources. Can't you use a reverse proxy for example Apache with DEFLATE? Otherwise this should be a plugin which serves resources compressed.

purplesyringa:
You didn't understand me. This optimization is not for proxies, this is for local clients. I (as a site owner) would compress my JS code to `main.js.gz` and use it like that.

Crypto-Anarchist:
> the client would send an Accept-Encoding: gzip request header, and then the server may decide to compress (if not already) and send a Content-Encoding: gzip and optionally Transfer-Encoding: chunked response header.

https://stackoverflow.com/questions/11641923/transfer-encoding-gzip-vs-content-encoding-gzip

Crypto-Anarchist:
@imachug 
```
For example,

     Transfer-Encoding: gzip, chunked

   indicates that the payload body has been compressed using the gzip
   coding and then chunked using the chunked coding while forming the
   message body.
```
https://tools.ietf.org/html/rfc7230#section-3.3.1

Crypto-Anarchist:
Python's urllib will do the unzipping automatically if you implement Transfer-Encoding.

purplesyringa:
Based on [this SO thread](https://stackoverflow.com/a/24425286), it looks like Transfer-Encoding refers to *transport* layer and Content-Encoding is *payload*-related. This means that we should use Content-Encoding for files which were precompressed, and we might use Transfer-Encoding as an optimization in case the client supports it. This PR only handles the first usecase.

Crypto-Anarchist:
`Transfer-Encoding` and `Content-Encoding` should **never** be used together! 
`Content-Encoding` is needed if you want to allow clients to **save the** `.gz` **file**,  otherwise the`Transfer-Encoding` should be used!

HelloZeroNet:
I would recommend to use `<script src="any.zip/any.js"` instead, because this way eg. tar.gz files became .tar files when they got downloaded from the browser.

purplesyringa:
@Crypto-Anarchist Right, that's **exactly** why I used Content-Encoding. I want the browser to save .js.gz and only decompress it in case use it in `<script>` tag and such.

purplesyringa:
@HelloZeroNet I'm wondering if we can add some flag to files (e.g. a query string such as `?decompress`) because gzip'ing files is really common. It removes zip overhead, compresses files a bit better (I haven't measured that but I think 20% file size decrease is likely) and is also widely supported by bundlers such as webpack.

HelloZeroNet:
I just did some tests between tar vs tar.gz vs zip and the overhead/difference is minimal (less than 100bytes)

purplesyringa:
Hm. Anyway, bundler support (as well as the fact that it's really common) idea still stands.

Crypto-Anarchist:
Ref: https://github.com/http-party/http-server/issues/300
Ping: @BigBlueHat Sorry for the ping BigBlueHat, can you comment on this topic?

HelloZeroNet:
`?zeronet-transfer-encoding=gzip` / `?zeronet-content-encoding=br` could be a good solution and probably it should be in the core and not in the plugin as it does not do any unpacking and if someone disables the plugin could also make future sites unusable.
Other solution is to add support .gz files the same way as we have tar.gz files and the content could be referenced as `src="any.js.gz/any.js"`, but this way the content will be served as cleartext (proxies would not benefit from the compression)
The tar.bz support got removed earlier because it's pretty slow compression and crashed the Android client.

Crypto-Anarchist:
I still think that using `Transfer-Encoding` is good and more beneficial than hacking ZeroNet itself.

purplesyringa:
> @imachug I think you can use Transfer-Encoding and compress your resources on-the-fly.

This is **not** for on-the-fly compression, it's to minimize network resource usage, it's to optimize peer-to-peer communications.

purplesyringa:
>  The tar.bz support got removed earlier because it's pretty slow compression and crashed the Android client.

Got it, I'll remove all the instances of `bz` then, ok?

purplesyringa:
> I still think that using Transfer-Encoding is good and more beneficial than hacking ZeroNet itself.

This is **not for proxies**, that's for all the clients. I want this change to propagate the whole network before launching my service.

Crypto-Anarchist:
Alternatively we can switch from HTTP 1.1 to HTTP 2.0 and than you can use Push or Early Hint to "minimize network resource usage, or to optimize peer-to-peer communications."

purplesyringa:
What don't you understand in "peer-to-peer communications"? It's to minimize ZeroNet protocol bandwidth, not HTTP.

purplesyringa:
Here's my usecase. The problem is that ZeroNet tends to share big files worse than several small files, that's why I split my scripts into several parts. Publishing compressed files would help me minimize file count and probably abandon splitting at all.

purplesyringa:
Was that enough to prove my point? @Crypto-Anarchist

Crypto-Anarchist:
Not like you can save a lot by compressing with gzip or zip.... BitTorrent need seeders if nobody seeds your big file than of course sharing it is problematic. I do share big files using ZeroNet (videos well over 1Gb) and because I have various nodes which stores the seeded content it is pretty fast. When someone request it. 

I can't use compression to share for example a 1.7Gb video. Tar or zip would save nothing.. in the other hand your javascript probably not soooo big.

purplesyringa:
> Not like you can save a lot by compressing with gzip or zip

gzip has 75% compression ratio for my scripts

> in the other hand your javascript probably not soooo big.

It's 1.6 MiB right now and I haven't implemented a half of what I wanted.

Crypto-Anarchist:
@imachug https://diracdeltas.github.io/blog/backdooring-js/ 

"Don’t minify/compress server-side code unless you have to, and make sure you run browser tests/scans against code post-minification."

Crypto-Anarchist:
> It's 1.6 MiB right now and I haven't implemented a half of what I wanted.

Can I ask what are you up to? :thinking:

purplesyringa:
> Can I ask what are you up to? 🤔

I don't want to release any details right now, sorry.

Crypto-Anarchist:
@HelloZeroNet what do you think? @imachug did you tested this commit? Do we really need this in the main repository? Is hard to understand what are you doing imachug, I think this should be not merged until you did not finished the your work. When you done you can present it and than we can merge this. I think no need for rushing.

purplesyringa:
First, this would be helpful for other sites as well. Second, I want to get this merged so that as many people as possible had an updated version at the release moment.

Crypto-Anarchist:
Okay, but we should not merge anything which is promising something mysterious thing in the future. For example, the license issue is more important than this commit.

purplesyringa:
Do you **really** think software is less important than license? Anyways, do you want me to show you a usecase or what? I think I've explained it well enough.

Crypto-Anarchist:
Yes, I **really** think. Everyone would be happy if you show the use case.

purplesyringa:
I've shown it already: big files get propagated slowly. Compressing scripts would a) speed that up b) minimize required bandwidth

Crypto-Anarchist:
I hope you get it merged. :+1:

purplesyringa:
There we go, I added `?zeronet_content_encoding=...` query string handling to core. E.g.: `<script src="main.js.gz?zeronet_content_encoding=gzip"></script>`

HelloZeroNet:
- there is a self.get in the ui_request, so you don't have to parse it yourself
- probably adding a gzip/br whitelist as accepted values would be better to avoid possible security issues
- bz2/xz compression in the test still make sens to know if there is a platform that is not compatible with that compression.

purplesyringa:
There we go, sorry for endless problems.

purplesyringa:
(misclick)

inb4: gzipped HTML files are sandboxed correctly.

purplesyringa:
Thanks.

BigBlueHat:
> Ref: [http-party/http-server#300](https://github.com/http-party/http-server/issues/300)
> Ping: @BigBlueHat Sorry for the ping BigBlueHat, can you comment on this topic?

No worries! Sorry I only just saw this ping. Did you need something from http-server? Or just asking for input?

PR #2388, opened 2020-01-06
Update wrapper.html

Crypto-Anarchist:


Crypto-Anarchist:
Would be nice to use languages and directions correctly. Farsi would look like <html lang="fa" dir="rtl"> unsupported messages can be tagged up with lang="en" dir="rtl" (because they are English) or we need to translate those messages as well.

Crypto-Anarchist:
@HelloZeroNet many `class=` use the `'` instead of `"` is this was intentional?

imachug:
@Crypto-Anarchist Out of interest: why do you use `&#x02BC;` instead of `'`?

Crypto-Anarchist:
Unicode version 1 has endorsed `&#x02BC;` as apostrophe. 

Here is the difference:
# ʼ
# ' 

**This is especially important for word processors!** 
`&#x02BC;` U+02BC (MODIFIER LETTER APOSTROPHE) should be used for apostrophe and not `'` U+2019 (RIGHT SINGLE QUOTATION MARK). 

You can see in they name what are they. ' is a right quotation mark while &#x02BC; is the original and **correct apostrophe**. 

@imachug you can read about this at https://tedclancy.wordpress.com/2015/06/03/which-unicode-character-should-represent-the-english-apostrophe-and-why-the-unicode-committee-is-very-wrong/

imachug:
Huh, that's interesting. Thanks for the PR.

PR #2393, opened 2020-01-11
Update ContentManager.py

Crypto-Anarchist:
ZeroNet version in content.json is not needed for anything. If someone want to know the version it can be seen in various locations already. I removed also ZeroNet_ from the title.

### DIFF
--- a/src/Content/ContentManager.py
+++ b/src/Content/ContentManager.py
@@ -677,7 +677,7 @@ class ContentManager(object):
                 content = {"files": {}, "signs": {}}  # Default content.json
 
             if inner_path == "content.json":  # It's the root content.json, add some more fields
-                content["title"] = "%s - ZeroNet_" % self.site.address
+                content["title"] = "%s" % self.site.address
                 content["description"] = ""
                 content["signs_required"] = 1
                 content["ignore"] = ""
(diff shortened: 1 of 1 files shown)


Crypto-Anarchist:
@imachug

src/Content/ContentManager.py
```diff
@@ -677,7 +677,8 @@ def sign(self, inner_path="content.json", privatekey=None, filewrite=True, updat
                 content = {"files": {}, "signs": {}}  # Default content.json
 
             if inner_path == "content.json":  # It's the root content.json, add some more fields
-                content["title"] = "%s - ZeroNet_" % self.site.address
+                content["title"] = "%s" % self.site.address
+                content["optional"] = "(video/.*|audio/.*)"
```
purplesyringa: Is this change necessary? It looks like a backward compatibility breaker and also I really don't understand why specific directory names should be hardcoded.

imachug:
Removing ZeroNet_ sounds good to me.

src/Content/ContentManager.py
Crypto-Anarchist: No, is not necessary.

purplesyringa: Can you remove this line then please?

filips123:
`zeronet_version` contains version of ZeroNet that was used for signing and publishing site, not version of current ZeroNet.

It is currently not used, but it could be in the future if `content.json` format is changed to provide backward compatibility with old versions based on their `zeronet_version`.

src/Content/ContentManager.py
Crypto-Anarchist: Done.

PR #2395, opened 2020-01-14
Update UiConfigPlugin.py

Crypto-Anarchist:


Crypto-Anarchist:
Maybe I'm wrong but I believe we don't need this in the end. UiRequest.py has the encoding correctly stated for these MIME types. 

@imachug @filips123 Any comment?

Crypto-Anarchist:
`@PluginManager.registerTo("UiRequest")
`

imachug:
You're wrong. This change is will introduce a type error.

Crypto-Anarchist:
@imachug yeah, you are right. 
I changed `utf8` to `utf-8` as it is described at https://docs.python.org/3/howto/unicode.html

imachug:
> but I strongly support to use utf-8 instead.

Is there a reason for this?

Crypto-Anarchist:
I just updated the previous comment.

filips123:
> Some common encodings can bypass the codecs lookup machinery to improve performance. These optimization opportunities are only recognized by CPython for a limited set of (case insensitive) aliases: utf-8, utf8, latin-1, latin1, iso-8859-1, iso8859-1, mbcs (Windows only), ascii, us-ascii, utf-16, utf16, utf-32, utf32, and the same using underscores instead of dashes.

Crypto-Anarchist:
Yeah I seen that too @filips123 
> the same using underscores instead of dashes

filips123:
Yes, but as page says, it's the same if you use `utf8`, `utf-8` or `utf_8`. So the change is not actually needed.

Crypto-Anarchist:
It is not the same! `utf8` and `utf-8` is an alias for `uft_8`!

filips123:
> Some common encodings can bypass the codecs lookup machinery to improve performance

Crypto-Anarchist:
@filips123 yes, exactly that is why we should not use any alias!

filips123:
Why? If it is written that it it is the same?

Crypto-Anarchist:
Please read the previous comment. https://github.com/HelloZeroNet/ZeroNet/pull/2395#issuecomment-574584378

Crypto-Anarchist:
@imachug can you tell to @filips123 what is the difference between a codec and the alias of the codec? Also explain to him how this will speed up execution.

imachug:
@Crypto-Anarchist According to the docs, `utf8` and other commonly used aliases are as performant as `utf_8`.

Crypto-Anarchist:
I going to cleanup the entire source code of zeronet from aliases. I find and replace all `utf8` with `utf_8`. Python docs is clear about which is the fastest and safest.

imachug:
No it's not, RTFM.

imachug:
https://github.com/HelloZeroNet/ZeroNet/pull/2395#issuecomment-574581530 According to this, utf8 is as fast as utf-8 and utf_8.

Crypto-Anarchist:
@imachug that is a reference to CPython.

Crypto-Anarchist:
@filips123 :laughing: :+1:

imachug:
Said a person that upvoted a couple of his own comments.

Crypto-Anarchist:
> Said a person that upvoted a couple of his own comments.

Just because you @imachug are blind and don't you see that the quote what @filips123 wrote is related to CPython, you don't need to talk off topic. I get it you are angry. You can't think that you and others in this repository are always right. 

If you and filips can't post any constructive comment I kindly ask to cease posting anything.

filips123:
> is related to CPython

So which implementation would you use? Other implementations probably also bypass lookup for common aliases.

And just that you know, "Using alternative aliases for these encodings may result in slower execution." is also just for CPython!

> If you and filips can't post any constructive comment I kindly ask to cease posting anything.

Sure, just a few quotes by you:

> Than go and fuck yourself!
> You are a joke! Microsoft sucker!

And many more, but they were already deleted.

Crypto-Anarchist:
@filips123 Far as I know there is no lookup for encoding if you not using aliases, so you are again wrong. `utf_8` is **not an alias** and so it is execute faster.

filips123:
Documentation specifies that it is the same for common codecs and aliases if you use alias or not. So you are wrong...

Crypto-Anarchist:
I'm not going to repeat myself just because you don't speak English and don't understand even the documentation. `utf_8` is **not an alias** and so it is executes faster than using your stupid aliases which introduce additional lookup (which in some cases also fails).

filips123:
You don't understand English! Documentation says that common aliases bypass lookup so they are the same as usage without alias.

Crypto-Anarchist:
@filips123 **Only in the case of CPython!!!!!!!!!!!!** You and @imachug are so retarded that I'm crying in my sleep.

filips123:
And it is also slower only in CPython!!! And other common implementation also bypass lookup for common aliases.

And again, which implementation would you use if not CPython? It is the most popular implementation and it is also included in ZeroBundle.

And some other implementations will run ZeroNet slower, but not because of stupid UTF 8 codec, but because they can't execute some C/C++ extensions which ZeroNet relies on so fast as CPython, as well as other reasons.

filips123:
> You and @imachug are so retarded that I'm crying in my sleep.

Remainder:

> If you [...] can't post any constructive comment I kindly ask to cease posting.

Crypto-Anarchist:
> And it is also slower only in CPython!!! And other common implementation also bypass lookup for common aliases.
> 
> And again, which implementation would you use if not CPython? It is the most popular implementation and it is also included in ZeroBundle.
> 
> And some other implementations will run ZeroNet slower, but not because of stupid UTF 8 codec, but because they can't execute some C/C++ extensions which ZeroNet relies on so fast as CPython, as well as other reasons.

The bottom line is that using aliases you risk failure and slower execution. **Not using any alias but directly stating the encoding will speed up execution and it is fail-safe!** Period.

filips123:
Surely you have some proof about that? And even if there are any differences in performance, those half a millisecond will have big impact on performance?

Crypto-Anarchist:
See: https://www.python.org/dev/peps/pep-0263/

```
#!/usr/bin/python
# -*- coding: <encoding name> -*-
```

Regex: 
**^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+)**

Show me where you see here 'utf8' and tell me how this matching the regex from above.

filips123:
1. This is something completely different...
2. Your regex is not correct...
3. If you are using the correct regex, it [matches](https://regex101.com/r/sl9zyu/1):

match

Crypto-Anarchist:
Don't you see is not my regex? How can be someone so blind like you... :roll_eyes: 

> @filips123 See: https://www.python.org/dev/peps/pep-0263/
> 
> ```
> #!/usr/bin/python
> # -*- coding: <encoding name> -*-
> ```
> 
> Regex:
> **^[ \t\f]_#._?coding[:=][ \t]*([-_.a-zA-Z0-9]+)**
> 
> Show me where you see here 'utf8' and tell me how this matching the regex from above.
> 
> @HelloZeroNet @imachug : https://savannah.gnu.org/bugs/?50357#comment0 This is why aliases are bad! I suggest not to use any alias in ZeroNet's source code.
> 
> https://savannah.gnu.org/bugs/?50357#comment5
> 
> > The standards don't know about an alias "utf8".

filips123:
?

filips123:
Again, PEP 263 (source code encodings) is completely different from encoding strings.

Crypto-Anarchist:
WTF is this? :anger: 
> match

Show me a Python code which understand your regex and encoding with 'utf8' fu**g retard.

Crypto-Anarchist:
> Again, PEP 263 (source code encodings) is completely different from encoding strings.

Yeah, you think I editing the source code on the fly? Your strings can be destroyed when I saving the file because you disregard the standards. Easy, fast and trashy work. Congratulations.

filips123:
WTF are you talking? Do you thing editor will somehow delete static strings in source code because of something?

filips123:
> Show me a Python code which understand your regex and encoding with 'utf8' fu**g retard.

Sure:

```py
import re

regex = r'^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+)'
search = '# -*- coding: utf8 -*-'

result = re.search(regex, search)
print(result[1])

# Will output 'utf8'
# Is it so hard?
```

filips123:
> Yeah, you think I editing the source code on the fly? 

You editing the source code on the fly?

> Your strings can be destroyed when I saving the file because you disregarding the standards.

What? Which strings? And how can they be "destroyed"?

Crypto-Anarchist:
https://bugs.python.org/issue4487

> When using MIME email package you can specify "utf8" as the encoding. It will be accepted but it is not rendered correctly in some MUA.

Response: 
> **I recommend accepting this patch or a slightly fancier one that would also do "utf_8".**

filips123:
Again, this is not related to encoding `str` objects! And issue you linked is 10 years old and already fixed.

Crypto-Anarchist:
> Again, this is not related to encoding `str` objects! And issue you linked is 10 years old and already fixed.

If you read it than you would understand what and how... 

> The attached patch instead uses the codecs module as a fallback if the charset name does not appear in the email package's ALIASES or CHARSETS tables.

In this case if there is no alias than it will fallback to the **STANDARD**! 
_uft8 is an alias_ not a CODEC like **uft_8**!

filips123:
And?

Crypto-Anarchist:
and zeronet should follow standards not trashing the source code with aliases just because faster to write an alias 'utf8' than the standard codec 'utf_8'. Is so difficult to write and underline? :roll_eyes:

filips123:
And? If there are no differences when using alias or standard codec as alias is converted to standard codec?

Crypto-Anarchist:
There is a huge difference between aliases and codecs! One is standard the other is just for retards like you.

filips123:
# 🤣

Crypto-Anarchist:
@filips123 can you send me an email to russiansecretservice@yandex.ru? I have a job for you. I pay you 1,500 USD in Bitcoin if you write me a small program. :P

Crypto-Anarchist:
you are boring so you just argue all the time... can't be more shameful... you contributed about 8 times to ZeroNet think about that...

imachug:
You're making PRs just to contribute. It doesn't matter what to contribute, you just want your name in the git log. The upstream code works and as fast as the code you're appealing for. I don't think there's a reason to change something that isn't broken and won't be in the future.

PR #2408, opened 2020-02-02
please use, from @caseyjwach

cjwach1:
when will #bitcoin add to the list of important projects barcode scanning like Wechat in China? It's literally Tencent's Kudegra to CC and debit cards, much less cash... let's start now.


filips123:
?

HelloZeroNet:
I think it's not relevant for us, because we don't use bitcoin as payment system

PR #2421, opened 2020-02-13
Added Custom Openssl Path for Native Clients

canewsin:
This Parameter helpful where openssl path is not fixed always, we can also use this to reduce code verbosity by providing other like these and provide them as parameter
```
            else:
                self.openssl_bin = "openssl"
```


canewsin:
closed in favour of https://github.com/HelloZeroNet/ZeroNet/pull/2422 due unmerged upstream changes and further fixes

PR #2338, opened 2019-12-02
Switch to sslcrypto for cryptography tasks

purplesyringa:
[sslcrypto](https://github.com/imachug/sslcrypto) is a new MIT-licensed library I've recently made. It supports AES and ECC. I originally made it to get rid of GPLv3 pyelliptic, but interestingly it turned out even more useful.

First, it replaces pyelliptic, pybitcointools and a part of CryptMessage so there's less code involved. Second, it supports both OpenSSL backend and pure-Python fallback both for AES and ECC (pyelliptic doesn't support pure-Python ECC so CryptMessage didn't work on some machines). Third, it's licensed under MIT (I can add any other license you choose) so we wouldn't have problems if we wanted to re-license a part of ZeroNet code under some lax license. Fourth, it supports *all* OpenSSL versions since 0.9.8 (other cryptography libraries support a lot less). Fifth, its fallback implementation is a bit safer than pybitcointools because it fixes a vulnerability (I don't want to name it here to avoid possible private key leaks).

Sixth, it's fast. Plain performance tests say that it's about as fast as pyelliptic (both for AES and ECC) and that its fallback implementation is 1.5 times faster than pybitcointools (ECIES only, ECDSA is about the same). However, some tests I ran on ZeroMail show that message decryption is about 2 or 3 times faster than the current version (I've only run tests on single-thread model because more threads are slower for me for some reason both for sslcrypto and your version).

Additionally, sslcrypto is [tested](https://github.com/imachug/sslcrypto/actions) on Python 3.4 to 3.8 with OpenSSL 0.9.8, 1.0.0, 1.0.1, 1.0.2, 1.1.0 and 1.1.1.

/cc @filips123

---

This PR closes #1974, closes #917 and closes #993.

PR #2422, opened 2020-02-13
Added Custom Openssl Path for Native Clients and Fixed Tracker files provided are not loading on some systems

canewsin:
This Parameter helpful where openssl path is not fixed always, we can also use this to reduce code verbosity by providing other like these and provide them as parameter
```
            else:
                self.openssl_bin = "openssl"
```

Also Fixed Tracker files provided are not loading on some systems like Android client where path storage access is restricted

### DIFF
--- a/src/Config.py
+++ b/src/Config.py
@@ -31,6 +31,7 @@ class Config(object):
 
         self.config_file = self.start_dir + "/zeronet.conf"
         self.data_dir = self.start_dir + "/data"
+        self.openssl_path = "default"
         self.log_dir = self.start_dir + "/log"
 
         self.trackers_file = False
@@ -108,6 +109,7 @@ class Config(object):
 
         config_file = self.start_dir + "/zeronet.conf"
         data_dir = self.start_dir + "/data"
+        openssl_path = "default"
         log_dir = self.start_dir + "/log"
 
         ip_local = ["127.0.0.1", "::1"]
@@ -259,6 +261,7 @@ class Config(object):
         self.parser.add_argument('--ip_external', help='Set reported external ip (tested on start if None)', metavar='ip', nargs='*')
         self.parser.add_argument('--offline', help='Disable network communication', action='store_true')
 
+        self.parser.add_argument('--openssl_path', help='Custom Path to OpenSSL Binary', default=openssl_path, metavar="path")
         self.parser.add_argument('--disable_udp', help='Disable UDP connections', action='store_true')
         self.parser.add_argument('--proxy', help='Socks proxy address', metavar='ip:port')
         self.parser.add_argument('--bind', help='Bind outgoing sockets to this address', metavar='ip')
@@ -321,6 +324,9 @@ class Config(object):
                     trackers_file_path = trackers_file
                 elif trackers_file.startswith("{data_dir}"):  # Relative to data_dir
                     trackers_file_path = trackers_file.replace("{data_dir}", self.data_dir)
+                elif trackers_file.startswith("data/"):
+                    # Relative to data_dir Useful for Unix Type Systems and Android Clients
+                    trackers_file_path = trackers_file.replace("data", self.data_dir)
                 else:  # Relative to zeronet.py
                     trackers_file_path = self.start_dir + "/" + trackers_file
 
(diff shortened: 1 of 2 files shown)


HelloZeroNet:
Thanks, but there is a `{data_dir}` variable in trackers file path that should solve this problem:
https://github.com/HelloZeroNet/ZeroNet/blob/py3/src/Config.py#L322

canewsin:
@HelloZeroNet
data_dir variable is start_dir + "data_dir" 
default start_dir is "." 
On Android "./data" Is not a valid path thus I changed it to acceptable one without breaking the code.

HelloZeroNet:
You can change the path using the data_dir configuration value.
If you have {data_dir} in your trackers_file value, then it will follow the location.

It does the same as your modification, but replaces "{data_dir}" and not all "data" part in the variable:
```py
...
                elif trackers_file.startswith("{data_dir}"):  # Relative to data_dir
                    trackers_file_path = trackers_file.replace("{data_dir}", self.data_dir)
                elif trackers_file.startswith("data/"):
                    # Relative to data_dir Useful for Unix Type Systems and Android Clients
                    trackers_file_path = trackers_file.replace("data", self.data_dir)
...
```

russianagent:
Well-well. I use Termux on Android and openssl is a big problem. I can't run zeronet from terminal because of this. Maybe we should ship openssl with zeronet. Also `pip install -r requirements.txt` has some issues with the versions... but as I said only OpenSSL has problem and this is why I can't run zeronet from Termux. @imachug @HelloZeroNet @canewsin

russianagent:
If @imachug was replaced able to write up https://github.com/HelloZeroNet/ZeroNet/pull/2338 than I guess you @imachug can also write a library for openssl and ship that one with ZeroNet. 

I can write a new zeronet application which is not based on Kivy. @HelloZeroNet

canewsin:
@russianagent I was already on works of Android client which is not based on kivy

russianagent:
@canewsin I know what are you up to! 
# :stuck_out_tongue_closed_eyes:

canewsin:
@HelloZeroNet
No what I was saying is take for example
"data/trackers.json" loop will starts like first it checks for data_dir, unless configured it will remain as "./data/xxx" on linux systems it will work but not on Android, "./" is not valid path on Android, hope you understand what I was talking.

russianagent:
By the way ZeroNet-Kivy sucks! We need to fix the OpenSSL thing in ZeroNet possibly ship with the source code and than you @canewsin can run ZeroNet from Termux which have a python interpreter. Need to install more packages than it is stated in the requirements.txt but is not a problem. I will fix that in my fork in the coming days.

russianagent:
[time] CryptConnectionManager RSA ECC SSL CAcert generation failed, CAcert or CAkey files not exist. (/data/data/com.termux/files/usr/bin/sh: 1: openssl: not found)
[time] TorManager Disabling Tor, because error while accessing Tor proxy at port 127.0.0.1:9050: Error ([Errno 111] Connection refused)

if this solved than you can run ZeroNet using Termux on Android. @imachug @filips123 @HelloZeroNet OpenSSL sh*pile

canewsin:
@russianagent idk why you are getting openssl error, because I almost installed 20+ times via termux on various phones and emulators never faced such an issue, since this is pull request use another issue or medium to discuss your problem, also mention there which proceedure did you followed to install ZeroNet there so I can help you, I tried same thing now on my phone no openssl error occurred.

russianagent:
I will reinstall it, maybe that will solve this.
ZeroNet is running perfectly I can visit 127.0.0.1:43110 on Tor Browser for Android but when I try to access one of my site it will not find any peer which is absurd... I running many ZeroNet peers...

canewsin:
@HelloZeroNet Could someone look into the pull req because, i already answered the queries above !

HelloZeroNet:
I'm not familiar with Android restrictions, but if ./ is not a valid path there, then probably it would be better to change the getStartDir function, because eg. the log dir is also `./log` by default.

I think the best would be adding a `--start_dir` configuration value that allows to override the working dir location manually.

canewsin:
> I'm not familiar with Android restrictions, but if ./ is not a valid path there, then probably it would be better to change the getStartDir function, because eg. the log dir is also `./log` by default.
> 
> I think the best would be adding a `--start_dir` configuration value that allows to override the working dir location manually.

will do ...

canewsin:
since new changes are requested and also there are upstream changes, closing this pull in favour of https://github.com/HelloZeroNet/ZeroNet/pull/2426, which contains latest upstream changes.

PR #2426, opened 2020-02-18
Added Custom Openssl Path for Native Clients and start_dir config

canewsin:
This Parameter helpful where openssl path is not fixed always, we can also use this to reduce code verbosity by providing like these and provide them as parameter

            else:
                self.openssl_bin = "openssl"
Also Added Custom start_dir config option since android path issue of not valid "./" path, where files via provided path are not loading on some systems like Android client.

for more detailed conversation see pull request [#2422](https://github.com/HelloZeroNet/ZeroNet/pull/2422)

### DIFF
--- a/src/Config.py
+++ b/src/Config.py
@@ -29,6 +29,7 @@ class Config(object):
         ])
         self.start_dir = self.getStartDir()
 
+        self.openssl_path = "default"
         self.config_file = self.start_dir + "/zeronet.conf"
         self.data_dir = self.start_dir + "/data"
         self.log_dir = self.start_dir + "/log"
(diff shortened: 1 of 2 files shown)


canewsin:
@HelloZeroNet done for requested changes

HelloZeroNet:
Hm for some reason the test are failing...probably not related to the change, checking it

canewsin:
@HelloZeroNet yeah i do see that in previous pull req https://github.com/HelloZeroNet/ZeroNet/pull/2421 when i removed the pull req and pushed the changes to https://github.com/HelloZeroNet/ZeroNet/pull/2422 errors are gone, may be there is a build config issue with travis

PR #2431, opened 2020-02-20
Fixes Tracker Files not accessed when there is custom data dir

canewsin:
If we have string form of data dir like "data/trackers.json" and custom data conf applied to zeronet, there is no condition in loop to catch it. this will cause zeronet defaults to last default option, which in return is not valid path.

### DIFF
--- a/src/Config.py
+++ b/src/Config.py
@@ -326,6 +326,9 @@ class Config(object):
                     trackers_file_path = trackers_file
                 elif trackers_file.startswith("{data_dir}"):  # Relative to data_dir
                     trackers_file_path = trackers_file.replace("{data_dir}", self.data_dir)
+                elif trackers_file.startswith("data/"):  # Relative to data_dir in string form.
+                    # This is Required if we use custom data dir
+                    trackers_file_path = trackers_file.replace("data", self.data_dir)
                 else:  # Relative to zeronet.py
                     trackers_file_path = self.start_dir + "/" + trackers_file
 


canewsin:
@HelloZeroNet

HelloZeroNet:
It should be fixed by https://github.com/HelloZeroNet/ZeroNet/commit/a9c75a3146c05dce260165a190b6bc5b7dd3edb4#diff-8fddfd9ebb27329fde8fa7eb72e7f796R58
in Rev4452

canewsin:
> It should be fixed by [a9c75a3#diff-8fddfd9ebb27329fde8fa7eb72e7f796R58](https://github.com/HelloZeroNet/ZeroNet/commit/a9c75a3146c05dce260165a190b6bc5b7dd3edb4#diff-8fddfd9ebb27329fde8fa7eb72e7f796R58)
> in Rev4452

No, how did it solve the problem, it still exists.

HelloZeroNet:
Here is how I tested:
 - Created a file: `start-test/data/trackers.txt` with the content `zero://new.tracker:15441`
 - Started the client: with `python3 zeronet.py --trackers_file data/trackers.txt --start_dir start-test`
 - The tracker `zero://new.tracker:15441` from the file got loaded from  `start-test/data/trackers.txt` file

canewsin:
> Here is how I tested:
> 
> * Created a file: `start-test/data/trackers.txt` with the content `zero://new.tracker:15441`
> * Started the client: with `python3 zeronet.py --trackers_file data/trackers.txt --start_dir start-test`
> * The tracker `zero://new.tracker:15441` from the file got loaded from  `start-test/data/trackers.txt` file

i was talking about modifiing trackers in Config Page of ZeroNet not cmd line option, also my data folder is not in start-dir so your example also breaks here.



> Here is how I tested:
> 
> * Created a file: `start-test/data/trackers.txt` with the content `zero://new.tracker:15441`
> * Started the client: with `python3 zeronet.py --trackers_file data/trackers.txt --start_dir start-test`
> * The tracker `zero://new.tracker:15441` from the file got loaded from  `start-test/data/trackers.txt` file

if i give --data_dir some/dir conf option to zeronet above will fail and in turn uses `start-test/data/trackers.txt` which results in file doesn't exist on that path.

canewsin:
For example follow this
For Example if have windows machine 
use this cmd to start zeronet
python zeronet.py "--data_dir "some-dir"
this will start zerenet
now go to http://127.0.0.1:43110/Config
in Trackers files section
provide your trackers file there eg : data/trackers.txt
 
this will give you an error 
> Error loading trackers file: [Errno 2] No such file or directory: './data/trackers.txt'

hope you get it.

canewsin:
@HelloZeroNet

HelloZeroNet:
yes, but it is intentional, if you want to follow the data dir, then you need to change it to `{data_dir}/trackers.txt`

canewsin:
@HelloZeroNet some platforms it will be easy to use, but on some like android
if you use public facing data dir you have to enter something like
> /storage/emulated/0/Android/data/in.canews.zeronet/files/data/trakers.txt

which is not convenient to user. 
instead if user use some thing like 
> data/trackers.txt
 we can simply assume data dir is zeronet data dir

also if user has more than 1 file to use, it will be more difficult 
> /storage/emulated/0/Android/data/in.canews.zeronet/files/data/trakers.txt
> /storage/emulated/0/Android/data/in.canews.zeronet/files/data/somefile.txt
> /storage/emulated/0/Android/data/in.canews.zeronet/files/data/another.txt
> /storage/emulated/0/Android/data/in.canews.zeronet/files/data/trakers0.txt
> /storage/emulated/0/Android/data/in.canews.zeronet/files/data/datafile.txt
> /storage/emulated/0/Android/data/in.canews.zeronet/files/data/trakers1.txt
> /storage/emulated/0/Android/data/in.canews.zeronet/files/data/trakers2.txt
> /storage/emulated/0/Android/data/in.canews.zeronet/files/data/trakers3.txt

canewsin:
for the worst if you want to use something like sychronite trackers you have to enter this long below
> /storage/emulated/0/Android/data/in.canews.zeronet/files/data/15CEFKBRHFfAP9rmL6hhLmHoXrrgmw4B5o/cache/1/_h_.html

if we allow data/ as valid path it will be reduced to 
> data/15CEFKBRHFfAP9rmL6hhLmHoXrrgmw4B5o/cache/1/_h_.html
> data/trackers.txt

HelloZeroNet:
If you use `--start_dir /storage/emulated/0/Android/`, then `data/trackers.txt` in trackers file will load `/storage/emulated/0/Android/data/trackers.txt` (that was the reason of adding start_dir argument)

Also:
Instead of `/storage/emulated/0/Android/data/in.canews.zeronet/files/data/trakers.txt` you can use `{data_dir}/in.canews.zeronet/files/data/trakers.txt` which is not much longer than`data/in.canews.zeronet/files/data/trakers.txt`

canewsin:
ok one minute, i was thinking below is data_dir variable
> {data_dir}

it's not, currently in my working language {} are used for resolving variables, so i thought it is data_dir variable, so its not a variable but it is a string.

canewsin:
Closing because we don't need it, as resolve already exists.

PR #2437, opened 2020-02-24
Allow uploading files via websocket

### LINKED ISSUE #2436: Stream big file upload
**Is your feature request related to a problem? Please describe.**
I need to upload a big file (say, 2 GiB) from a web app but I generate the data dynamically. I can't save the whole file to memory because it might not fit, so I'd like to stream it. It seems like using ReadableStream as request body is what I'm looking for but unfortunately no browsers support it.

**Describe the solution you'd like**
An API call similar to `bigfileUploadInit` to create a new WebSocket, not an HTTP stream. It would be used like this:
```javascript
const size = 1024 * 1024; // 1 MiB
const origin = location.href.replace(/(\:\/\/.*?)\/.*/, "$1").replace("http", "ws");
const {url} = await page.cmdp("bigfileUploadInitWebsocket", ["test.bin", size]);
const ws = new WebSocket(url.replace("{origin}", origin));
let pos = 0;
ws.onmessage = e => {
    if(e.data === "poll") {
        part = (new Uint8Array(1024)).buffer; // for example
        ws.send(part);
        pos += part.byteLength;
    }
};
```
Some non-intuitive details:
1. The backend should send `"poll"` message after it receives a packet, asking you to send a new packet. This is required because there's no way to control websocket send queue (as opposed to classic file upload when the browser handles it) so sending too much data might break the process. Sending packets only when the backend tells you is a decent workaround.
2. A `"poll"` message should be sent by the backend when the socket is open. Alternatively, the client could just listen to `open` websocket event but a `"poll"` message looks more consistent.

**Describe alternatives you've considered**
1. Asking the browser developers to allow using ReadableStream as request body. :man_shrugging: 
2. Using `fileWrite`: it uses too much memory.
3. Using the main websocket for file upload instead of a separate one. This is something to consider. I (intuitively) think that using a separate websocket would block the main one less but I'm not sure.

**Additional context**
[Standards](https://github.com/whatwg/fetch/pull/425)

purplesyringa:
Closes #2436

### DIFF
--- a/plugins/Bigfile/BigfilePlugin.py
+++ b/plugins/Bigfile/BigfilePlugin.py
@@ -61,13 +61,44 @@ class UiRequestPlugin(object):
         })
 
         self.readMultipartHeaders(self.env['wsgi.input'])  # Skip http headers
+        result = self.handleBigfileUpload(upload_info, self.env['wsgi.input'].read)
+        return json.dumps(result)
 
+    def actionBigfileUploadWebsocket(self):
+        ws = self.env.get("wsgi.websocket")
+
+        if not ws:
+            self.start_response("400 Bad Request", [])
+            return [b"Not a websocket request!"]
+
+        nonce = self.get.get("upload_nonce")
+        if nonce not in upload_nonces:
+            return self.error403("Upload nonce error.")
+
+        upload_info = upload_nonces[nonce]
+        del upload_nonces[nonce]
+
+        ws.send("poll")
+
+        buffer = b""
+        def read(size):
+            nonlocal buffer
+            while len(buffer) < size:
+                buffer += ws.receive()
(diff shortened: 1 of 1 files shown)


purplesyringa:
Thanks!

HelloZeroNet:
Great solution, it's more simple than I thought it's going to be. Maybe it would be better to merge actionBigfileUploadInitWebsocket and actionBigfileUploadInit functions to avoid code duplication and adding an optional `protocol=xhr|websocket` argument instead.

purplesyringa:
Looks like a good idea, I didn't notice the code duplication for some reason. Working on the PR.

purplesyringa:
Done, check #2438 please.

PR #2438, opened 2020-02-24
Avoid code duplication in bigfileUploadInit

purplesyringa:
See #2437.


purplesyringa:
There are no tests unfortunately but I've verified it manually and it works well.

purplesyringa:
Thx.

PR #2439, opened 2020-02-25
Switch from gevent-websocket to gevent-ws

purplesyringa:
Replacing the slow library with my own fast MIT-licensed library. I'm not sure if this will break bundled clients though, you might want to copy the source (https://github.com/imachug/gevent-ws) to the `lib` directory.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ rsa
 PySocks>=1.6.8
 pyasn1
 websocket_client
-gevent-websocket
+gevent-ws
 coincurve
 python-bitcoinlib
 maxminddb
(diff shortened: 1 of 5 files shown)


HelloZeroNet:
It will broke the bundles/updates from web interface, so supporting both is necessary. (looks easy to do)
Also: The run_application modification is necessary for error handling.

purplesyringa:
K, will fix.

purplesyringa:
I've restored `run_application` (refactored it a bit) and also added `gevent_ws` to `lib`. I think that supporting both libraries is not necessary because `gevent_ws` is way faster.

purplesyringa:
@HelloZeroNet

HelloZeroNet:
Thanks, I will test it and merge later today. Do you have numbers how much faster is it?

purplesyringa:
> Do you have numbers how much faster is it?

IIRC `gevent-websocket` can handle around 50 Mbps and my library can handle around 750 Mbps (just streaming zeroes).

HelloZeroNet:
Thanks, looking great!

PR #2449, opened 2020-02-28
Polish translation

krzotr:
I have added polish translation to `Trayicon` and `UiConfig` plugin. I also added some missing translations.


HelloZeroNet:
Thanks!

PR #2453, opened 2020-02-28
Fixed `Cache-Control` for .js and .css files - 10 minutes cache

krzotr:
Hello,

When sending request `GET http://127.0.0.1:43110/uimedia/all.css?rev=4458` the response contains `Cache-Control: no-cache, no-store, private, must-revalidate, max-age=0`.
It means `.css` and `.js` files are not cached by the browser

I fixed that problem, now receive: `Cache-Control: public, max-age=600`

* * *

I change the order of code, check the `content-type` (`application/javascript` or `text/css`), at the end add charset `utf-8.


HelloZeroNet:
Thanks!

PR #2454, opened 2020-03-03
Fix sslcrypto thread safety

purplesyringa:
A continuation of #2338.


src/lib/sslcrypto/openssl/library.py
```diff
@@ -0,0 +1,93 @@
...
+import sys
+import ctypes
+import ctypes.util
+
+
+# Disable false-positive _MEIPASS
+# pylint: disable=no-member,protected-access
+
+# Discover OpenSSL library
+def discover_paths():
```
canewsin, line 12: Recently there is new cmd line option for openssl lib
https://github.com/HelloZeroNet/ZeroNet/commit/1cc0ec3f31b3b8db4a534ec64ebe82629b81fa2e
i don't see any usage of this in this file change, i think it should also be considered while checking for openssl lib,

canewsin (changes requested):

src/lib/sslcrypto/openssl/library.py
purplesyringa: IIRC ZeroNet redefines find_library so it should work. It's worth testing though.

purplesyringa:
This PR is WIP.

src/lib/sslcrypto/openssl/library.py
purplesyringa: Should work now.

purplesyringa:
@HelloZeroNet Ready to be merged.

purplesyringa:
> tests failing
> segfault

FFFUUUUUUUU...

purplesyringa:
There's some evidence the problem is caused by sqlite, not sslcrypto (https://gist.github.com/imachug/1078f745984777f1b918e623c6726420#file-travis-log-L1351, backed up just in case). There's also a person who had the same issue: https://github.com/mrozekma/Sprint/issues/194#issuecomment-33131798.

purplesyringa:
@HelloZeroNet Could you please restart the test to make sure the segfault can't be reproduced?

HelloZeroNet:
Done, recently I have seen it multiple times with Python 3.4. I think it's safe to ignore it, maybe there is some incompatibility with latest libs. (maybe we could mark it as allow [fail environment](https://docs.travis-ci.com/user/customizing-the-build#jobs-that-are-allowed-to-fail))

canewsin:
@HelloZeroNet 
fails seems to be one on one, I see that if a commit test fails it was succeed in next push.

purplesyringa:
@HelloZeroNet Do you think you can merge this?

HelloZeroNet:
Seems fine to be, I will run the test for a few times, to see if there is any related problems happens.

canewsin:
@imachug i recently faced libcrypto.so and other crypto libs not found error on ZeroMail Zite, will it solve those.

purplesyringa:
It most likely will.

purplesyringa:
@HelloZeroNet Have you run the tests yet?

HelloZeroNet:
Yeah, no error so far, I will do some more runs, but I think we can merge it now.

purplesyringa:
Oh god... again

HelloZeroNet:
Latest run on travis got a different kind of segfault:
https://travis-ci.org/HelloZeroNet/ZeroNet/jobs/657714713
https://api.travis-ci.org/v3/job/657714713/log.txt

Probably not related (it does not blocks this PR): I have seen it before, I was unable to reproduce it, not sure when/why is it happening, but it happens on shutdown, after tests have run successfully. I think it's not critical, but would be great to find a way to reproduce it/fix it.  Maybe it's related to the custom `__del__` function of threadpools: https://github.com/HelloZeroNet/ZeroNet/blob/py3/src/util/ThreadPool.py#L58
Related: https://joekuan.wordpress.com/2015/06/30/python-3-__del__-method-and-imported-modules/

The `__del__` function is required to avoid memory leak when deleting threadpools without calling kill() on it, but we don't remove threadpools on runtime, so it should not be an issue for us.

purplesyringa:
For the record: I've noticed that Travis sometimes manually restarts builds so here's the backed up log: https://gist.github.com/imachug/dcc1934f7a3f8571d4cc7be83ed03059.

purplesyringa:
@HelloZeroNet That could be the reason. Do you think we could rewrite ThreadPool as a context manager? AFAIK it's only deleted at shutdown (which we're not interested in anyway) and in tests so it shouldn't be a problem.

purplesyringa:
@HelloZeroNet Do you think this PR can be merged then?

purplesyringa:
I've managed to reproduce a segfault on GitLab, but it turns out it can be reproduced *without* sslcrypto. WTF.

purplesyringa:
I can confirm that replacing `__del__` most likely fixes the segfault.

HelloZeroNet:
Thanks, I have pushed the change: https://github.com/HelloZeroNet/ZeroNet/commit/c4f65a5d7b48ad43646c9a57b3c1161042b3dd52

I will run it for a few times, let's see if the segfault happens again.

purplesyringa:
SQLite is failing again. Can you run the tests on py3 branch several times to make sure it's not related to sslcrypto?

HelloZeroNet:
Yeah, I will run it a few more times, but I have seen it before, so 90% sure it's not realted

HelloZeroNet:
I was unable to trigger segfault both on this and both on current branch. Do you think we can merge this or you want to do more tests?

purplesyringa:
> Do you think we can merge this

LGTM

purplesyringa:
Yay, thanks.

PR #2457, opened 2020-03-05
Make ThreadPool a context manager to prevent memory leaks

purplesyringa:
A more thought-out fix of the segfault reproduced in #2454.


purplesyringa:
Thanks

PR #2471, opened 2020-03-10
Search for any OpenSSL version in LD_LIBRARY_PATH

### LINKED ISSUE #2470: Openssl Lib Not Found Error.
```
if "LD_LIBRARY_PATH" in os.environ:
        lib_dir_paths = os.environ["LD_LIBRARY_PATH"].split(":")
        for path in lib_dir_paths:
            try:
                return [lib for lib in os.listdir(path) if "libcrypto.so.1.0" in lib [0]
```

Currently I was shipping `libcrypto.so.1.1` on ZeroNet Mobile, this results in openssl Lib not found error
@imachug @HelloZeroNet

purplesyringa:
Fixes #2470. Please check @canewsin.

### DIFF
--- a/src/util/OpensslFindPatch.py
+++ b/src/util/OpensslFindPatch.py
@@ -42,7 +42,7 @@ def getOpensslPath():
         lib_dir_paths = os.environ["LD_LIBRARY_PATH"].split(":")
         for path in lib_dir_paths:
             try:
-                return [lib for lib in os.listdir(path) if "libcrypto.so.1.0" in lib][0]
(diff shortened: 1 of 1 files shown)


purplesyringa:
As a side note: sslcrypto supports all versions of OpenSSL, so even if this change didn't work before (with pyelliptic and such), it should work now.

canewsin:
@imachug will find_library handle .so versioned libs ?

purplesyringa:
Yeah, it's `in` operator so it will.

canewsin (approved):

HelloZeroNet:
Thanks!

PR #2502, opened 2020-03-29
Add support for compressed addresses

### LINKED ISSUE #2501: Support for compressed Bitcoin vanity addresses
**Is your feature request related to a problem? Please describe.**

For example, I want to have a vanity address for my site. I use some vanity generator (like VanityGen or VanitySearch) to generate it and get address and private key.

But Bitcoin has two types of classic (`1`) addresses: compressed and uncompressed. And it seems that ZeroNet only supports uncompressed version, which makes it impossible to use it with compressed vanity addresses (it is possible if you convert it to uncompressed version but this doesn't make sense because you would lose vanity prefix).

For example, let's say that I generated some compressed vanity address, for example `1FQbDerUcLDTDYe1FqXnQrgzww4DA7C9io` (it's not vanity or used on any site but just example). Its compressed private key is `L3ocupNWioTwhLgu64xcSs5mGSZEPd6LevpUMmiZ2Q8E25Pah1x9`. Based on bitaddress.org, the uncompressed version of this private key is `5KJp75PYf21fg3eb1F3BJBtq6K2QstgGnJBJXXyYGiweqq25stV`, which results in `1K5XQVPpdxAyBnocSp6GhqnfJ77TESQB7d` as uncompressed address.

I can successfully use `1K5XQVPpdxAyBnocSp6GhqnfJ77TESQB7d` (uncompressed address) as address of my site. Signing it with uncompressed private key works normally.

However, given that my vanity address is compressed, I want to to use it for site address. So I create site with "vanity" compressed address `1FQbDerUcLDTDYe1FqXnQrgzww4DA7C9io` in the same way as previously. But when I try to sign it, it doesn't work.

When I try to sign it with compressed private key, it results in this signing error:

```
Error signing content.json
Private key invalid! Valid signers: ['1FQbDerUcLDTDYe1FqXnQrgzww4DA7C9io'], Private key address: 1CVdzwL9c2WVEkHMmkrHS9o6CD4WQaj2iR
```

When I try it with uncompressed private key, it also results in a similar error:

```
Error signing content.json
Private key invalid! Valid signers: ['1FQbDerUcLDTDYe1FqXnQrgzww4DA7C9io'], Private key address: 1K5XQVPpdxAyBnocSp6GhqnfJ77TESQB7d
```

In this case, you can see that it detects private key's address as uncompressed version, but fails to sign it because I want to use it on compressed address.

**Describe the solution you'd like**

ZeroNet should support both compressed and uncompressed Bitcoin addresses.

filips123:
This adds support for compressed Bitcoin site addresses and closes #2501.

It's using `sslcrypto` changes from imachug/sslcrypto#7 to detect if WIF private key has compression flag and then uses it to correctly get address and sign content.

Apparently `verify` does not need to be modified. Sites with compressed addresses can be viewed on older versions, but signing them does not work (I'm not sure for signing of the user's content).

### DIFF
--- a/src/Crypt/CryptBitcoin.py
+++ b/src/Crypt/CryptBitcoin.py
@@ -58,9 +58,11 @@ def privatekeyToAddress(privatekey):  # Return address from private key
     try:
         if len(privatekey) == 64:
             privatekey_bin = bytes.fromhex(privatekey)
+            is_compressed = False
         else:
             privatekey_bin = sslcurve.wif_to_private(privatekey.encode())
-        return sslcurve.private_to_address(privatekey_bin, is_compressed=False).decode()
+            is_compressed = sslcurve.is_compressed(privatekey.encode())
+        return sslcurve.private_to_address(privatekey_bin, is_compressed=is_compressed).decode()
     except Exception:  # Invalid privatekey
         return False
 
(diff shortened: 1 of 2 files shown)


HelloZeroNet:
Thanks. 
Looking good to me, but approve from @imachug would be great as it modifies the lib and we don't want to fork out.

Maybe a test case would be great to Test/TestCryptBitcoin.py

purplesyringa:
Just wondering if you've checked whether verifying works, I'm pretty sure it doesn't. I'm currently working on a better PR with automatic compressed key detection and it looks like signing actually *always* signs with an uncompressed key.

filips123:
Verifying doesn't fail, but I don't know if it actually works. I was able to sign and publish site with compressed address on instance with this addition, and view it on another instance with normal version.

PR #2509, opened 2020-04-07
Fix redirect loop in Chrome 80

SuperSandro2000:
This fixes the following error message from Chrome 80

A cookie associated with a cross-site resource at http://zeronet.example.com/ was set without the `SameSite` attribute. It has been blocked, as Chrome now only delivers cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.

### DIFF
--- a/plugins/disabled-UiPassword/UiPasswordPlugin.py
+++ b/plugins/disabled-UiPassword/UiPasswordPlugin.py
@@ -65,7 +65,7 @@ class UiRequestPlugin(object):
                 url = self.env.get("HTTP_REFERER", "")
                 if not url or re.sub(r"\?.*", "", url).endswith("/Login"):
                     url = "/" + config.homepage
-                cookie_header = ('Set-Cookie', "session_id=%s;path=/;max-age=2592000;" % session_id)  # Max age = 30 days
+                cookie_header = ('Set-Cookie', "session_id=%s;path=/;max-age=2592000;secure;samesite=none;" % session_id)  # Max age = 30 days
                 self.start_response('301 Redirect', [('Location', url), cookie_header])
                 yield "Redirecting..."
 


purplesyringa:
Will it work without https?

SuperSandro2000:
> Will it work without https?

no, it won't. I guess we need to add something to fix that.
There seems to be no existing config option tough....

purplesyringa:
Perhaps using config options is not the correct solution, because then either HTTP or HTTPS will break. What if we make the browser send the protocol to the server?

SuperSandro2000:
> What if we make the browser send the protocol to the server?

That would work I guess but I am not sure how we want to do this. I don't think there is a header for this except if you run it behind a reverse proxy which might set ``X-Forwarded-Proto``.

purplesyringa:
We could just use JavaScript on the client side.

SatoshiNakamotoSwitzerland:
We should not follow request from users to change the core, when they have browser related issues! ZeroNet is only secure and anonymous if it is used with the [Tor Browser](https://www.torproject.org/download/). I'm against all and any modification of the software just because "Google does this and that". Additionally is worth noting that @rllola working on ZeroNet browser and she needs help. https://github.com/rllola/ZeronetBrowser I don't want to see any change in the core because of Google (or any other corporation for that matter) and I know that many developers agree with me.

purplesyringa:
Lots of people use Chrome even if we don't like it. We should fix problems if want ZeroNet to be widely used.

SatoshiNakamotoSwitzerland:
> Lots of people use Chrome even if we don't like it. We should fix problems if want ZeroNet to be widely used.

I see any modification in the core or any change to work with a specific browser as a security threat. I think people whouls be motivated to use Tor Browser with ZeroNet securely, not the day-to-day use browser with plugins installed in that. 

I personally going to get rid all of the Windows, Chrome related code from the core of ZeroNet and only run it with the absolutely necessary code. Any browser or OS related commit is just making more complicated the development of ZeroNet. 

Today Google do this, you implement something, tomorrow Google does something else and again you provide fix and this repeats over and over until you find yourself developing software for Google and not for the people. 

In my opinion all OS and browser related issues should be ignored or at least should be separated from the core. Makes no sense in my opinion to work for Chrome and Windows, in Firefox ZeroNet works perfectly so need to motivate people either use Firefox or Tor Browser for privacy reasons and not pushing the limits. 

You can maybe create a different repository and not implementing here all kinds of trash from Google or anyone else.

SuperSandro2000:
> We could just use JavaScript on the client side.

well, I don't think I can do this in a proper clean way in ZeroNet right now. Help would be appreciated.

SuperSandro2000:
> We should not follow request from users to change the core, when they have browser related issues!

Then ZeroNet is dead in 5 years, most 10.

> ZeroNet is only secure and anonymous if it is used with the Tor Browser.

For that matter. According to Chromestatus Firefox is going to follow that soon so Tor is going to get that change eventually, too. Didn't find a confirmation in the source tough.

Also the Change is not some random change from the Chromium or Google people but a [standard](https://tools.ietf.org/html/draft-west-cookie-incrementalism-00) they follow and I do not think ZeroNet should stand against standards for no good reason.

> Additionally is worth noting that @rllola working on ZeroNet browser and she needs help.

It is neither based on Firefox or Chromium so I won't support that. An addon to Firefox or Tor or a fork of one of them would be perfectly fine and wouldn't need a reimplementation of such a complex thing as a browser.

> I don't want to see any change in the core because of Google (or any other corporation for that matter) and I know that many developers agree with me.

Then the password plugin will be broken as soon as Firefox follow the new standard. To bad but I want password protection infront of my zeronet.

This also has nothing to do with Windows and technically I already forked ZeroNet and I have no intent to fork it just to fix this nitty little bug.

Also this is probably in Chromium already (didn't test) so we just excluded a vast majority of all people using zeronet with password authentication.

SatoshiNakamotoSwitzerland:
**password plugin** :roll_eyes: Who the hell need that plugin anyway?

**I going to get rid off all unnecessary plugins, UI, and trashy code related to OS and browsers from ZeroNet.** 

**ZeroNet going to be renamed too** and going to be way faster and stable than it is right now.

filips123:
> I going to get rid off all unnecessary plugins, UI, and trashy code related to OS and browsers from ZeroNet.
> ZeroNet going to be renamed too and going to be way faster and stable than it is right now.

Yes. When you remove all those things, rename it to "zero".

purplesyringa:
> problem with the "password plugin" so the solution to your problem should be implemented in that plugin and not in the core!

And now for something completely different: this patch changes the plugin, not the core.

SuperSandro2000:
> password plugin roll_eyes Who the hell need that plugin anyway?

then remove it and don't offer it at all.

Chromium is also effected. Just tested it.

If you know what the right solution would be to fix this in the password plugin without affecting the core I am more than welcome to accept it. But this was the fastest and easiest solution to fix my problem and I just wanted to make sure it gets fixed upstream as well.

and I think you are getting a bit off topic here. Sure there are more important things to do but this would have been an easy fix where we don't need to argue over and over again but just do it and get over it.

Edit:
Wait a second. Switzerland flag, joined yesterday, wasting everyones time. I kinda remember such a person from another issue I don't want to mention now.

Back to topic:

Secure and samesite=none can only be set on https so this would break the login on http sites as the cookie then wouldn't be set at all.

HelloZeroNet:
One solution could be a built-in let's encrypt client for https cert, but as far as I know let's encrypt does not support cert for IP addresses, so it would require a domain name.

The cookie based authentication has other issues (Eg. it's missing from custom font file requests), so some better alternative would be great.

ghost:
@SuperSandro2000 set up a reverse-proxy and than you can use letsencrypt or cookies (**add headers to the proxy or gateway as you like**). @HelloZeroNet on April 8, 2020 renewed the zeronet.io's certificate right after commenting here. :rofl: 

> The cookie based authentication has other issues (Eg. it's missing from custom font file requests), so some better alternative would be great.

Some sites doesn't use cookies at all! So we don't need cookies in ZeroNet just to satisfy Google... 
The solution is setting up zeronet proxies or gateways with the headers Google wishes in its shitty Chrome... You can also just give everything to the US government or contact directly the NSA without Google...

SuperSandro2000:
@warcriminal I have that already and the plugin would still break.

> Some sites doesn't use cookies at all!

Which does not work for login protection from non static IPs without having some sort of certificate or cookie or you enter the login details for every request.
You still don't get that this will eventually also come to Firefox cause the current way of handling cookie is just insecure.

SuperSandro2000:
I think 36d96d484ed8dc3370c2dab9b50b714af702a2c4 closed this issue. Closing for now.

PR #2510, opened 2020-04-08
Use Gevent prerelease for Python 3.8

filips123:
This changes Gevent for Python 3.8 from using Git archive to the latest prerelease version. This enables wheel installation on supported systems and results in faster dependency installation.

However, a special case for Python 3.8 is still needed, because the changes were not released as 1.5 yet.


HelloZeroNet:
Thanks!

filips123:
Ironically, two days after this PR (yesterday), stable Gevent 1.5.0 was released. This means that we probably don't need condition anymore and can just require `gevent>=1.1.0` for all Python versions.

Issue #2517, opened 2020-04-11
Debian cleared (removed) all modules! ZeroNet broken by Debian! Python 3.8


ghost:
After apt update && apt update && apt autoremove --purge on Debian removed all requirements! I have that feeling that is related to Python 3.8... 

@imachug @HelloZeroNet @filips123 

- Starting ZeroNet...
[09:26:06] PluginManager Plugin Bigfile load error: ModuleNotFoundError: No module named 'merkletools' in PluginManager.py line 138 > Bigfile/__init__.py line 1 > Bigfile/BigfilePlugin.py line 20
[09:26:06] PluginManager Plugin CryptMessage load error: ModuleNotFoundError: No module named 'base58' in PluginManager.py line 138 > CryptMessage/__init__.py line 1 > CryptMessage/CryptMessagePlugin.py line 7 > CryptBitcoin.py line 12 > __init__.py line 4 > 1 > aes.py line 4 > __init__.py line 2 > ecc.py line 4 > _ecc.py line 4
[09:26:06] - Version: 0.7.1 r4473, **Python 3.8.2** (default, Apr  1 2020, 15:52:55) 
[GCC 9.3.0], Gevent: 1.4.0
[09:26:06] - Unhandled exception: No module named 'base58'
Traceback (most recent call last):
  File "zeronet.py", line 17, in main
    main.start()
  File "/opt/zeronet/src/main.py", line 602, in start
    actions.call(config.action, action_kwargs)
  File "/opt/zeronet/src/main.py", line 133, in call
    back = func(**kwargs)
  File "/opt/zeronet/src/main.py", line 141, in main
    from Ui import UiServer
  File "/opt/zeronet/src/Ui/__init__.py", line 1, in <module>
    from .UiServer import UiServer
  File "/opt/zeronet/src/Ui/UiServer.py", line 10, in <module>
    from .UiRequest import UiRequest
  File "/opt/zeronet/src/Ui/UiRequest.py", line 14, in <module>
    from User import UserManager
  File "/opt/zeronet/src/User/__init__.py", line 1, in <module>
    from .User import User
  File "/opt/zeronet/src/User/User.py", line 9, in <module>
    from Crypt import CryptBitcoin
  File "/opt/zeronet/src/Crypt/CryptBitcoin.py", line 12, in <module>
    from lib import sslcrypto
  File "/opt/zeronet/src/lib/sslcrypto/__init__.py", line 4, in <module>
[...]

PR #2518, opened 2020-04-11
Update requirements.txt

ghost:
https://github.com/HelloZeroNet/ZeroNet/issues/2517

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-gevent>=1.2.2
+gevent>=1.4.0
 msgpack>=0.4.4
 base58
 merkletools
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 gevent>=1.2.2
-gevent==1.5.0 ; python_version >= "3.5"
+gevent==1.5.0; python_version >= "3.5"
 msgpack>=0.4.4
 base58
 merkletools
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
-gevent>=1.4.0; python_version < "3.8"
-gevent>=1.5.0; python_version >= "3.8"
+gevent>=1.5.0
 msgpack>=0.4.4
 base58
 merkletools
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
-gevent>=1.5.0
+gevent>=1.2.2
+gevent>=1.5.0 ; python_version >= "3.5"
 msgpack>=0.4.4
 base58
 merkletools
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 gevent>=1.2.2
-gevent==1.5.0; python_version >= "3.5"
 msgpack>=0.4.4
 base58
 merkletools
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-gevent>=1.1.0; python_version < "3.8"
-gevent>=1.5.0a2; python_version >= "3.8"
+gevent>=1.4.0; python_version < "3.8"
+gevent>=1.5.0; python_version >= "3.8"
 msgpack>=0.4.4
 base58
 merkletools
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 gevent>=1.2.2
-gevent>=1.5.0 ; python_version >= "3.5"
+gevent==1.5.0 ; python_version >= "3.5"
 msgpack>=0.4.4
 base58
 merkletools


ghost:
@filips123 was fast so he added a pre-release to requirements.txt. The stable version uploaded yesterday, see: https://pypi.org/project/gevent/1.5.0/#history

I also changed from 1.1.0 to 1.4.0 but be honest "gevent 1.5 runs on Python 2.7.9 and up, and Python 3.5, 3.6, 3.7 and 3.8" so this line going to be removed as well.

ghost:
@filips123 opened fast another pull request when he seen my one: https://github.com/HelloZeroNet/ZeroNet/pull/2519 My pull request should be merged for obvious reasons.

filips123:
I created this PR before I saw yours (although your was first). But yes, yours can be merged if you want.

ghost:
@filips123 yes I want this to be merged, can you close your pull request? Thank you very much!

filips123:
Gevent 1.5 does not support Python 3.4 but ZeroNet does.

ghost:
@filips123 
> Gevent 1.5 does not support Python 3.4 but ZeroNet does.

https://pypi.org/project/gevent/1.5.0/#supprted-platforms well, maybe its is the time to drop support for 3.4. Don't you think @imachug? 

requirements.txt should only include gevent 1.5.0 because according to gevent maintainer older versions are not supported. 

> Users of older versions of Python 2 or Python 3 may install an older version of gevent. **Note that these versions are generally not supported.**

If someone running on older versions of python than that person should install the correct version of gevent which is **in most cases is version 1.5.0**. Since **Gevent 1.5.0 supports all version from 2.7.9 to up**, except 3.4 (:confused: strange) we should just drop support for that version.

ultramassiveblackhole:
Ping-pong @HelloZeroNet

filips123:
> well, maybe its is the time to drop support for 3.4. 

Time to drop 3.4 support was already few years ago...

> Since Gevent 1.5.0 supports all version from 2.7.9 to up, except 3.4

It support Python 2.7+, except 3.0, 3.1, 3.2 and 3.3. Basically Python 2.7 and 3.5+.

ghost:
@filips123 
> It support Python 2.7+, except 3.0, 3.1, 3.2 and 3.3. Basically Python 2.7 and 3.5+.

It support Python 2.7+, except 3.0, 3.1, 3.2, 3.3 **and 3.4**. Basically Python 2.7 and 3.5+. :+1: 

So what now? You @filips123 run 3.8 and I guess @imachug 3.7 which is 3.5+, I think most people have at least python version 3.5. 

Again, older versions of gevent are not supported! Why would someone who has Python 3.7 want to install gevent 1.1.0? Makes no sense. 

I recommend to merge this and if anyone have Python 3.4 probably we will see it when opening an issue about it and than we can link this conversation. I don't think that anyone from here running Python 3.4, probably not even 2.7. My guess is that in 2020 at least Python version 3.6 is installed for most.

purplesyringa:
> https://pypi.org/project/gevent/1.5.0/#supported-platforms well, maybe its is the time to drop support for 3.4. Don't you think @imachug?

I don't think so. Python 3.4 is still the latest available version on some devices. We've dropped Python 2 support just around a year ago, it's not the right time to drop 3.4 yet.

ghost:
> > https://pypi.org/project/gevent/1.5.0/#supported-platforms well, maybe its is the time to drop support for 3.4. Don't you think @imachug?
> 
> I don't think so. Python 3.4 is still the latest available version on some devices. We've dropped Python 2 support just around a year ago, it's not the right time to drop 3.4 yet.

https://github.com/HelloZeroNet/ZeroNet/pull/2518#issuecomment-612415864 
Show me @imachug  just one device which uses Python 3.4. :thinking: What device you talk about some robot on Mars? 

Termux on Android for that matter has Python version 3.8.2. 
@styromaniac what version of Python you have?

purplesyringa:
Okay, let me say what I meant another way. Currently, ZeroNet doesn't use any Python 3.5-specific functions or something, so there's no reason to stop Python 3.4 support when the only Python 3.4-related detail is a single line in `requirements.txt`.

ghost:
So,

gevent=1.2.2 ; python_version=="3.3"
gevent=1.3.7 ; python_version=="3.4"
gevent=1.4.0 ; python_version=="3.4.3"
gevent=1.5.0

@filips123

@imachug this is not just one line...

purplesyringa:
> this is not just one line

It used to be a single line before your PR. Anyway, can't we just request `>=1.1.0` for Python 3.4 and below and `>=1.5.0` for Python 3.5 and above?

ghost:
> > this is not just one line
> 
> It used to be a single line before your PR. Anyway, can't we just request `>=1.1.0` for Python 3.4 and below and `>=1.5.0` for Python 3.5 and above?

@imachug 
No because https://pypi.org/project/gevent/1.5.0/#older-versions-of-python clearly shows that 1.1.0 is not good for Python version 3.4.3 for example. We can also separate requirements.txt because again, gevent 1.5.0 supports python version 2.7.9 and 3.5+ but not versions 3.0-3.4.3. 

gevent

in the other hand **1.1.0 is just about 4 years old** release of gevent: https://pypi.org/project/gevent/1.5.0/#history

purplesyringa:
How is it "not good"? It's `>=`, not `=`.

ghost:
> How is it "not good"? It's `>=`, not `=`.

@imachug 
It is just stupid to use a 4 year old version of gevent when there is newer and very specifically stating what version of gevent supports which version of python. You can't just use 1.1.0 for Python version 3.4.3 because it is not supported! 

It is not better to have the correct version of gevent for the correct version of python? Pip would install version 1.1.0 for python 2.7.9 when in reality gevent 1.5.0 supports that! 

It is better to say the correct version even if its takes more lines in requirements.txt

ghost:
@jamadden can you comment on this? :thinking:

purplesyringa:
> It is not better to have the correct version of gevent for the correct version of python? Pip would install version 1.1.0 for python 2.7.9 when in reality gevent 1.5.0 supports that!

Do you know what `>=` means? It installs *any* version greater than or equal to the given one. If a compatible version is already installed, it doesn't do anything; if it's not, it downloads the latest compatible version which is exactly what we want.

ghost:
> > It is not better to have the correct version of gevent for the correct version of python? Pip would install version 1.1.0 for python 2.7.9 when in reality gevent 1.5.0 supports that!
> 
> Do you know what `>=` means? It installs _any_ version greater than or equal to the given one. If a compatible version is already installed, it doesn't do anything; if it's not, it downloads the latest compatible version which is exactly what we want. 

> gevent 1.3 runs on Python 2.7 and Python 3. Releases 3.4, 3.5 and 3.6 of Python 3 are supported. (Users of older versions of Python 2 need to install gevent 1.0.x (2.5), 1.1.x (2.6) or 1.2.x (<=2.7.8); gevent 1.2 can be installed on Python 3.3.) 

Something like this would be better @imachug :

gevent==1.2.2; python_version<=2.7.8
gevent==1.4.0; python_version>=3.3;<3.5
gevent=1.5.0

or 

gevent>=1.2.2
gevent==1.5.0; python_version=2.7.9;>3.5 :thinking: 

@filips123 any comment? 


**gevent>=1.2.2
gevent==1.5.0 ; python_version == '2.7.9' or python_version >= '3.5'** :+1:

purplesyringa:
> `python_version == '2.7.9'`

We don't support Python 2 anymore so this part is useless.

ultramassiveblackhole:
> > `python_version == '2.7.9'`
> 
> We don't support Python 2 anymore so this part is useless.

gevent>=1.2.2
gevent==1.5.0 ; python_version >= "3.5" 

is good @imachug?

ghost:
@imachug in Travis all failed above Python 3.5 :confused:

purplesyringa:
I see

ghost:
What the hell? :confused:

ghost:
gevent>=1.4.0 :+1: Problem solved, 1.4.0 also pretty fresh, at least is not 4 years old and @imachug just reminded me that ZeroNet not supporting python 2.x anymore so it would be irresponsible to have 1.1.0 as minimum version for gevent. All checks have passed with  gevent>=1.4.0 :hugs: :partying_face:

HelloZeroNet:
gevent 1.4.0 not supported on Python 3.8 and as far as I know there is no incompatibility issues with earlier versions of gevent, so no reason to require 1.4.0.

I have changed to >=1.5.0 on python 3.8: https://github.com/HelloZeroNet/ZeroNet/commit/f3a839f422afd50f5bf8f17d83a8807571bc1dfe

PR #2522, opened 2020-04-12
Update requirements.txt (again)

ghost:
@HelloZeroNet gevent>=1.5.0; python_version >= "3.5" is not needed but I let it for now (change it in a few minutes), gevent>=1.4.0 (and ONLY this) is way enough to be in requirements.txt!!! Ask @filips123, if you are so f**ng ignorant that just searching for excuses to close my PR.  
@imachug can you talk with Tamas? :/

PR #2523, opened 2020-04-12
Update .travis.yml

ghost:
Travis here: https://github.com/HelloZeroNet/ZeroNet/pull/2522#issuecomment-612655676 used a cached version of gevent but we want this: https://github.com/HelloZeroNet/ZeroNet/pull/2522#issuecomment-612658601 (don't use pip cache) 
@imachug just pointed out here https://github.com/HelloZeroNet/ZeroNet/pull/2522#issuecomment-612665000 that;

> The problem is when gevent-1.4.0 is already cached (e.g. gevent was installed when 1.5.0 wasn't available yet). 

--upgrade --no-cache-dir :thinking: :+1:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ before_install:
   # - docker build -t zeronet .
   # - docker run -d -v $PWD:/root/data -p 15441:15441 -p 127.0.0.1:43110:43110 zeronet
 install:
- - pip install --upgrade -r requirements.txt
+ - pip install --upgrade --no-cache-dir -r requirements.txt
  - pip list
 before_script:
   - openssl version -a


purplesyringa:
Do you understand that this only "fixes" the problem on Travis but not on users' machines? Perhaps specifying the correct versions we actually need, not making pip guess what we want is the correct way.

ghost:
> Do you understand that this only "fixes" the problem on Travis but not on users' machines? Perhaps specifying the correct versions we actually need, not making pip guess what we want is the correct way.

@imachug we discussed this with @filips123 before, **installing gevent 1.4.0 support those versions of python which are not supported by gevent 1.5.0!** The users you talk about will download and install ZeroNet so users will have a fresh install by pip with the **correct version of gevent**!  

Totally unnecessary and be clear, it is also stupid to have anything more in requirements.txt than **gevent>=1.4.0** :+1:

purplesyringa:
Again. Please install `gevent>=1.4.0` for old Python versions and `gevent>=1.5.0` for new Python versions. The reasoning is: pretend someone uses a new Python version but they have installed gevent before 1.5.0 was published. In this case, installing requirements.txt will still use gevent 1.4.0. In case of PyPy this means breaking ZeroNet.

ghost:
> Again. Please install `gevent>=1.4.0` for old Python versions and `gevent>=1.5.0` for new Python versions. The reasoning is: pretend someone uses a new Python version but they have installed gevent before 1.5.0 was published. In this case, installing requirements.txt will still use gevent 1.4.0. In case of PyPy this means breaking ZeroNet.

@imachug 
Nobody installed gevent 1.4.0 before! You know why? Because, **requirements.txt had a version of 1.1.0 in ZeroNet for years!** If someone has lower version than 1.4.0, requirements.txt will upgrade it!

HelloZeroNet:
It should work as it is now.

PR #2524, opened 2020-04-12
Update README.md

ghost:
"Install from source" in a more beautiful way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,6 @@ Decentralized websites using Bitcoin crypto and the BitTorrent network - https:/
  - `wget https://github.com/HelloZeroNet/ZeroNet-linux/archive/dist-linux64/ZeroNet-py3-linux64.tar.gz`
  - `tar xvpfz ZeroNet-py3-linux64.tar.gz`
  - `cd ZeroNet-linux-dist-linux64/`
- - `sudo -- sh -c 'apt update && apt -y upgrade && apt install python3-pip && python3 -m pip install --upgrade --no-cache-dir -r requirements.txt'`
  - Start with: `./ZeroNet.sh`
  - Open the ZeroHello landing page in your browser by navigating to: http://127.0.0.1:43110/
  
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,12 +87,11 @@ There is an official image, built from source at: https://hub.docker.com/r/nofis
...
 
- - `wget https://github.com/HelloZeroNet/ZeroNet/archive/py3/ZeroNet-py3.tar.gz`
- - `tar xvpfz ZeroNet-py3.tar.gz`
- - `cd ZeroNet-py3`
- - `sudo apt-get update`
- - `sudo apt-get install python3-pip`
- - `sudo python3 -m pip install -r requirements.txt`
+ - `sudo su`
+ - `apt update && apt full-upgrade && apt install git python3-pip`
+ - `git clone https://github.com/hellozeronet/zeronet`
```
purplesyringa: `git` is not available on many machines by default and installing it just for ZeroNet is odd, why didn't you like the way ZeroNet is downloaded now?

README.md
```diff
@@ -87,12 +87,11 @@ There is an official image, built from source at: https://hub.docker.com/r/nofis
...
+ - `cd zeronet`
+ - `pip3 install --upgrade --no-cache-dir -r requirements.txt`
```
purplesyringa: Please at the very least replace `pip3` with `python3 -m pip` because `pip3` might refer to PyPy pip version or can not exist at all (e.g. on some Windows builds).

README.md
purplesyringa: The official URL is PascalCase, can you please change it?

README.md
```diff
@@ -87,12 +87,11 @@ There is an official image, built from source at: https://hub.docker.com/r/nofis
...
 ### Install from source
 
- - `wget https://github.com/HelloZeroNet/ZeroNet/archive/py3/ZeroNet-py3.tar.gz`
- - `tar xvpfz ZeroNet-py3.tar.gz`
- - `cd ZeroNet-py3`
- - `sudo apt-get update`
- - `sudo apt-get install python3-pip`
- - `sudo python3 -m pip install -r requirements.txt`
+ - `sudo su`
+ - `apt update && apt full-upgrade && apt install git python3-pip`
```
purplesyringa: `apt full-upgrade` also updates core modules which is most likely not what you expect when installing a userland app. Why is just `apt update` not enough?

README.md
```diff
@@ -87,12 +87,11 @@ There is an official image, built from source at: https://hub.docker.com/r/nofis
 
 ### Install from source
 
- - `wget https://github.com/HelloZeroNet/ZeroNet/archive/py3/ZeroNet-py3.tar.gz`
- - `tar xvpfz ZeroNet-py3.tar.gz`
- - `cd ZeroNet-py3`
- - `sudo apt-get update`
- - `sudo apt-get install python3-pip`
- - `sudo python3 -m pip install -r requirements.txt`
+ - `sudo su`
```
purplesyringa: Please also add `exit` command to the end if you switch to `sudo su`. Some users who don't know Linux well might continue working under root and accidentally break something.

README.md
purplesyringa: `--no-cache-dir` disables, uh, cache which is inefficient. This is also an ugly solution to the "old gevent" problem when a clean one exists.

HelloZeroNet:
Thanks for PR, but I think it's easier to read in current form.

PR #2530, opened 2020-04-13
Replace SHA-512 hashing function with the ~10 times faster BLAKE3.

ghost:
blake3

BLAKE3 is a cryptographic hash function that is:

- **Much faster** than MD5, SHA-1, SHA-2, SHA-3, and BLAKE2.
- **Secure**, unlike MD5 and SHA-1. And secure against length extension,
  unlike SHA-2.
- **Highly parallelizable** across any number of threads and SIMD lanes,
  because it's a Merkle tree on the inside.
- Capable of **verified streaming** and **incremental updates**, again
  because it's a Merkle tree.
- A **PRF**, **MAC**, **KDF**, and **XOF**, as well as a regular hash.
- **One algorithm with no variants**, which is fast on x86-64 and also
  on smaller architectures.

The [chart below](https://github.com/BLAKE3-team/BLAKE3-specs/blob/master/benchmarks/bar_chart.py) shows BLAKE3's performance on modern server hardware, an Intel Cascade
Lake-SP 8275CL processor:
​
speed

BLAKE3 paper: https://github.com/BLAKE3-team/BLAKE3-specs/blob/master/blake3.pdf

---

**This is my last attempt to contribute to this repository** and the third time I changed the requirements.txt to include only gevent>=1.4.0! If this pull request is not going to be merged than this is going to be fastest fork of "ZeroNet". For that matter I going to rename it and will be not called "ZeroNet" anymore! 

I would like to give a huge thanks to @imachug and @filips123 for they great and hard work on this software, keep up the good work guys! :hugs: 

I wish that not @HelloZeroNet (@shortcutme) owns this repository. :cry:  It would be in better hands with @imachug for example. @imachug is way more respectful and active than the current "owner" :rage:.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ before_install:
   # - docker build -t zeronet .
   # - docker run -d -v $PWD:/root/data -p 15441:15441 -p 127.0.0.1:43110:43110 zeronet
 install:
- - pip install --upgrade -r requirements.txt
+ - pip install --upgrade --no-cache-dir -r requirements.txt
  - pip list
 before_script:
   - openssl version -a
--- a/plugins/Bigfile/BigfilePlugin.py
+++ b/plugins/Bigfile/BigfilePlugin.py
@@ -101,8 +101,8 @@ class UiRequestPlugin(object):
                 read, upload_info["size"], upload_info["piece_size"], out_file
             )
 
-        if len(piecemap_info["sha512_pieces"]) == 1:  # Small file, don't split
-            hash = binascii.hexlify(piecemap_info["sha512_pieces"][0])
+        if len(piecemap_info["blake3_pieces"]) == 1:  # Small file, don't split
+            hash = binascii.hexlify(piecemap_info["blake3_pieces"][0])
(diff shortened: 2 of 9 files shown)


purplesyringa:
Performance is a problem for ZeroNet given that it's written in Python. Hashes are checked really often (see a recent "Eternal Word" issue), so making them faster is important. BLAKE3 looks like a good alternative, but changing the hash function without backward compatibility is wrong. We could add both `blake3` and `sha512` fields when signing and check `blake3` on new versions and `sha512` on old versions on verifying.

ghost:
@imachug I'm not in favor of backward compatibility, at least not creating or signing sites with sha512.  Downloading the content which hashed by sha512 is enough.
Creating, signing new content should be hashed using blake3.
```
                else:  # Backward compatibility
                    hash_type = "sha1"
```
until now this :point_up: was in ContentManager.py and **turned out that is useless**, so I removed. Adding back sha512 instead of sha1 will still fail. There was no backward compatibility at all using this code.

At this point I see that this is going to be the fastest fork of ZeroNet. The @HelloZeroNet repository comes not even close to the speed of this fork with blake3.

Issue #2544, opened 2020-05-01
all.js didn't load in Firefox Nightly when UiPassword enabled


earthengine:
### My Environment
  * ZeroNet version: 0.7.1 (rev4206)
  * Operating system: Windows
  * Web browser: Firefox Nightly 77.0a1 (2020-04-30)
  * Tor status: not available
  * Opened port: yes
  * Special configuration: UiPassword Turned on

### The problem:
When the UiPassword turned on, the login form will set a session cookie. Firefox Nightly downloads the all.js file, but later when it opens the websocket it violates the Content Security Policy set by Mozilla.

Chrome  81.0.4044.129 instead, shows a warning message in the delovper console:

>A cookie associated with a cross-site resource at http://127.0.0.1/ was set without the `SameSite` 
> attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are 
> set with `SameSite=None` and `Secure`. You can review cookies in developer tools under 
> Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.

#### Steps to reproduce:

  1. Download and install the given version of Firefox
  2. Download and install ZeroNet
  3. Enable plugin UiPassword
  4. Open http://127.0.0.1:43110 in Firefox and enter the password

#### Observed Results:
A blank page with only the ZeroNet icon in the right top corner.

#### Expected Results:
A fully functional page like in Chrome  81.0.4044.129.

earthengine:
Firefox Nightly didn't allow sessions tracking in a HTTP site, even it is in localhost. To workaround, people need to use [local-ssl-proxy](https://www.npmjs.com/package/local-ssl-proxy) or similar.

ZeroNet should consider provide things like that out of box.

HelloZeroNet:
Thanks for the tip, creating a self-signed ssl could be a solution or by using letsencrypt + dyndns-like service.

Unfortunately we have seen various issue with cookie based authentication + sandboxed iframes: eg. css font file requests are called without any cookies, so some alternative would be nice, eg. ssl cert based authentication.

filips123:
Why Multiuser plugin, which also uses a cookie, works but UiPassword doesn't?

HelloZeroNet:
Because in the Multiuser plugin we only use the cookie when we load the wrapper html and for the websocket connection, which booth outside of the iframe.

[...]

PR #2545, opened 2020-05-01
Add SameSite flag to work with Firefox Nightly

earthengine:
A quick fix for #2544.

Tested in my own computer with doing the same to the `UiPasswordPlugin.py` file.


earthengine:
This is not working yet. It fixed in Firefox Nightly but breaks Chrome.

The test result is: if we set "SameSite=None" only, it works in Chrome. But in Firefox Nightly it stops the user from logging in (keep showing the login page). On the other hand, if we set "SameSite=None;secure", Firefox Nightly likes it but Chrome (and Edge) keep showing the login page.

I don't know what the correct solution should be; but I think ZeroNet will have to implement HTTPS and so "SameSite=None;secure" will work on both if using HTTPS.

PR #2556, opened 2020-05-10
Remove unnecessary debugger

gqgs:
It doesn't exist in the original `src/Ui/media/Wrapper.coffee` file.


HelloZeroNet:
Thanks, I forgot to re-compile it after the removal

PR #2614, opened 2020-09-18
Fixing languages in ZeroNet

JabbaDesilijicTiure:
- Rename `jp.json` files to `ja.json` because there is **no such language code** as **jp**!
- Rename all `zh.json` files to `zh-hans.json` (Chinese **S**implified)
- Rename all `zh-tw.json` files to `zh-hant.json` (Chinese **T**raditional) 

Regarding `zh-tw` I must say that Taiwan is not a separate country from China and traditional Chinese also written in Hong-Kong, Macau and other Chinese territories. Using the `zh-tw` language code is wrong (I even consider it as Sinophobic) and goes against the recommendation of the World Wide Web Consortium which strongly recommends to use `zh-hant` instead.


JabbaDesilijicTiure:
@imachug actually Mozilla also referring to `zh-hans` and `zh-hant`: 
https://github.com/mozilla/language-mapping-list/blob/6a1ba99f9625486705f0872750df228f52498bea/language-mapping-list.js#L837

imachug:
> RFC should be followed not the practices of corporations or other projects

Agreed, but if everyone uses a non-RFC standard, it's just dumb to follow RFC instead of 'actual' standard. Luckily, this is not a problem in this case.

imachug:
>  It is enough to use Language subtag plus Script subtag: either zh-hant or zh-hans.

Agreed

PR #2616, opened 2020-09-23
Segmentation fault with latest greenlet

JabbaDesilijicTiure:
<frozen importlib._bootstrap>:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
<frozen importlib._bootstrap>:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
<frozen importlib._bootstrap>:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
<frozen importlib._bootstrap>:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
<frozen importlib._bootstrap>:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
<frozen importlib._bootstrap>:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
<frozen importlib._bootstrap>:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
<frozen importlib._bootstrap>:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
<frozen importlib._bootstrap>:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
<frozen importlib._bootstrap>:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
Segmentation fault

PR #2617, opened 2020-09-23
Update requirements.txt

JabbaDesilijicTiure:
See: https://github.com/HelloZeroNet/ZeroNet/pull/2616#issuecomment-697341526

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-gevent>=1.1.0; python_version < "3.8"
-gevent>=1.5.0; python_version >= "3.8"
+gevent>=1.1.0; python_version < "3.7"
+gevent>=20.9.0; python_version >= "3.7"
 msgpack>=0.4.4
 base58
 merkletools
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
-gevent>=1.1.0; python_version < "3.7"
+gevent==1.1.0; python_version <= "3.4"
+gevent<=1.4.0; python_version <= "3.7"
 gevent>=20.9.0; python_version >= "3.7"
 msgpack>=0.4.4
 base58
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 gevent==1.4.0; python_version <= "3.6"
-greenlet<=0.4.16; python_version <= "3.6"
+greenlet==0.4.16; python_version <= "3.6"
 gevent>=20.9.0; python_version >= "3.7"
 msgpack>=0.4.4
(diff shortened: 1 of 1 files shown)


JabbaDesilijicTiure:
This good for you @imachug ?

requirements.txt
```diff
@@ -1,5 +1,6 @@
-gevent>=1.1.0; python_version < "3.8"
-gevent>=1.5.0; python_version >= "3.8"
+gevent==1.1.0; python_version < "3.0"
```
purplesyringa: This line looks odd, ZeroNet doesn't support Python 2 anymore so it looks like this can be safely removed.

requirements.txt
```diff
@@ -1,5 +1,6 @@
...
+gevent<=1.4.0; python_version < "3.7"
+gevent>=20.9.0; python_version >= "3.7"
```
purplesyringa, line 3: Why `1.4.0` instead of `1.5.0` or later? If that's for Python 3.4 support, I'd be more precise and use `==1.4.0` for 3.4 and `>=20.9.0` for 3.5 and above, but I haven't tested `20.9.0` on 3.5 so if it doesn't work there I'm ok with your solution (just replace `<=1.4.0` with `==1.4.0`).

JabbaDesilijicTiure: > On Python 3.7 and above, using greenlet 0.4.17 **requires** gevent 20.9.0.
https://github.com/HelloZeroNet/ZeroNet/pull/2616#issuecomment-697341526

JabbaDesilijicTiure: @imachug check out this: https://github.com/HelloZeroNet/ZeroNet/pull/2617/commits/16583cbcb104f0955b2e941a870f894711d4dbdc

JabbaDesilijicTiure: I don't think that anyone use Python versions from 3.1 to 3.3, so it is not necessary for us to support it by stuffing requirements.txt. In this way we support from 3.4 up to 3.7 with gevent 1.4.0 and above 3.7 installing at least 20.9.0. 

Below 3.7 I did not tested and it can be a real problem if below 3.7 they grabbing the latest greenlet because that is requires gevent 20.9.0. Maybe a `greenlet<=0.4.16; python_version <= "3.6"` needed. 

@imachug

JabbaDesilijicTiure: Can I mark it resolved? Are you okay with this @imachug ?

purplesyringa: I'm okay with this if it passes tests.

imachug (approved):

PR #2631, opened 2020-10-05
Allow sites to request several CORS permissions at once

### LINKED ISSUE #2630: [CorsPlugin] Multicors: is it possible?
Currently, there is a [CorsPlugin](https://github.com/HelloZeroNet/ZeroNet/blob/py3/plugins/Cors/CorsPlugin.py), which allows you to use a `cmd` for granting cross-site access. As far as I can see, it works this way:

1) You request for access by doing `page.cmd("corsPermission", ziteAddress)`
2) `wrapperConfirm` is fired to confirm Cors request, where user can "Grant&add" (if zite was not downloaded locally yet), "Grant" (if zite exists in local data) or user can close the wrapper toast.
3) Zite is copied to your local data (or permission is granted, if already there)
4) You then have an option to reach `"inner_path"` (directory/file) with `cors-` prefix to work with other zite's contents.

The inconvenient part comes up when you have a bunch of zites to grant cors to. 
I.e. if I have 20 or more zites to be cors'ed, I will need to do something like:

```js
for (let key in zitesList) {
  page.cmd("corsPermission", zitesList[key]);
}
```

But it will fire up only 2 active toast messages at a time (two last ones) - other messages will disappear.
I've came up with a workaround, relying on promises:

```js
// Sends a cmdp, waits for resolving and then self-calls this func again without queued element
  getCorsPermissionsOneByOne(zitesList = ['zite1', 'zite2', ... , 'zite20']) {
    console.info("Waiting for " + zitesList[0]);
    page.cmdp("corsPermission", zitesList[0]).then(
      (resolve) => {
        inactiveZites.splice(0, 1);
        if (inactiveZites.length !== 0) this.getCorsPermissionsOneByOne(zitesList);
      }
    );
  }
```

While this works as I need to, the user still will need to click "Grant&add" button 20 or more times. 

**The solution I would like to see**
Would it be possible to add some sort of multiple cors request to a bunch of zites in one call? I.e., you can pass an array of zites in arg `page.cmdp("corsPermission", arg)`, and user will be alerted with just one toast message with something like: "This site requests read permission to **20 sites**" (where **20 sites** could be a link -> when clicked, slides down the div with all the addresses to see).

_The main aim for me is to make it easier for user to grant cors for a bunch of resources. From the point of UI/UX, it is better for user to click just once to grant required permissions instead of clicking them one by one for a (possible) long period of time._

Misc questions: what possible pitfalls can there be? How will it download/hang some of unreachable zites and how to track their downloading/copying process?

purplesyringa:
One of the old patches I forgot to submit. Closes #2630.

Allows using `corsPermission([addr1, addr2, ...])` instead of issuing several `corsPermission` commands.

### DIFF
--- a/plugins/Cors/CorsPlugin.py
+++ b/plugins/Cors/CorsPlugin.py
@@ -94,7 +94,7 @@ class UiWebsocketPlugin(object):
 
             if not (site and "Cors:" + address in self.permissions):
                 # No site or no permission
-                site_names.append(cgi.escape(site_name))
+                site_names.append(site_name)
                 site_addresses.append(address)
 
         if len(site_names) == 0:
--- a/plugins/Cors/CorsPlugin.py
+++ b/plugins/Cors/CorsPlugin.py
@@ -121,10 +121,7 @@ class UiWebsocketPlugin(object):
         for address in addresses:
             site = self.server.sites.get(address)
             if not site:
-                # Spawn gevent thread
-                thread = gevent.spawn(self.server.site_manager.need, address)
-                if thread:
-                    thread.join()
+                gevent.spawn(self.server.site_manager.need, address)
 
 
 @PluginManager.registerTo("UiRequest")
--- a/plugins/Cors/CorsPlugin.py
+++ b/plugins/Cors/CorsPlugin.py
@@ -2,6 +2,7 @@ import re
 import html
 import copy
 import os
+import gevent
 
 from Plugin import PluginManager
 from Translate import Translate
@@ -74,28 +75,56 @@ class UiWebsocketPlugin(object):
         return self.corsFuncWrapper("actionOptionalFileInfo", to, inner_path, *args, **kwargs)
 
     def actionCorsPermission(self, to, address):
-        site = self.server.sites.get(address)
-        if site:
-            site_name = site.content_manager.contents.get("content.json", {}).get("title", address)
-            button_title = _["Grant"]
+        if isinstance(address, list):
+            addresses = address
         else:
-            site_name = address
-            button_title = _["Grant & Add"]
+            addresses = [address]
 
-        if site and "Cors:" + address in self.permissions:
+        button_title = _["Grant"]
+        site_names = []
+        site_addresses = []
+        for address in addresses:
+            site = self.server.sites.get(address)
+            if site:
+                site_name = site.content_manager.contents.get("content.json", {}).get("title", address)
+            else:
+                site_name = address
+                # If at least one site is not downloaded yet, show "Grant & Add" instead
+                button_title = _["Grant & Add"]
+
+            if not (site and "Cors:" + address in self.permissions):
+                # No site or no permission
+                site_names.append(cgi.escape(site_name))
+                site_addresses.append(address)
+
+        if len(site_names) == 0:
             return "ignored"
 
         self.cmd(
             "confirm",
-            [_["This site requests <b>read</b> permission to: <b>%s</b>"] % html.escape(site_name), button_title],
-            lambda res: self.cbCorsPermission(to, address)
+            [_["This site requests <b>read</b> permission to: <b>%s</b>"] % ", ".join(map(html.escape, site_names)), button_title],
+            lambda res: self.cbCorsPermission(to, site_addresses)
         )
 
-    def cbCorsPermission(self, to, address):
-        self.actionPermissionAdd(to, "Cors:" + address)
-        site = self.server.sites.get(address)
-        if not site:
-            self.server.site_manager.need(address)
+    def cbCorsPermission(self, to, addresses):
+        # Add permissions
+        for address in addresses:
+            permission = "Cors:" + address
+            if permission not in self.site.settings["permissions"]:
+                self.site.settings["permissions"].append(permission)
+
+        self.site.saveSettings()
+        self.site.updateWebsocket(permission_added=permission)
+
+        self.response(to, "ok")
+
+        for address in addresses:
+            site = self.server.sites.get(address)
+            if not site:
+                # Spawn gevent thread
+                thread = gevent.spawn(self.server.site_manager.need, address)
+                if thread:
+                    thread.join()
 
 
 @PluginManager.registerTo("UiRequest")


plugins/Cors/CorsPlugin.py
```diff
@@ -74,28 +75,56 @@ def actionOptionalFileInfo(self, to, inner_path, *args, **kwargs):
...
+        self.site.saveSettings()
+        self.site.updateWebsocket(permission_added=permission)
+
+        self.response(to, "ok")
+
+        for address in addresses:
+            site = self.server.sites.get(address)
+            if not site:
+                # Spawn gevent thread
+                thread = gevent.spawn(self.server.site_manager.need, address)
```
purplesyringa: Apparently running `need` in the same thread caused some blocking, though I can remember neither the reproduction steps nor why `join` is used here.

purplesyringa:
Is GitHub not going to trigger the pipeline?

You can check the tests in my repo, though. https://github.com/imachug/ZeroNet/actions/runs/290028296

plugins/Cors/CorsPlugin.py
```diff
@@ -74,28 +75,56 @@ def actionOptionalFileInfo(self, to, inner_path, *args, **kwargs):
...
+            if site:
+                site_name = site.content_manager.contents.get("content.json", {}).get("title", address)
+            else:
+                site_name = address
+                # If at least one site is not downloaded yet, show "Grant & Add" instead
+                button_title = _["Grant & Add"]
+
+            if not (site and "Cors:" + address in self.permissions):
+                # No site or no permission
+                site_names.append(cgi.escape(site_name))
```
scsmash3r: v. 0.7.1 r4206, Python 3.7.3

This line generates an error for me:
`Internal error: NameError: name 'cgi' is not defined`

scsmash3r (commented):
Can confirm it is working just as I expected :)
image

plugins/Cors/CorsPlugin.py
scsmash3r: Okay, missed the 'fix_typo' commit. Nevermind this :)

plugins/Cors/CorsPlugin.py
HelloZeroNet: We don't need to wait for the result if we don't want to process it.

If only one site is added it does not make any difference, but in case of multiple sites it will be slower with join, because it will wait for the result (complete site download) before starts downloading next site.

purplesyringa: Okay, I believe I would have left a better comment if I knew `thread.join` is important, so I'm going to remove it.

JabbaDesilijicTiure:
This is great but I would like to see an example why one site need stuff from a bunch of others. 
It is maybe useful when you want a voluntary search engine. 

What if I have a font on one site and use the CORS plugin to grab that font and use it on all other sites? 
Than I need to allow them one by one? This makes no sense. 

The plugin should have some list where you can add permissions on the bases of both incoming and outgoing.

scsmash3r:
> This is great but I would like to see an example why one site need stuff from a bunch of others.
> It is maybe useful when you want a voluntary search engine.
> 
> What if I have a font on one site and use the CORS plugin to grab that font and use it on all other sites?
> Than I need to allow them one by one? This makes no sense.
> 
> The plugin should have some list where you can add permissions on the bases of both incoming and outgoing.

You got it just right: for the search purposes. For now the early alpha looks like this:
image

For each entity there is a CORS requests needed to be done to a bunch of zites, from which the info will be pulled.
Regarding the font grabbing question: I think, it is not what CORS should be used for. You can shift the responsibility of getting the content on a client's side, but that will create an additional unnecessary step from user's side: he will need to allow the downloading of a content.

Instead, you should use [as](https://zeronet.io/docs/site_development/zeroframe_api_reference/#as) cmd to do that - and this action implies that you're the owner of the zite with a font, so you are able to actually make a request and get the content.

>The plugin should have some list where you can add permissions on the bases of both incoming and outgoing.

This is a good one point. When client allowed some CORS request to be made to another zite, the CORS plugin basically makes a copy of that zite with a reachable virtual directory with `cors-` prefix. I admit it would be a nice additional feature for the plugin itself to show what zite requested what other zite's contents - and to be able to 'untie' (undo the cors action) for the specific zite, if needed.

JabbaDesilijicTiure:
If I'm you, I would rename it to **SEARCH** instead of ~~ZSEARCH~~. There are forks out there where sites are called **sites** and not ~~zites~~ and those forks don't even use the name of ~~ZeroNet~~, ~~ZeroFrame~~... It was easy to guess that you are working on a search engine. 

Regarding the font example, CORS is built exactly for the reason I described above. 1FoNTaJHcUk1JkrcUJXxTVYTchxuhSHfp should allow hundreds of sites to load its resources.

JabbaDesilijicTiure:
It was easy to guess that you are working on a search engine. 

Regarding the font example, CORS is built exactly for the reason I described above. 1FoNTaJHcUk1JkrcUJXxTVYTchxuhSHfp should allow hundreds of sites to load its resources.

scsmash3r:
> It was easy to guess that you are working on a search engine.
> 
> Regarding the font example, CORS is built exactly for the reason I described above. 1FoNTaJHcUk1JkrcUJXxTVYTchxuhSHfp should allow hundreds of sites to load its resources.



> 1FoNTaJHcUk1JkrcUJXxTVYTchxuhSHfp

Unfortunately, there are no peers found on that zite. But I'll take that into consideration.
For my point of view, it is better for user interaction to NOT to ask a permission on whether to load or not to load particular font resource. If your zite is using some custom font, just serve it with zite's _own_ CSS and other _own_ resources. User will download it, and the font/resource will be 'just there, instantly used', without asking for permissions.

The possible backfire would be the content duplication... But still, if it comes to content serving, for me it is better to serve the files from your own folder - and you will be sure they are served. Rather than to hope that some another zite has peers and reachable at the moment (not saying about critical resources, that just need to be there)...

All in all, the stuff you're described, comes down to simple thing: the **control** on the client's side for what should be downloaded (CORS request with permission granted manually, by user).

**Personal thoughts on CORS stuff**
I'd like to see an auto-cors _(and this can be added as an option in plugin itself, if possible)_, but with manual possibility of switching that off. What I mean by saying 'auto-cors' is something like this:

1) For now, CorsPlugin is triggering a request, that needs to be solved manually by user. That means each visitor should allow each zite, so the currently visited zite could access the contents of another zite by copying its contents into visitor's data.
2) Visitor grants permission -> zite is copied -> zite and all of its files are accessible for the visitor by virtual path.

How would auto-cors work:

1) Visitor enters the zite -> while checking for peers and content.json (preloading page with red-squared loader), there is also a process of corsed zites downloading shown -> when finished, site loads fully, where all the corsed zites' files are already available for getting/quering by virtual path.
2) Visitor then can remove any cors dependencies, but here I don't really know how it will work out in a real use cases. This can lead to a situation when the visited zite won't work at all, cause it tightly relies on other's zite files - and this should not happen, imo.

I suppose there is pretty low chances to make it happen, because it implies some automation - and that implies less control from client's side. But it would be much easier for the content to dissiminate and for the UX/UI to have lesser manual steps to do.

HelloZeroNet:
Thanks for the PR, I have bumped the revision number to 4538 make it easier detect availability

PR #2635, opened 2020-10-11
Fixing templates

JabbaDesilijicTiure:
I have added two **missing body element** and **moved a style element to the head** section from body.

### DIFF
--- a/plugins/UiConfig/media/config.html
+++ b/plugins/UiConfig/media/config.html
@@ -7,7 +7,7 @@
  <meta http-equiv="content-type" content="text/html; charset=utf-8" />
  <link rel="stylesheet" href="css/all.css?rev={rev}" />
 </head>
-
+<body>
 
 <h1>ZeroNet config</h1>
 
--- a/plugins/UiPluginManager/media/plugin_manager.html
+++ b/plugins/UiPluginManager/media/plugin_manager.html
@@ -9,7 +9,6 @@
 </head>
 <body>
 
-
 <h1>ZeroNet plugin manager</h1>
 
 <div class="content" id="content"></div>
--- a/plugins/UiPluginManager/media/plugin_manager.html
+++ b/plugins/UiPluginManager/media/plugin_manager.html
@@ -7,6 +7,7 @@
  <meta http-equiv="content-type" content="text/html; charset=utf-8" />
  <link rel="stylesheet" href="css/all.css?rev={rev}" />
 </head>
+<body>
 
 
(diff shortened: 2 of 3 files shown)


src/Ui/template/wrapper.html
```diff
@@ -1,103 +1,119 @@
...
+    console.log("Changing url from " + document.getElementById("inner-iframe").src + " to " + iframe_src)
+    document.getElementById("inner-iframe").src = document.getElementById("inner-iframe").src // Workaround for Firefox back button bug
+    document.getElementById("inner-iframe").src = iframe_src
+    address = "{address}"
+    wrapper_nonce = "{wrapper_nonce}"
+    wrapper_key = "{wrapper_key}"
+    ajax_key = "{ajax_key}"
+    postmessage_nonce_security = {
+      postmessage_nonce_security
+    }
```
purplesyringa: Excess newlines

src/Ui/template/wrapper.html
```diff
@@ -1,103 +1,119 @@
...
+    file_inner_path = "{file_inner_path}"
+    permissions = {
+      permissions
+    }
```
purplesyringa: Excess newlines

src/Ui/template/wrapper.html
```diff
@@ -1,103 +1,119 @@
...
+    show_loadingscreen = {
+      show_loadingscreen
+    }
```
purplesyringa: Excess newlines

src/Ui/template/wrapper.html
JabbaDesilijicTiure: I see. When I remove the newlines, the site I created is not displaying at all.

src/Ui/template/wrapper.html
JabbaDesilijicTiure: @imachug file_inner_patch breaks my new empty site. I need to remove it to see the site.

imachug:
LGTM now, thanks for the PR!

JabbaDesilijicTiure:
There are some small changes @imachug. Added viewport, charset and <!DOCTYPE html>
 where was missing. The type as you see here https://github.com/HelloZeroNet/ZeroNet/pull/2635/commits/7532828c406caffb363b8fdd13d4b4ba38fb0083 is removed from all HTML files. 

I can fix all other HTML files in all plugins and sites if needed, like the homepage @HelloZeroNet. Possibly I will fix other plugins as well before I move to the homepage.

PR #2641, opened 2020-10-13
Delete all.css

JabbaDesilijicTiure:
.


JabbaDesilijicTiure:
> Does the `src/Ui/media/all.js` file exist and contain anything? Are/were you using `--debug` option at some moment of time?

I did used debug and all my css files are trashed! 

This can't be deleted at all: 
Screenshot from 2020-10-13 00-26-18

I replacing now all my files in Ui/media...

PR #2642, opened 2020-10-13
Use more unique yet short paths for logging

purplesyringa:
Now ZeroNet logs tend to be literally useless for debugging because most of the paths are incomprehensible. I propose replacing the path shortening algorithm with a more subtle one. Here are the major changes:

1. Paths such as `src/Ui/__init__.py`, should they appear in a traceback, are now encoded as `Ui/__init__.py` (was: `__init__.py`).
2. Paths such as `plugins/Chart/ChartDb.py` are now encoded as `ChartDb.py` (was: `Chart/ChartDb.py`). In general, paths such as `A/AB/ABC/X.py` are now encoded as `ABC/X.py` to remove unnecessary prefixes.
3. Paths such as `<frozen importlib._bootstrap>`, which may appear in a traceback due to import errors, are now skipped (was: `<frozen importlib._bootstrap>`).
4. As the algorithm has changed from 'only keep specific prefixes' (ex.: plugin name) to 'only remove unnecessary prefixes' (ex.: `.../src/`), paths outside ZeroNet root, should they appear in a traceback, are now anonymized: `/root/ZeroNet/src/main.py` is now encoded as `?/src/main.py` (was: `main.py`).
5. Third-party module paths now include module name: `.../site-packages/gevent/__init__.py` is now encoded as `<gevent>/__init__.py` (was: `__init__.py`).
6. Builtin python modules are now visibly separated from ZeroNet files: `/usr/lib/python3.8/os.py` is now encoded as `<os>` (was: `os.py`).
7. Gevent paths such as `src/gevent/event.py` are correctly parsed as `<gevent>/event.py` (was: `event.py`).

I've also added some tests.

### DIFF
--- a/src/Debug/Debug.py
+++ b/src/Debug/Debug.py
@@ -28,7 +28,12 @@ def formatExceptionMessage(err):
     return "%s: %s" % (err_type, err_message)
 
 
-python_lib_dir = os.path.dirname(os.__file__)
+python_lib_dirs = [path for path in sys.path if path.endswith("/site-packages") or path.endswith("/dist-packages")]
+python_lib_dirs.append(os.path.dirname(os.__file__))  # TODO: check if returns the correct path for PyPy
+
+root_dir = os.path.realpath(__file__)
+for _ in range(3):
+    root_dir = os.path.dirname(root_dir)
 
 
 def formatTraceback(items, limit=None, fold_builtin=True):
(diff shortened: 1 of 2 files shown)


JabbaDesilijicTiure:
I like it. Great idea. Thanks @imachug! :partying_face:

HelloZeroNet:
Thanks, looks better this way, added some fix for Windows path separators: https://github.com/HelloZeroNet/ZeroNet/commit/ecfb6b6b3aa80d62ef9c3b1d617d25a8fed320d6

PR #2645, opened 2020-10-14
Proper python3 syntax in zeroname_updater.py

JabbaDesilijicTiure:
@imachug @cclauss Can you take a look at this?

### DIFF
--- a/plugins/Zeroname/updater/zeroname_updater.py
+++ b/plugins/Zeroname/updater/zeroname_updater.py
@@ -1,4 +1,3 @@
-from __future__ import print_function
 import time
 import json
 import os
@@ -27,7 +26,7 @@ def processNameOp(domain, value, test=False):
         return False
     if "zeronet" not in data and "map" not in data:
     # Namecoin standard use {"map": { "blog": {"zeronet": "1D..."} }}
-        print("No zeronet and no map in ", data.keys())
+        print("No zeronet and no map in ", list(data.keys()))
         return False
     if "map" in data:
     # If subdomains using the Namecoin standard is present, just re-write in the Zeronet way
@@ -69,7 +68,7 @@ def processNameOp(domain, value, test=False):
     # Note: Requires the file data/names.json to exist and contain "{}" to work
     names_raw = open(names_path, "rb").read()
     names = json.loads(names_raw)
-    for subdomain, address in data["zeronet"].items():
+    for subdomain, address in list(data["zeronet"].items()):
         subdomain = subdomain.lower()
         address = re.sub("[^A-Za-z0-9]", "", address)
         print(subdomain, domain, "->", address)


plugins/Zeroname/updater/zeroname_updater.py
```diff
@@ -69,7 +68,7 @@ def processNameOp(domain, value, test=False):
     # Note: Requires the file data/names.json to exist and contain "{}" to work
     names_raw = open(names_path, "rb").read()
     names = json.loads(names_raw)
-    for subdomain, address in data["zeronet"].items():
+    for subdomain, address in list(data["zeronet"].items()):
```
purplesyringa: `list` is not required around `items()` in a `for` loop

JabbaDesilijicTiure: I think is better to leave it this way.

plugins/Zeroname/updater/zeroname_updater.py
```diff
@@ -27,7 +26,7 @@ def processNameOp(domain, value, test=False):
         return False
     if "zeronet" not in data and "map" not in data:
     # Namecoin standard use {"map": { "blog": {"zeronet": "1D..."} }}
-        print("No zeronet and no map in ", data.keys())
+        print("No zeronet and no map in ", list(data.keys()))
```
cclauss:
```suggestion
        print("No zeronet and no map in ", list(data))
```

plugins/Zeroname/updater/zeroname_updater.py
cclauss: I agree with @imachug.  The `list()` increases memory use and slows processing time which is why Python 3 moves towards iteration over creating fixed lists in RAM.  `list()` would be required if the contents of `data["zeronet"]` was being changed inside the for loop.

plugins/Zeroname/updater/zeroname_updater.py
cclauss: `dict.keys()` is almost never needed.

cclauss (approved):

plugins/Zeroname/updater/zeroname_updater.py
```diff
@@ -1,4 +1,3 @@
-from __future__ import print_function
```
cclauss, line 1: Do we still need to support Python 2?
If so, then this line can not be deleted or Python 2 will print line 29 as a tuple.

purplesyringa: I believe ZeroNet doesn't support py2 anymore.

purplesyringa: I believe ZeroNet doesn't support py2 anymore.

Issue #2622, opened 2020-09-30
Extremely sluggish propagation of the optional files to peers


slrslr:
### Step 1: Please describe your environment

  * ZeroNet version: rev4528
  * Operating system: W10
  * Web browser: FF beta
  * Tor status: available
  * Opened port: yes

### Step 2: Describe the problem:

Sorry for yet another whinning, but i have to as it seems like an issue that have to be fixed. I have published post on Millchan (famous zeronet site) with a few files attached (some were bigfiles no bigger than 10MB each)
Because i know how bad zeronet can sometimes be in seeding/downloading i have opened that poSt on three other servers each located in different location and each server has open poRt and clicked to seed all files of that post. In recent hours i was repeating the seed all request on that post on three servers as i was seeing servers has problems to download files of that post, only some files, some parts was downloaded. After around 6 hours like only half of the content was downloaded by my servers and it is not like all servers downloaded same parts, one has full file, other 0% of same file, etc. And it is just under 20MB total size of the files and my seeder connectivity is not bad + open port. That 3 servers hosts millchan and indeed maybe hundred/s of images/files on it while maybe another hundred/s of optional files are queued for download or being "paused" due to no peer.

On the host who published mentioned post, there is alot of content downloaded/uploaded. ZN directory has 5 million files and quite significant data transfer during previous hours. Zeronet is slow past years and swapping the computer after maybe 1-2 days of runtime. The CPU and HDD iops does not seem to be the bottleneck. If it is connection limit somewhere then zeronet users should be instructed (via ZN interface) to modify the limits or apply patch. If it can not be modified then zeronet should be supplied with the special browser that has no such limits.

Mentioned terribly slow propagation of the content is tragedy. Something has to be done about this so Zeronet can grow and people stay using it and do not giveup being too annoyed of the slowness. Thank you for the Zeronet, think about ways to allow it to go to the next level regarding performance.

mx5kevin:
If the file are larger than 10MB and in content.json not added this files as optional parameter. Then not possible to another users download this files. If the optional parameter added then the client not download with those files with the site. In both cases, files what larger than 10MB are dead. **The seed button is not used by anyone (if integrated in the site).** **And the Help distribute added optional files option setted up default way to 10MB and no one can use it.** Files what  larger than 10MB or setted up as optional they will be inaccessible. There is currently no workable solution to this problem.

[...]

Issue #2602, opened 2020-08-06
5 things that cause a serious problem in the operation of ZeroNet for several years


mx5kevin:
### Step 1: Please describe your environment

  * ZeroNet version: 0.7.1 (rev 4496) and all previous versions
  * Operating system: Windows 10
  * Web browser: Firefox 79.0
  * Tor status: not always
  * Opened port: no
  

### Describe the problem:

For simple users a few things are too unnecessarily complicated which should be simplified because this things make the software unusable for many user about long time. Setting up websites and downloading files should be made much easier for simple users.

### As website owner: 
Setting up websites should be made easier by copying and signing files to a folder.

-1.UTF-8 encoding: All pages must be converted to UTF-8 otherwise, the characters will not be displayed correctly in the browser. This requires a separate program for several files, which is difficult to find on the Internet.

-2.In content.json file, files larger than 10 mb are required as set as optional files or they cannot be downloaded. This is very inconvenient for site owners many times the larger files are part of the page. Not everyone can use the content.json commands. It would be simpler if no separate parameter is specified in content.json larger files would be downloaded in the same way as smaller files.

-3.Special line in HTML files: `<base href="" target="_top" id="base">` and `<meta charset="UTF-8">`
Don't have to setup the `<base href="" target="_top" id="base">` line to navigate correctly the pages. Many people don’t know to need to set it up and without it, you cannot navigate properly between pages.

### As user: 
Large files are not available on the network and this makes the use of the software unenjoyable. For simple users we should make it easy to download files. If it’s not even available right away when he visit back can check it out. There should be no complicated options for this because these cannot be handled by users. Most users cannot handle a +SEED button and all they see is that they can’t download anything.

-4.Large file download in single user zites: Download all large files automatically with the website. For simple users don’t have to worry about why the content (like larger fideo or etc files) isn’t available and why not. Advanced users should have a button if they do not want to download larger files on the site then this can be turned off at the touch of a button.

[...]

Issue #2566, opened 2020-05-22
Optional large file sharing/downloading major difficulties and problems


mx5kevin:
### Step 1: Please describe your environment

  * ZeroNet version: 0.7.1 (rev4486) and all previous versions
  * Operating system: Windows 10
  * Web browser: Firefox 76.0.1
  * Tor status: always
  * Opened port: no

### Step 2: Describe the problem:

1. Auto download big file size limit 10MB preset value: Many files are larger than 10MB.A less knowledgeable user don't know what this feature is for. If the page contains 20MB 100MB 1GB file the user can not download it. The other problem is that the user does not know what size have the largest file on the page.From files under 10MB non optional files such as HTML can even total 1GB/site.And for files larger than 10MB optional files can even altogether 40MB in the same site.

2. If in the content.json file not set the optional files then files what larger than 10MB they cannot be downloaded: A beginner website owner will not be able to set it.Users cannot download the files larger than 10MB.There are no easy-to-find detailed descriptions on how to set this up.

3. When the page loads,large files which are larger than 10MB  cannot be downloaded automatically.These files doe not seeded correcly in single sites this way.And in many cases, users cannot download it.Very confusing because they are part of the website in the same way. It can greatly degrade the enjoyment value of pages.

4. Can not resume files download without seed button: If nobody seed the file it cannot be set up easily to try to download the files the program later.

5. Priority: It takes a lot of time for the pages to appear: HTML,CSS,Pictures,js should download first to load the pages faster.When the download of these files is complete.The other files will be loaded even later.Which are not displayed by the browser.

#### Steps to reproduce:

Create a simple simple user website below and above 10MB mixed files.If the site contains 1GB under 10MB non optional files.And total 500GB mixed files 15-100MB files the problems are spectacular.But with a simple index.html and a 20MB file setup also given the problems.

Issue #2243, opened 2019-10-24
Download large files default way in single user zites, and make possible to upload/download extreme large files like 10GB or larger files too


mx5kevin:
**Is your feature request related to a problem? Please describe.**

In large (optional) files currently the are no way to automatic way download them when the user load a zite.And users can not know a optional file in a single user zite are how much big.Currently must enable in the right 0Net icon with a limit and a simple user he doesn't even know exactly what it is. All these is that there are files on a site which are not available.The site owners see the big files does not have seed this way but another files are correctly seeded.

**Describe the solution you'd like**

Currently all zite ask the users if the site lagen than 10MB.If we default way allow do download all optional files.And we make a button what this can Enable/disable when the zite are loading.Default way enable. Then those who do not know how to use computers, or 0Net than older or not pro  users.They could easily access the files rar,zip,mp4,mp3 etc.I heard the developers working a new setup. Started working on next big feature: [Content addressed data that should enable site-independent file storage and de-duplication](https://github.com/HelloZeroNet/ZeroNet/issues/2192).This can be made much more effective.We can make a option what show the total file size in a single zite.And we can keep the in the right 0Net icon but with a function what show the total files size in a zite.Example if a zite are loading show a button what default way enable to download all files larges and small files too without user interaction.If the user think the zite are too large can enable him to this button can turn off large files download.The zite ask the user when a zite larger than 10MB so that's not a problem.In this period I recommend show the large and small files size too.And this pint show a button which requires no user intervention if the user want all files.If he not want but later change his mind the are the right 0Net icon. And it would be easier for everyone to use the software.

**Describe alternatives you've considered**

In torrent clients the user can download all files what include the current torrent files.And all files are seeded.In torrent files extreme large 10-20-60GB or larger single files are working,in 0Net are not.Torrent clients resume the download.Utorrent web which you can play HTML5 files using the browser without completing the download resume the download and download all files default too.And if something a user doesn't want then picks out the check mark in the current file.For advanced users here on 0Net it could be solved in the same way.And who didn't know too much from this its the solution above the simplest.

**Additional context**

With single files it would be would be [necessary a button what shows and allow resume the download when the user click a file without +seed button integratoin](https://github.com/HelloZeroNet/ZeroNet/issues/2041).This can works with multiple user sites and single sites too.With multiple user sites in the right 0Net icon  would be easier to show the total files too.It would be useful to allow multiple user sites to download not just the whole site files.But select some users too.

[...]

PR #2647, opened 2020-10-14
Turn on auto for optional files and raise site limit to 1024 MB

JabbaDesilijicTiure:
https://github.com/HelloZeroNet/ZeroNet/issues/2622

I think we should raise the file limit as well, but need to see first what the BigFilePlugin do about this...

### DIFF
--- a/src/Config.py
+++ b/src/Config.py
@@ -251,7 +251,7 @@ class Config(object):
                                  metavar='address')
         self.parser.add_argument('--dist_type', help='Type of installed distribution', default='source')
 
-        self.parser.add_argument('--size_limit', help='Default site size limit in MB', default=10, type=int, metavar='limit')
+        self.parser.add_argument('--size_limit', help='Default site size limit in MB', default=1024, type=int, metavar='limit')
         self.parser.add_argument('--file_size_limit', help='Maximum per file size limit in MB', default=10, type=int, metavar='limit')
         self.parser.add_argument('--connected_limit', help='Max connected peer per site', default=8, type=int, metavar='connected_limit')
         self.parser.add_argument('--global_connected_limit', help='Max connections', default=512, type=int, metavar='global_connected_limit')
@@ -299,7 +299,7 @@ class Config(object):
         self.parser.add_argument('--threads_crypt', help='Number of threads for cryptographic operations', default=2, type=int)
         self.parser.add_argument('--threads_db', help='Number of threads for database operations', default=1, type=int)
 
-        self.parser.add_argument("--download_optional", choices=["manual", "auto"], default="manual")
+        self.parser.add_argument("--download_optional", choices=["manual", "auto"], default="auto")
 
         self.parser.add_argument('--coffeescript_compiler', help='Coffeescript compiler for developing', default=coffeescript,
                                  metavar='executable_path')
--- a/src/Config.py
+++ b/src/Config.py
@@ -252,7 +252,7 @@ class Config(object):
         self.parser.add_argument('--dist_type', help='Type of installed distribution', default='source')
 
         self.parser.add_argument('--size_limit', help='Default site size limit in MB', default=1024, type=int, metavar='limit')
-        self.parser.add_argument('--file_size_limit', help='Maximum per file size limit in MB', default=10, type=int, metavar='limit')
+        self.parser.add_argument('--file_size_limit', help='Maximum per file size limit in MB', default=30, type=int, metavar='limit')
         self.parser.add_argument('--connected_limit', help='Max connected peer per site', default=8, type=int, metavar='connected_limit')
         self.parser.add_argument('--global_connected_limit', help='Max connections', default=512, type=int, metavar='global_connected_limit')
         self.parser.add_argument('--workers', help='Download workers per site', default=5, type=int, metavar='workers')


JabbaDesilijicTiure:
Raise the default limits to avoid: https://github.com/HelloZeroNet/ZeroNet/issues/2654

imachug:
> Raise the default limits to avoid: #2654

Bugfixing doesn't work like that.

JabbaDesilijicTiure:
You have that blog only, nobody else so the that is your bug. You fix it.

JabbaDesilijicTiure:
You have that bug only, nobody else, so that is your bug. You fix it.

imachug:
> You have that bug only

I'm the only one who cares enough to report it.

mx5kevin:
`self.parser.add_argument('--size_limit', help='Default site size limit in MB', default=10, type=int, metavar='limit')`

If a user click a link and only load the clicked content have a sense if someone is not interested in the whole website just a part of it. Forums and multi user zite like in Zero Talk with this option, comments will not be displayed properly if the user not allow larger limit.

`self.parser.add_argument('--file_size_limit', help='Maximum per file size limit in MB', default=10, type=int, metavar='limit')`

The line doesn't make any sense, it just causes problems, no matter what the limit is. Site owners can not share files what larger than the limit. Users can not download files what larger than the limit. **The program ask in if the whole site limit larger than 10MB to allow larger limit. But not ask if a files size larger than 10MB (optional and not optional files) to allow larger limit include the largest file in the zite to download default way.** Single user sites will be not work correctly this way. That would make sense HTML, CSS, JS, pictiures, PDF these are displayed by the browser. And ask if the zite contains other files if they have already been downloaded to download these as well. All other files could be treated as optional files default way without adding any line in the .json file. Because all other files are what the browser does not display. This is an optional option if the user does not need larger files because the downloads, video files etc will not work this way, or must wait the user to someone seed the files this way. Or select what files want the user like in bittorrent. Default way what the site owner share with the zite they are part of the website without it the website will be unenjoyable. In a single user zite default setup download all files so users have instant access to everything. If anyone has this problem then he will set separately which two other options want.

mx5kevin:
`        self.parser.add_argument('--size_limit', help='Default site size limit in MB', default=10, type=int, metavar='limit')
        self.parser.add_argument('--file_size_limit', help='Maximum per file size limit in MB', default=10, type=int, metavar='limit')
`
If both lines were completely removed and the limit was not changed, the entire program would work much more efficiently. These are the two preset values only make the program currently unusable. Currently, there would only be benefits to deleting the values completely.

imachug:
> The line doesn't make any sense, it just causes problems, no matter what the limit is.

I think this only applies to required files, not optional files. I think it's a good idea to make big files optional anyway.

imachug:
> Currently, there would only be benefits to deleting the values completely.

So if you accidentally open a malicious site, it will immediately fill up your disk? Sorry, that does not sound like a good idea.

mx5kevin:
> 
> 
> > Currently, there would only be benefits to deleting the values completely.
> 
> So if you accidentally open a malicious site, it will immediately fill up your disk? Sorry, that does not sound like a good idea.

The program will not work with with these pre-set limits. Web pages can be deleted completely. The system shows that files are loading and what site are downloading files. So they can't load the hard drive either. Even if that happened in mobile devices deletes the user of that page and the problem is resolved. Currently, it would have many more advantages than disadvantages. We need transparent information about the content size of the entire page and not limits. And as optional options for advanced users by not forced all user to this too. And by default, not all users are forced with pre-set limits. The problem is that these features do not work correctly. By removing the two rows which is a few seconds. The whole system would be much more efficient. If ask the system a question from the user like „Site is larger than allowed example 2000MB/10MB set limit to 2000MB”  an average user doesn’t even know what it is. It also just causes a problems. And it doesn't even download large optional files if the user set it up. In vain  want to get them the user the setup does not do this. The whole thing just limits the operation of the system. A modern website 1-2 GB large. And an optional file are not  10MB or 30MB they are 100MB 500MB and larger. And over 30MB with the new setup nothing can be downloaded in a single site what larger than 10MB or the new setup with 30 MB. I agree with the optional option for advanced users. But the current setup it crashes the whole system. A selective solution would work to separate those files what the browser use to display a web page. And another files video, downloadable content and etc which can be downloaded later. Because they are not required to display the full content of the website. The key thing is not to have to ask for everything to be downloaded. If the user are not prompted for large files then  can set an option when the web page loads. But this should not be imposed on those who cannot use it. Because the user will see that the content on the website does not work. By default, the program does not want to decide what file to download and what not. In a single user zite by default, download everything or by the way the content will not work for the user. For publishers the limitations this service is made into a limited storage space on a closed network. And there's the problem the optional parameter must setup in the content.json file and this cannot be set by many website owners. If we limit this would be worth measuring for the storage device and warn the user if example want use the 10% of the current site the full storage. 

See the issues: https://github.com/HelloZeroNet/ZeroNet/issues/2622 , https://github.com/HelloZeroNet/ZeroNet/issues/2566, https://github.com/HelloZeroNet/ZeroNet/issues/2243 , https://github.com/HelloZeroNet/ZeroNet/issues/2602,

mx5kevin:
> 
> 
> > The line doesn't make any sense, it just causes problems, no matter what the limit is.
> 
> I think this only applies to required files, not optional files. I think it's a good idea to make big files optional anyway.

Optional files are limited to 10MB in the right menu. The option is not even found by most users. Even if they click on it, even files what more than 10 megabytes will not be downloaded. The other problem is that nothing indicates the largest file size on the page. And here the option should be set separately to a higher value what the biggest single optional file size in the zite. There is no automatic way to download the full page in a single zite with all files. If the user set higher limit then 10MB and the files are not optional but larger than 10MB the download just doesn't work again. Optional files also have a preset 10MB limit. Almost no one can use this option. There are few visitors and they can’t use almost any option which is a little more complicated. It’s really not good for the program to decide for itself that, however, it refuses to download large optional files. These preset restrictions they simply prevent you from downloading large files easily. And then there's that site owners need to setup the optional parameters. And not all website owners can do that. In multi user zites like Zero Up if the option are set up there it downloads all user files. All files which is less than 10MB and it can be several terabytes under 10MB files.

imachug:
Sorry, it was difficult to get through your rant, and I don't get your point. You said modern sites are 1-2 GiB each. An average user browses at least 100 sites. Do you think that using 100-200 GiB of disk storage is acceptable? That *is* worse, people will get angry because ZeroNet eats up their disk space, not because of a notification. And if someone doesn't know what a megabyte is, they'd better leave ZeroNet, there are enough idiots on the Clearnet.

JabbaDesilijicTiure:
**Yes, 200GB is perfectly acceptable.** Not everyone using ZeroNet in the first place so those who will run with these limits already aware what is going on. 10 MB sites are a big joke, a video included in a site would immediately trigger the request for allow more space, other users would freak out when they get the notification to allow more space. 

@mx5kevin I don't have homepage and I use absolutely nothing which is shipped in ZeroNet and not necessary for running the software, meaning that all plugin including UiFileManager and HelloZeroNet for me is non-existent. 

@imachug nowadays everyone has a pendrive with capacity around 256GB, 512GB, and 1TB. 
I do run ZeroNet only from external storage with large free space in a virtual environment. 

Lastly, those who don't have @HelloZeroNet home page will never ever get a notification about the file size in the first place.

imachug:
> Yes, 200GB is perfectly acceptable

Heck, some of the devices I'm running ZeroNet on don't have half of that, and other devices don't have half so much free space. Most disk drive that are installed in modern laptops are 256 to 512 GiB, Windows/Linux/Mac OS (bonus points if you do dual boot) take up at least a quarter of that, user data takes up a lot of space too. Do you want to know the reason why people use web apps instead of installing them? -- right, because they take up space.

imachug:
> nowadays everyone has a pendrive with capacity around 256GB, 512GB, and 1TB.

First: many people, not everyone. Second: that capacity is not for ZeroNet exclusively.

> Lastly, those who don't have @HelloZeroNet home page will never ever get a notification about the file size in the first place.

You are several years behind the mainline.

JabbaDesilijicTiure:
Anyway what is the problem? I raised the default file size limit to 30MB and the site size limit to 1024MB what's your problem with this @imachug. ZeroNet is not a "web app" and all site/file you visit/open you will download and seed as well. 

Having a 10MB site size limit, while the file size limit is also 10MB is stupid as fck!

imachug:
> Downloading a few big file, may make your CSS, fonts and even HTML files to fail downloading and so makes your site unusable!

Files are downloaded in a specific order so this usually shouldn't happen. If it does, please send me how to reproduce the issue.

> I raised the default file size limit to 30MB and the site size limit to 1024MB what's your problem with this

I don't want to accidentally download a whole 1 GiB of content (possibly illegal) in background. 10 MiB is a good limit for basic code and some text. When there's not enough space for user data, a nice notification pops up with a big 'Increase size' text on it, so I judge if the site is important to me and if it does, I allow the increase.

JabbaDesilijicTiure:
While? Torrent is not illegal? You talk BS. **The FILE LIMIT IS 30 MB!  NOT 1 GB!**

imachug:
>  NOT 1 GB!

But site limit is.

> Torrent is not illegal?

Not all of them are illegal, like not all ZeroNet sites are illegal. The difference is that most torrent clients show you the list of files while ZeroNet does not.

mx5kevin:
imachug your idea to download small parts of everything doesn’t work in practice. All such solutions make the content inaccessible. I say this from experience as zite and file owner. And after years of torrenting, where there is a basis for comparison. The result of all such efforts will be that the content will not be available for the users.

Allow the increase site limit popup option does not download optional or large files.

As zite and file owner I can tell from experience the half-downloaded files an inoperable thing. The single files and downloading just part of a single zite too. If the entire content is not downloaded the are no seeders in sinle user zites and files too. No one other than the content owner will share the content. It is a waste of resources. In torrent always download all users the entire content. And there is no limit to how big the files or total file size in the torrent can be. The total content size of the website must be showed and who needs to download it, who doesn’t delete it. ZeroNet is a P2P network the user pays to help share the file. That someone downloads a little bit of everything inoperable on such a network. It’s like stealing the content owner who seed the file. There are communities on the network that regularly use specific sites. Casual visitors do not seed the content. Visiting 100 sites a day is not necessarily well-intentioned. There is no need for malicious scanner visitors on the network. And this network was not invented for this. Preset site size limit in MB, file size limit in MB does not make sense because it just crash sites. It is worth looking at how much user storage space is available. And to measure that how much space the content takes up. Today there are a lot of big data storage 300MB 500MB already basic in a old pc. 2TB 4TB 10TB are basic in modern PC and 5G are comming.  Limiting files and zites how many MB can be default way are pointless the same limit in a 10GB storage and 10TB storage. Today it is an average website 1-2 GB large, it will be even bigger in the future. Hundreds of megabytes of files 10MB-1,5GB are not uncommon on websites. Everyone wants everything right away. And this is not possible if users download incomplete pages and files. The downside to this is the need for disk space. This can be freed by deleting the files and zites which are not regularly visited. It cannot be based on a clearnet service because there is no stable bandwidth, no 24/7 online users.

imachug:
> In torrent always download all users the entire content

Sorry, what? You are kind of wrong.

JabbaDesilijicTiure:
@imachug what should be the limit for files and sites? Be reasonable.

imachug:
Most sites on ZeroNet that I host are under 10 MB. My response is obvious.

imachug:
I'm okay about increasing required file limit from 1 MB to 10 MB though.

JabbaDesilijicTiure:
There is no such thing as required files. This pull request is not about you @imachug a few ttf font and one video already hitting the limits in any sites. 10MB is not good for files or for the site. The site limit in this pr is 1024MB, and the file limit is 30MB, I'm fine with lowering the site size limit to 512MB.

imachug:
> There is no such thing as required files.

Required files are all files except optional files.

imachug:
> a few ttf font and one video already hitting the limits in any sites

ZeroNet. Does. Not. Work. Well. Without. BigFile. Plugin. That's it, big optional files, even if they are always downloaded, work better than big required files.

mx5kevin:
`self.parser.add_argument('--file_size_limit', help='Maximum per file size limit in MB', default=10, type=int, metavar='limit')`

The most troubling parameter prevents site owners from sharing larger files than 10MB as non optional. 

`self.parser.add_argument('--size_limit', help='Default site size limit in MB', default=10, type=int, metavar='limit')`

At the value specified in the limit, the program downloads files randomly. And other users can no longer download from those who have not raised the limit by then which is the real size of the website.

Need to decide what a site owner is destining for a page. That visitors download all files or choose who wants what. If there is an option that users can choose from there is a natural process that some of the files become inaccessible. If there was no limit in single file size per site then every site owner could decide that larger files are part of the entire page or files intended for different users. As long as there is such a limit, no matter how much it restricts the operation of the websites. This is not tolerable on a website what the site owner intended so that all content is immediately available. Here if all users not download the full content. Then the full content cannot be downloaded from these users. Very few users seed the contents not just here, but in torrent networks everywhere. Nowhere is the content size limited in torenting. This alone is a unique idea of the ZeroNet network what not working. The size of the entire content must be shown and the user has to decide need it or not. A security feature would make sense based on the total storage capacity of the user. And what only ask when a site use the 10% of the total capacity. The current setup force all users to always give a higher value. At this value, the program downloads random files. No longer downloadable the full content from them who does not have the full content. With single files what larger than the limit it can't even be downloaded automatically. The program would already know these functions but is limited with the 10MB/file and 10MB/total site size values. No matter how high this limit is, it only limits content owners and users. And it is no longer possible to get the full content from users who have not downloaded the full content just random things from the full content. This will prevent users from displaying the entire content correctly. Torrent programs also have a mobile and all platform version. But there is no such limitations in torrent programs. If a site owner setup a zite choose who wants what because its contains  more gigabytes muvie, game etc files there html, css, pictures, js there are also all downloaded for the page to appear. Or multi user zites. And another files only when the user click on it. Thus, as a second option, a website can have multiple terabytes. No one needs the full content on it so it is no problem if some of the content is permanently lost. For sites where users need full content this is an unacceptable option.  If the site owner intended a page to make all content available to all users. And there the users start sorting then the content will be unenjoyable. The other problem is that the program unnecessarily asks questions that users don’t know what to answer. The files that should be selected should be a separate one Downloads folder so it would be easier to find them than scattered on websites.

JabbaDesilijicTiure:
> The most troubling parameter prevents site owners from sharing larger files than 10MB as non optional. - @mx5kevin 

I agree with him @imachug, we need to find a solution to this. 
I asking everyone to propose new site and files sizes in the comments below.

mx5kevin:
> 
> 
> I asking everyone to propose new site and files sizes in the comments below.

`Maximum per file size limit in MB', default=2000,`

`Default site size limit in MB', default=1000,`

2000 MB perfect for a larger video or etc file like a 1920X1080 Full HD video. A fair limit which also fits the largest file on the website. Even a slower internet is an acceptable file size. No one will upload a single file larger than this in a single user zite as non optional.

1000 MB total files are enough for a quality website. The site is larger than allowed question under this limit are unnecessary.
I calculated from real data using single user modern sites what i seed. If someone bring here his own website whit this parameters can run the website in the ZeroNet  network without problems. In a 100 Mbit/s internet websites are downloaded without any problems with the same parameters. I used it within the given parameters the system without problems.

Where the goal is to make a website where everything is immediately available to everyone these are correct parameters for the users and website owners too. The advantage is that all files are downloaded by all users. So nothing is lost from the content.

JabbaDesilijicTiure:
:rofl: 

1GB max site size, and 2GB max file size seems very logic to me. 

Max file size should be a little bit more than 10MB and site size should be around 512MB. What you think @imachug? Can you test the BigFilePlugin with 30MB/file?

mx5kevin:
Using a file priority download HTML, picturs, CSS, JS  are need to bisplay a website the browser. If these are downloaded first then the website will load much faster. If the program check the total disk space and warn if the website want use the disk space 10% then ask site is larger than allowed question then should not be have a Maximum per file size limit in MB and Default site size limit in MB. With the "Default site size limit" should be raised in direct proportion with the "Maximum per file size limit" when the program asks the site is larger than allowed question. This way, no matter how big files can be uploaded to a website without any "Maximum per file size limit" restrictions. If this solution is feasible later then there is no need preset "Default site size limit in MB" and "Maximum per file size limit in MB" parameters. This would allow the system to adapt to changing needs and system requirements in long term. The solution would work perfectly in a small capacity 10 GB storage and a 10 TB storage too. And there would be no need for unnecessary user interactions if someone has more storage capacity.

JabbaDesilijicTiure:
I think is better to just leave this now. Need testing because as @imachug [said](https://github.com/HelloZeroNet/ZeroNet/pull/2647#issuecomment-719506221) BigFilePlugin is needed for ZeroNet... Raising the limits would make that plugin obsolete.
Will reopen when needed.

JabbaDesilijicTiure:
@mx5kevin Are you Hungarian?

PR #2650, opened 2020-10-14
GitHub Actions upgrade to actions/checkout@v2

cclauss:


JabbaDesilijicTiure:
I think there are more important things need to be cared for. I don't see you here a lot so now that you are here you should maybe help here: https://github.com/HelloZeroNet/ZeroNet/issues/2622

I seen that you closed https://github.com/HelloZeroNet/ZeroNet/pull/2649 which was also not urgent.

PR #2653, opened 2020-10-15
Warning: Never run ZeroNet as root user!

JabbaDesilijicTiure:
Add some warning instead of encourage users to run ZeroNet as root


imachug:
That applies to literally all software though. It may be a good idea to change the spelling to 'For extra security, run ZeroNet under a separate user with no access to your personal data' or alike.

JabbaDesilijicTiure:
We talk about ZeroNet not all software so this stays how it is.

JabbaDesilijicTiure:
This not so important now.

PR #2657, opened 2020-10-20
Adding OGG MIME

JabbaDesilijicTiure:
.ogg is **not listed** in the mimetypes module: https://github.com/python/cpython/blob/3.9/Lib/mimetypes.py and since **we don't know whether the ogg file contains audio or video**, we need an `application/ogg` mime for those files with the `.ogg` extension.


imachug:
This is not going to fix the mentioned issue but it's a worthy change anyway. Please add `.ogv` and `.oga` too though.

JabbaDesilijicTiure:
@imachug since **we don't know whether the ogg file contains audio or video**, we need an `application/ogg` mime for those files with the `.ogg` extension.

imachug:
Yes, that's why I asked you to add support for `.ogv` and `.oga` extensions as well, for which more specific MIME types can be defined.

JabbaDesilijicTiure:
@imachug Makes no sense to add two mime types, when one is enough: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Configuring_servers_for_Ogg_media

imachug:
Listen in. Even the article you're referring to notes that there are `.ogg`, `.oga` and `.ogv` extensions. All I'm asking for is that, if you added `.ogg`, please also add MIME types for `.oga` and `.ogv`.

JabbaDesilijicTiure:
Adding **ova** and ogv extensions. :roll_eyes: 
There you go @imachug added the oga and ogv extensions... You was right, need all 3 of them. Thanks

imachug (approved):

PR #2668, opened 2020-11-01
Update LICENSE

JabbaDesilijicTiure:
.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -9,19 +9,3 @@ GNU General Public License for more details.
 
 You should have received a copy of the GNU General Public License
 along with this program. If not, see <https://www.gnu.org/licenses/>.
-
-
-Additional Conditions :
-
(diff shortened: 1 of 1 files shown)


JabbaDesilijicTiure:
Why the license need to referring to conflict? 

>All the non-key contributions are governed by [CC BY-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0/), unless specified above, a contribution is ruled by the type of contribution **if there is a conflict between two contributing parties of repo in any case.**

According to the deleted conditions, **rewriting the comments in the python source code** can be considered "non-key" contributions while in reality those changes are made in the source code of the software, and they **should be GPLv3 in any case**.

**The removed additional conditions actually did imposed restrictions on the distribution of the source code** which directly violates the conditions of the GPLv3 license.

PR #2694, opened 2020-12-03
Replaced logo-white.png with an SVG data URI. (logo-white.png deleted)

styromaniac:
I used the SVG file from Wikipedia and gave it a brightness of 300%.

The PNG didn't look as clean as the SVG while zoomed in.


HelloZeroNet:
Thanks, sometinhg was missing from the commit, but I have created a white logo svg and replaced the .png with it in Rev4555: https://github.com/HelloZeroNet/ZeroNet/commit/3ffa3c2f792a9e695711a93ec310b0468cb862d8

styromaniac:
You could eliminate logo-white.svg with
```
style="filter: brightness(3)"
```
on logo.svg

HelloZeroNet:
Yeah, it could also work, but I'm not sure about the possible performance impact in various devices, so I think it's better to use pre-rendered svg.

styromaniac:
> Yeah, it could also work, but I'm not sure about the possible performance impact in various devices, so I think it's better to use pre-rendered svg.

Good point. Also, I'm glad we can embed SVG files now instead of having to resort to data URIs.

PR #2756, opened 2021-07-09
remove merkletools dependency

rllola:
This PR removes dependency to pymerkletools (https://github.com/Tierion/pymerkletools).

It instead integrates the class in the codebase and update the BigFile plugin to use it. Only support for the hash function we use.


p2p-publisher:
Merged in https://github.com/peertopeernetwork/p2p/commit/d70cb539d64a864e0a956f84ec374d0bfa306060

caryoscelus:
> It instead integrates the class in the codebase

did you simply copy-paste code from merkletools to this PR ? one can argue for and against such approach in general , but for it to be legal you have to keep authorship and license notices from original codebase...

PR #2762, opened 2021-08-19
feature: launch namecoin

ethicnology:
# Intent
Launch a Namecoin client such as Namecoin Core if it's installed when Zeronet launches.

## NamecoinUtils.py give a set of functions.
+ is_namecoin_installed check if a **.namecoin** folder exists in the (linux) user directory.
+ is_namecoin_running check if a running process contain the string namecoin.
+ check_namecoin_rpc_conf ensures that the necessary variables are set.
+ start_namecoin execute a subprocess of **namecoind**
+ kill_namecoin kill the given pid process

## Add __del__ function to SiteManagerPlugin.py
+ __del__ function kill the Namecoin client if it was launched by ZeroNet on ZeroNet shuts down.

## Process
+ self.config is initiated with CLI args, if provided
+  is namecoin installed ?
    + is namecoin running ?
        + if conf is not set, check for namecoin rpc conf in **namecoin.conf**
            + if there is no **namecoin.conf** raise missing parameters error.
    +  start **namecoind** (if path is given)

## Limitations
For the time being, it only works on linux, we can upgrade it for windows and macos users later

## Nota bene
I pushed with debug logs to make it easier for you to test, i will remove them before merging, if you agree

### DIFF
--- a/plugins/disabled-ZeronameLocal/NamecoinUtils.py
+++ b/plugins/disabled-ZeronameLocal/NamecoinUtils.py
@@ -1,0 +1,43 @@
+import os, signal, json, requests, subprocess, time
+from pathlib import Path
+from configparser import ConfigParser
+import psutil
+
+
+def is_namecoin_installed():
+    home = str(Path.home())
+    dot_namecoin = os.path.join(home, '.namecoin')
+    is_dot_namecoin = os.path.exists(dot_namecoin)
+    return is_dot_namecoin
+
+
+def is_namecoin_running():
+    for proc in psutil.process_iter(['pid', 'name', 'username']):
+        if "namecoin" in proc.name():
+            return proc
+
+
+def get_namecoin_conf():
+    namecoin_conf = os.path.join(str(Path.home()), '.namecoin', 'namecoin.conf')
+    parser = ConfigParser()
+    with open(namecoin_conf) as stream:
+        parser.read_string("[fake-section]\n" + stream.read())
+        config = dict(parser.items('fake-section'))
+        return config
+
+
+def check_namecoin_rpc_conf(config):
+    if None in (config['rpcconnect'], config['rpcport'], config['rpcuser'], config['rpcpassword']):
+        return False
+    else:
+        return True
+
+
+def start_namecoin(path):
+    process = subprocess.Popen(["{}namecoind".format(path)])
+    return process.pid
+
+
+def kill_namecoin(pid):
+    os.kill(pid, signal.SIGTERM)
+    return True
--- a/plugins/disabled-ZeronameLocal/SiteManagerPlugin.py
+++ b/plugins/disabled-ZeronameLocal/SiteManagerPlugin.py
@@ -4,23 +4,57 @@ from Config import config
 from Debug import Debug
 from http.client import HTTPSConnection, HTTPConnection, HTTPException
 from base64 import b64encode
+from pathlib import Path
+from .NamecoinUtils import is_namecoin_installed, is_namecoin_running, get_namecoin_conf, check_namecoin_rpc_conf, start_namecoin, kill_namecoin
 
 allow_reload = False # No reload supported
 
 @PluginManager.registerTo("SiteManager")
 class SiteManagerPlugin(object):
-    def load(self, *args, **kwargs):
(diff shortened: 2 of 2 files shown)


plugins/disabled-ZeronameLocal/NamecoinUtils.py
```diff
@@ -0,0 +1,43 @@
+import os, signal, json, requests, subprocess, time
+from pathlib import Path
+from configparser import ConfigParser
+import psutil
+
+
+def is_namecoin_installed():
+    home = str(Path.home())
```
rllola, line 8: Does `Path.home()` will work for macOS and Windows ? 
Windows XP path is `C:\Documents and Settings\<username>\Application Data\Namecoin\namecoin.conf`
Windows Vista 7, 10, 11? path is `C:\Users\<username>\AppData\Roaming\Namecoin`
macOS path is `/Users/<username>/Library/Application Support/Namecoin/namecoin.conf`

plugins/disabled-ZeronameLocal/SiteManagerPlugin.py
```diff
@@ -4,23 +4,57 @@
...
+        self.config = dict({
+            'rpcconnect': config.namecoin_host,
+            'rpcport': config.namecoin_rpcport,
+            'rpcuser': config.namecoin_rpcuser,
+            'rpcpassword': config.namecoin_rpcpassword,
+            'path': config.namecoin_path
+        })
+        self.log.debug("Namecoin config through CLI ? {}".format(self.config))
+        self.log.debug("Namecoin installed ? {}".format(is_namecoin_installed()))
+        if is_namecoin_installed():
```
rllola: If we are going for now only support linux we can check first that the platform is linux before checking `is_namecoin_installed`.

```python
if sys.platform.startswith('linux') and is_namecoin_installed():
```

ethicnology: indeed, it is preferable 5c2911e028665b60a6349115f073a66c9f09e2d9

plugins/disabled-ZeronameLocal/NamecoinUtils.py
ethicnology: Windows XP, Vista and 7 are no longer supported, are they plenty of users with theses windows versions ?

wtfbuddy: @ethicnology do not hardcore any path.

wtfbuddy:
@rllola You claim to be a "zeronet specialist" on namecoin.org despite having no connection to ZeroNet. 

Why would you need to start Namecoin automatically? This is the stupidest pull request I seen for long time here. 

I appreciate your pr @ethicnology, I strongly suggest you to avoid @rllola. She is a pro-troll.

wtfbuddy:
If you remove all "killing namecoin" part and the "autostart" (launch) namecoin. the pr would look nice.

wtfbuddy:
@ethicnology think about that you can lookup names using Electrum-NMC too! You don't need to install Namecoin Core. Please rewrite this pr because in the current form is unacceptable.

wtfbuddy:
@rllola and @imachug: open your mouth and tell us why starting and then killing namecoin is necessary this way. Idiots, any of you even use Namecoin?

peertopeernetworkdeveloper:
Do not even dream about @rllola we not going to merge your crap that has Namespace restrictions. On the other hand ZeroNet already supports Namecoin Core over RPC. 

Also forget that you will ever merge Electrum NMC that also has serious restrictions on Namespaces while forcing on users useless namespaces like dd/, id/, idd/. 

You and @yanmaani multiple times tried to spam zeronet with unnecessary Namecoin related malware.

peertopeernetworkdeveloper:
@rllola be kind and remove yourself from namecoin.org, you're not a "zeronet specialist" you're just a french troll who trying to push malware into zeronet.

peertopeernetworkdeveloper:
@mx5kevin since Tamas is gone ZeroNet constantly under attack by French, American and other trolls who trying to add malware into ZeroNet, especially the Namecoin trolls.

PR #2835, opened 2023-09-04
Wupg98 patch 1

wupg98:


caryoscelus:
[here's](https://github.com/zeronet-conservancy/zeronet-conservancy/tree/zne-massive-rework) a backup with commit history

PR #2856, opened 2024-11-28
ZeroNameCached Plugin

wolfitdm:
You can now have and define multiple bit resolvers.

--bit_resolvers 1Name2NXVi1RDPDgf5617UoW7xA6YrhM9F 1SitesVCdgNfHojzf2aGKQrD4dteAZR1k 1E97TpiDiCj1WGhZWxoKjBV9KkVty1PFsq

A built-in cache to increase speed...

With the additional option to set the interval...

{

"domains": {},

"update_interval": 180,

"last_updated": 0,

"update_time": 0,

"use_cache": True
}

Path

core\domains_cache.json

A little April Fool's joke:.....

My next goal is to build a kind of blockchain plugin to replace Zeroid...

It already works in Python, and with it I want to offer ZeroNet the possibility of being the only platform to have a To have a kind of pluggable blockchain that serves for verification like Namecoin and as a currency...


wolfitdm:
cache

wolfitdm:
I think I found a bug here, and I made the plugin even faster, you can get into a need/get loop

wolfitdm:
Improved runtime and lighter support to prepare for pluggable blockchain,But due to Christmas, a lot to do, and only postponed until January 6, 7 or 8, 2025

